PR #7, opened 2016-10-26
install tor

BitWaspBitWasp:
Debian

after "apt-get install tor" TOR is not working
(I had 2 warning during installation:
  insserv: warning: script 'K02my_autorun' missing LSB tags and overrides
  insserv: warning: script 'my_autorun' missing LSB tags and overrides)

root@srv2 /var/log # /etc/init.d/tor  start
root@srv2 /var/log # /etc/init.d/tor  status
root@srv2 /var/log # /etc/init.d/tor  restart
root@srv2 /var/log # /etc/init.d/tor  reload
[FAIL] Reloading tor daemon configuration...failed (not running - there is no /var/run/tor/tor.pid).


weaselp:
Pull requests are not a venue for bug reports.  Please use proper channels.

PR #28, opened 2018-03-23
Use an AF_NETLINK socket to query the "external" IP address

dgoulet-tor:
This commit adds the new files src/common/netlink.{c|h} that implement this
code for Linux only.

Merged from yawning/bug12377_linux branch.

Fixes #12377


teor2345:
https://trac.torproject.org/projects/tor/ticket/12377 is needs_revision in tor unspecified.

PR #35, opened 2018-03-31
build system: use libtool to make "libtor"

bassosimone:


bassosimone:
With 08e8b32 the build seems to be passing. I should probably flag [the following rust warning](https://travis-ci.org/torproject/tor/jobs/360736500#L1840), which I don't know exactly how to avoid:

```
*** Warning: Linking the shared library src/or/libtor_api.la against the
*** static library ./src/rust/target/release/libtor_rust.a is not portable!
```

.travis.yml
```diff
@@ -59,7 +59,7 @@ addons:
 env:
   global:
     ## The Travis CI environment allows us two cores, so let's use both.
-    - MAKEFLAGS="-j 2"
+    - MAKEFLAGS="-j 2 V=0"
```
tlyu: I'm not sure this is the best strategy. libtool seems to be echoing the entire compiler command line with each compilation wrapper invocation, which seems to nearly double the size of the build log. We want the entire compiler command line in the log file because that can be important for troubleshooting a failed build. Is there some way to get libtool to be quieter? Also the `--disable-silent-rules` is what turns on the verbosity in the first place, so maybe that's the better place to make this sort of change?

tlyu:
The Trac ticket for this is https://trac.torproject.org/projects/tor/ticket/23846

.travis.yml
bassosimone: @tlyu:
> Is there some way to get libtool to be quieter?

Yes, [we can use the LIBTOOL environment variable](https://www.gnu.org/software/automake/manual/html_node/Libtool-Flags.html).

(It's _also_ possible to use `AM_LIBTOOLFLAGS`. The question is whether we want to disable libtool verbosity in Travis only or always.)

> Also the --disable-silent-rules is what turns on the verbosity in the first place, so maybe that's the better place to make this sort of change?

The `V=0` and `--enable-silent-rules` [seem to do the same job](https://www.gnu.org/software/automake/manual/html_node/Automake-Silent-Rules.html#Automake-Silent-Rules). The difference is whether one wants to do that at `./configure` time or at `make` invocation time.

Thanks for the other comments in the Track ticket! I'll reply to them soon.

teor2345:
https://trac.torproject.org/projects/tor/ticket/23846 is in needs_information in 0.3.5.

bassosimone:
We have actually [agreed on trac to use another approach](https://trac.torproject.org/projects/tor/ticket/23846#comment:42). As such, this pull request may be closed.

PR #36, opened 2018-04-02
[WIP] v3 onion service client authorization

haxxpop:
This pull request is for code review only because this is a big feature and it will take a lot more time to do it.


haxxpop:
This should push a new descriptor to HSDirs, if I change a client list.

src/or/hs_config.c
```diff
@@ -266,6 +265,94 @@ config_service_v3(const config_line_t *line_,
       have_num_ip = 1;
       continue;
     }
+    /* Client authorization */
```
asn-d6: Note: We need to add man page entries for these new suboptions. No need to do it haxxpop, but we will need to do it before merging.

src/or/hs_descriptor.c
```diff
@@ -234,53 +234,69 @@ build_mac(const uint8_t *mac_key, size_t mac_key_len,
...
+  uint8_t *secret_input = NULL;
+  size_t secret_input_len;
   crypto_xof_t *xof;
 
   tor_assert(desc);
+  tor_assert(secret_data);
   tor_assert(salt);
   tor_assert(key_out);
 
+  secret_input_len = secret_data_len + DIGEST256_LEN + sizeof(uint64_t);
```
asn-d6: Hmm, `build_secret_input()` has a bit weird interface now that it receives an already allocated buffer and puts stuff in it. I think perhaps a more clean way to do this would be to have `build_secret_input()` allocate the buffer and return it. That means you wont have to compare `dstlen` with `outlen` etc.

If you do that, `build_secret_input()` will not be able to fail, so you don't need to check its output for NULL. You can just assert it if you want.

src/or/hs_client.c
```diff
@@ -1210,11 +1210,23 @@ hs_client_decode_descriptor(const char *desc_str,
...
   uint8_t subcredential[DIGEST256_LEN];
   ed25519_public_key_t blinded_pubkey;
+  hs_client_service_authorization_t *client_auth = NULL;
+  curve25519_secret_key_t *client_sk = NULL;
 
   tor_assert(desc_str);
   tor_assert(service_identity_pk);
   tor_assert(desc);
 
+  /* Check if we have a client authorization for this service in the map.
```
asn-d6: Perhaps we should make this "Get client auth key given service identity" code into a function to make it more tidy? :)

src/or/hs_descriptor.c
```diff
@@ -1583,18 +1677,30 @@ desc_decrypt_encrypted(const hs_descriptor_t *desc,
...
   tor_assert(desc);
+  tor_assert(desc->superencrypted_data.clients);
   tor_assert(decrypted_out);
 
-  /* XXX: We need to decrypt the descriptor properly when the client auth
-   * is enabled. */
-  (void) client_sk;
+  /* If the client secret key is provided, try to find a valid descriptor
+   * cookie. Otherwise, leave it NULL. */
+  if (client_sk) {
```
asn-d6: Hmm, this "We have desc auth enabled, find descriptor cookie" block would be nice if it had its own function. Perhaps something like `find_client_auth_desc_cookie()`.

Also, I wonder if we should throw a special `log_warn()` if we have client auth enabled and we couldn't find the descriptor cookie.

src/or/hs_descriptor.c
```diff
@@ -1415,6 +1445,70 @@ encrypted_data_length_is_valid(size_t len)
...
 
+/* Decrypt the descriptor cookie given the descriptor, the auth client,
+ * and the client secret key. On sucess, return 0 and a newly allocated
+ * descriptor cookie descriptor_cookie_out. On error or if the client id
+ * is invalid, return -1 and descriptor_cookie_out is set to
+ * NULL. */
+static int
+decrypt_descriptor_cookie(const hs_descriptor_t *desc,
+                          const hs_desc_authorized_client_t *client,
+                          const curve25519_secret_key_t *client_sk,
```
asn-d6: `client_sk` is a fine variable name, but I wonder if we could make it more explicit that this is for optional client auth.
Something like `client_sk -> client_auth_sk`? What do you think?

src/or/hs_service.c
```diff
@@ -1031,6 +1034,222 @@ load_service_keys(hs_service_t *service)
...
+      curve25519_public_key_generate(&client->client_pk, &client_seckey);
+
+      /* Write a newly generated secret key to a file so that the service
+       * can send this secret key to the client in a secure out-of-band
+       * way. */
+      fname_seckey = client_seckey_path_from_name(config->directory_path,
+                                                  client->client_name);
+      /* We do not need to check the return value because the secret key
+       * was appropriately generated */
+      curve25519_secret_to_base64(buf, &client_seckey);
```
asn-d6: I also wonder if we should try to prefix secret keys with some strings so that we can recognize them as client auth private keys. And same for public keys.

I'm kind afraid of HS operators accidentally sending public keys to their clients instead of private keys, since they both have the same format, and the clients' authentication would fail without a reason...

Another way, to make this less likely, is to add some sort of header on top of the public/private key files to explain the functionality of those files.

What do you think?

src/or/hs_client.c
haxxpop: Sure

src/or/hs_descriptor.c
haxxpop: Agreed

src/or/hs_service.c
asn-d6: Hmm. I wonder if we should first figure out the whole UX of client authorization before figuring out the prefix here. That is, perhaps we should figure out how the control port interface UX should work etc. to see whether a prefix is useful or not... 

Also, I have mixed feelinsg about `curve25519-private-key.[base64]` mainly because it's big and also the dash and the dot makes it harder to copy-paste.

Need to think more. Perhaps a [tor-dev] post is in order.

src/or/hs_service.c
```diff
@@ -1053,6 +1056,222 @@ load_service_keys(hs_service_t *service)
...
+                                       client_keys_dir_path,
+                                       config->dir_group_readable, 1) < 0)) {
+    goto end;
+  }
+
+  /* If the client authorization is enabled, we must have a list of clients */
+  tor_assert(config->clients);
+
+  /* Load public keys from the file. If the file does not exist,
+   * parsed_clients will be empty. */
```
asn-d6: I wonder if having both `client_authorized_pubkeys` and `client_authorized_privkeys/alice.privkey` is confusing to people. It's not really clear which key should be sent to the client by the service, and I guess that many opeartors would think that  they need to send the pubkey since they are not used to sending privkeys around. (Case in point: I was confused when setting up the service/client for testing)

Suggestion:
Perhaps we could merge these two designs into one. Instead of having a separate `pubkeys` file and a separate `.privkey` file, we could only have ` client_authorized_keys/alice.key` and `client_authorized_keys/bob.key` and `client_authorized_keys/charlie.key`. That file could contain either Alice's pubkey (if the keypair was generated by alice and sent to the service), or it could contain both Alice's pubkey and private key (if the keypair was generated by the service). In any case, the service operator would just need to send the key file to Alice. And Tor, instead of checking the `pubkeys` file, it could check whether Alice has a .key` file.

What do you think? Do you think this could simplify things or not?

src/or/hs_service.c
asn-d6: This `BIP` specifies how bitcoin is prefixing their keys: https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki

They are using base32 tho, instead of base64. I wonder if that's gonna make our keys grow too huge, like our addresses...

src/or/hs_service.c
haxxpop: Why do we need to have one file for one client?

Similar to `authorized_keys` file in SSH, if we have one file, we can see both client name and public key of all clients in the same file.

haxxpop: Another option! To prevent the issue about copying private key to the client, I think maybe we can do the same thing as SSH. That is, we will not let the service generate a keypair by itself, but if it really wants to, it needs to use a command line to generate it for each client.

What do you think?

asn-d6: Hmm, I think we originally came up with one file per client, so that it's easier and more intuitive for the service to send the right file to the right client (i.e. to avoid copy-pasta fails, searching a big file for a client name,. etc.).

The main difference I can think with SSH here is that in the SSH case both the client and the service are tech people who know how to edit configuration files, whereas in our case the client might just be a web user. And asking the web user to generate credentials for a website and send it to the website owner might be a bit too much. That's why I think that preserving the "service sends key to the client" interface is a good idea from a UX perspective.

Perhaps the time to get in touch with the UX team is getting closer :)

src/or/hs_service.c
```diff
@@ -1053,6 +1056,222 @@ load_service_keys(hs_service_t *service)
...
+
+/** Parse the content of a client_key file in <b>client_keys_str</b>
+ * and add hs_service_authorized_client_t's for each parsed client to
+ * <b>parsed_clients</b>. Return the number of parsed clients as result
+ * or -1 for failure. */
+static int
+parse_client_keys(strmap_t *parsed_clients,
+                  const char *start,
+                  const char *end)
+{
```
dgoulet-tor: I think it would be much better to use the `smartlist_split_string` model in here like `rend_parse_service_authorization()` does. Each line, we split and extract.

src/or/hs_service.c
```diff
@@ -1638,6 +1711,20 @@ build_service_desc_keys(const hs_service_t *service,
     ret = -1;
```
dgoulet-tor: Oh I think we want a `goto end` here, so we don't generate the emphemeral auth kp as well. For this, the label will need to be added. Same goes for a `goto` in the next if condition.

src/or/hs_service.c
```diff
@@ -1638,6 +1711,20 @@ build_service_desc_keys(const hs_service_t *service,
...
   }
 
+  /* No need for extra strong, this is a temporary key only for this
+   * descriptor. Nothing long term. */
+  if (curve25519_keypair_generate(&desc->auth_ephemeral_kp, 0) < 0) {
```
dgoulet-tor: Do we need to create that key if we have _no_ client authorization?

src/or/hs_service.c
```diff
@@ -1638,6 +1711,20 @@ build_service_desc_keys(const hs_service_t *service,
...
+    log_warn(LD_REND, "Can't generate auth ephemeral keypair for "
+                      "service %s",
+             safe_str_client(service->onion_address));
+    ret = -1;
+  }
+
+  /* Random a descriptor cookie to be used as a part of a key to encrypt the
+   * descriptor. */
+  crypto_strongest_rand(desc->descriptor_cookie,
+                        sizeof(desc->descriptor_cookie));
```
dgoulet-tor: Same question as the above?

src/common/crypto_format.c
```diff
@@ -162,6 +162,44 @@ curve25519_public_from_base64(curve25519_public_key_t *pkey,
...
+  } else if (len == CURVE25519_BASE64_PADDED_LEN) {
+    char buf[128];
+    if (base64_decode(buf, sizeof(buf), input, len) != CURVE25519_SECKEY_LEN)
+      return -1;
+    memcpy(skey->secret_key, buf, CURVE25519_SECKEY_LEN);
+    return 0;
+  } else {
+    return -1;
+  }
+}
```
dgoulet-tor: There is probably a time we will want to merge those with the public key as well since they have the same exact code except with different objects. But we can do that later.

src/or/hs_descriptor.c
```diff
@@ -1179,6 +1179,36 @@ desc_encode_v3(const hs_descriptor_t *desc,
 
 /* === DECODING === */
 
+/* Given the token tok for an auth client, decode it as
+ * hs_desc_authorized_client_t. Return 0 on success else -1 on failure. */
```
dgoulet-tor: We should document here that the tok->args MUST contain at least 3 elements. The following function doesn't assess that so lets be extra careful.

src/or/hs_service.c
```diff
@@ -1319,8 +1321,91 @@ service_descriptor_new(void)
...
+  tor_assert(config1->clients);
+  tor_assert(config2->clients);
+
+  /* If the number of clients is different, it is obvious that the list
+   * changes. */
+  if (smartlist_len(config1->clients) != smartlist_len(config2->clients)) {
+    goto done;
+  }
+
+  /* An array that has the same lenght as the list of clients in config2.
```
dgoulet-tor: `length`

src/or/hs_service.c
```diff
@@ -1319,8 +1321,91 @@ service_descriptor_new(void)
...
+/* If two authorized clients are equal, return 1. Otherwise, return 0. */
+static int
+service_authorized_client_equal(const hs_service_authorized_client_t *client1,
+                                const hs_service_authorized_client_t *client2)
+{
+  tor_assert(client1);
+  tor_assert(client2);
+
+  /* If the client names are not equal, the whole structs are not equal. */
+  if (strcmp(client1->client_name, client2->client_name)) {
```
dgoulet-tor: Can those be NULL at this point?

src/or/hs_config.c
```diff
@@ -266,6 +265,94 @@ config_service_v3(const config_line_t *line_,
...
+        client->client_name = tor_strdup(client_name);
+        smartlist_add(config->clients, client);
+        log_debug(LD_REND, "Adding client name '%s'", client_name);
+      }
+      SMARTLIST_FOREACH_END(client_name);
+      SMARTLIST_FOREACH(clients, char *, cp, tor_free(cp));
+      smartlist_free(clients);
+      /* Ensure maximum number of clients. */
+      if (config->client_auth_type == HS_CLIENT_AUTH_TYPE_BASIC &&
+          smartlist_len(config->clients) > 512) {
```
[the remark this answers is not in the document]
haxxpop: In fact, I just copied this validation from the v2, we can remove it if we don't want it in v3 :)

src/or/hs_common.h
```diff
@@ -141,6 +141,12 @@ typedef enum {
   RSAE_OKAY        = 0   /**< Service added as expected */
 } hs_service_add_ephemeral_status_t;
 
+/* Type of client authorization. */
+typedef enum {
+  HS_CLIENT_AUTH_TYPE_NULL = 0,
+  HS_CLIENT_AUTH_TYPE_BASIC = 1,
```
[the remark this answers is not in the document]
haxxpop: We had so much discussion since I wrote this code. Yes, we need to change this thing! I think instead of specifying the auth type like this we should have just a flag indicating if the auth is enabled or not. (because in this PR we assume that only desc auth is available)

src/or/hs_service.h
```diff
@@ -172,6 +178,14 @@ typedef struct hs_service_config_t {
    * HiddenServiceNumIntroductionPoints option. */
   unsigned int num_intro_points;
 
+  /* Client authorization type that this service performs. Specified by
+   * HiddenServiceAuthorizeClient option. */
+  hs_client_auth_type_t client_auth_type;
```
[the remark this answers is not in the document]
haxxpop: I already changed this one to the flag indicating if the client auth is enabled or not already. So I think this is not a problem anymore :)

src/or/hs_service.c
haxxpop: The spec says that
```
     "desc-auth-ephemeral-key" SP key NL

      [Exactly once]

      This field contains an ephemeral x25519 public key generated by the
      hidden service and encoded in base64. The key is used by the encryption
      scheme below.

      If client authorization is disabled, the value here should be a fresh
      x25519 pubkey that will remain unused.
```

src/or/hs_service.c
haxxpop: No :) I will add an if statement here

src/or/hs_descriptor.c
```diff
@@ -691,69 +718,70 @@ encrypt_descriptor_data(const hs_descriptor_t *desc, const char *plaintext,
...
-
-  { /* Get those fakes! */
-    uint8_t client_id[8]; /* fake client-id */
-    uint8_t iv[16]; /* fake IV (initialization vector) */
-    uint8_t encrypted_cookie[16]; /* fake encrypted cookie */
-
-    FILL_WITH_FAKE_DATA_AND_BASE64(client_id);
-    FILL_WITH_FAKE_DATA_AND_BASE64(iv);
-    FILL_WITH_FAKE_DATA_AND_BASE64(encrypted_cookie);
-  }
```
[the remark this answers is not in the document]
haxxpop: We have a brand new `hs_desc_build_fake_authorized_client`. Could you please look at it if it answers your question?

src/or/hs_descriptor.c
```diff
@@ -691,69 +718,70 @@ encrypt_descriptor_data(const hs_descriptor_t *desc, const char *plaintext,
...
-     does is make fake clients */
-  int i = 0;
   smartlist_t *auth_client_lines = smartlist_new();
   char *auth_client_lines_str = NULL;
 
-  /* Make a line for each fake client */
-  const int num_fake_clients = CLIENT_AUTH_ENTRIES_BLOCK_SIZE;
-  for (i = 0; i < num_fake_clients; i++) {
-    char *auth_client_str = get_fake_auth_client_str();
-    tor_assert(auth_client_str);
```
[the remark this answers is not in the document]
haxxpop: The same answer above.

src/or/hs_service.c
haxxpop: No, but I will add the assert for both.

haxxpop:
Sample keypairs
(1)
descriptor:x25519:dz4q5xqlb4ldnbs72iarrml4ephk3du4i7o2cgiva5lwr6wkquja
4acth47i6kxnvkewtm6q7ib2s3ufpo5sqbsnzjpbi7utijcltosqemad:descriptor:x25519:zdsyvn2jq534ugyiuzgjy4267jbtzcjbsgedhshzx5mforyxtryq
(2)
descriptor:x25519:okoi2gml3wd6x7jganlk5d66xxyjgg24sxw4y7javx4giqr66zta
25njqamcweflpvkl73j4szahhihoc4xt3ktcgjnpaingr5yhkenl5sid:descriptor:x25519:fdreqzjqso7d2ac7qscrxfl5qfpamdvgy5d6cxejcgzc3hvhurmq

Don't forget to change the onion addresses. Those are just samples.

src/feature/hs/hs_client.c
```diff
@@ -1563,42 +1563,51 @@ hs_config_client_authorization(const or_options_t *options,
...
-      client_key_str = read_file_to_str(client_key_file_path, 0, NULL);
-      /* Free the file path immediately after using it. */
-      tor_free(client_key_file_path);
-
-      /* If we cannot read the file, continue with the next file. */
-      if (!client_key_str) {
-        continue;
-      }
+    if (!auth_key_filename_is_valid(filename)) {
+      log_warn(LD_REND, "The filename is invalid.");
```
dgoulet-tor: I tried to turn off a client authorization without removing the file by appending `.off` to the file name but then this warning was triggered.

I propose we put it at notice level and we say that we ignored the file instead so the user is aware.

src/feature/hs/hs_client.c
```diff
@@ -1563,42 +1563,51 @@ hs_config_client_authorization(const or_options_t *options,
...
-        /* Parse the onion address to get an identity public key and use it
-         * as a key of global map in the future. */
-        if (hs_parse_address(auth->onion_address, &identity_pk,
-                             NULL, NULL) < 0) {
-          client_service_authorization_free(auth);
-          continue;
-        }
+    /* If we cannot read the file, continue with the next file. */
+    if (!client_key_str) {
+      log_warn(LD_REND, "The file cannot be read.");
```
dgoulet-tor: Lets print the filename and mention that this is client authorization in `key_dir`

src/feature/hs/hs_client.c
```diff
@@ -1563,42 +1563,51 @@ hs_config_client_authorization(const or_options_t *options,
...
+    /* Free immediately after using it. */
+    tor_free(client_key_str);
+
+    if (auth) {
+      /* Parse the onion address to get an identity public key and use it
+       * as a key of global map in the future. */
+      if (hs_parse_address(auth->onion_address, &identity_pk,
+                           NULL, NULL) < 0) {
+        client_service_authorization_free(auth);
+        log_warn(LD_REND, "The onion address is invalid.");
```
dgoulet-tor: In this case, the .onion is invalid that is can't exists or work... so lets print it so the user can notice in which auth file it is.

src/feature/hs/hs_client.c
```diff
@@ -1563,42 +1563,51 @@ hs_config_client_authorization(const or_options_t *options,
...
-          log_warn(LD_REND, "Duplicate authorization for the same hidden "
-                            "service.");
-          goto end;
-        }
+      if (digest256map_get(auths, identity_pk.pubkey)) {
+        client_service_authorization_free(auth);
 
-        digest256map_set(auths, identity_pk.pubkey, auth);
+        log_warn(LD_REND, "Duplicate authorization for the same hidden "
+                          "service.");
```
dgoulet-tor: Again, I would print the .onion.  A bit annoying to encode address just for the logs but worth it I believe for the sake of our users.

src/feature/hs/hs_service.c
```diff
@@ -1135,23 +1135,31 @@ parse_authorized_client(const char *client_key_str)
...
   auth_type = smartlist_get(fields, 0);
   key_type = smartlist_get(fields, 1);
   pubkey_b32 = smartlist_get(fields, 2);
 
-  /* Currently, the only supported auth type is "descriptor" and the only
-   * supported key type is "x25519". */
-  if (strcmp(auth_type, "descriptor") || strcmp(key_type, "x25519")) {
+  /* Currently, the only supported auth type is "descriptor". */
+  if (strcmp(auth_type, "descriptor")) {
+    log_warn(LD_REND, "The auth type '%s' is not supported.", auth_type);
```
dgoulet-tor: I would add in which file this is.

src/feature/hs/hs_service.c
```diff
@@ -1135,23 +1135,31 @@ parse_authorized_client(const char *client_key_str)
...
+    goto err;
+  }
+
+  /* Currently, the only supported key type is "x25519". */
+  if (strcmp(key_type, "x25519")) {
+    log_warn(LD_REND, "The key type '%s' is not supported.", key_type);
```
dgoulet-tor: Same as above, add in which file this is.

src/feature/hs/hs_client.c
```diff
@@ -1219,11 +1232,19 @@ hs_client_decode_descriptor(const char *desc_str,
...
+  curve25519_secret_key_t *client_sk = NULL;
 
   tor_assert(desc_str);
   tor_assert(service_identity_pk);
   tor_assert(desc);
 
+  /* Check if we have a client authorization for this service in the map. */
+  client_auth = find_client_auth(service_identity_pk);
+  if (client_auth) {
+    client_sk = &client_auth->enc_seckey;
```
dgoulet-tor: Seems this is the place where a client knows if it will use client auth or not. I would add a `log_info()` that says "Ok we'll used client authorization for .onion X".

This kind of log will help us greatly in the future to debug issues client side.

src/feature/hs/hs_client.c
```diff
@@ -1233,7 +1254,7 @@ hs_client_decode_descriptor(const char *desc_str,
   }
 
   /* Parse descriptor */
-  ret = hs_desc_decode_descriptor(desc_str, subcredential, desc);
+  ret = hs_desc_decode_descriptor(desc_str, subcredential, client_sk, desc);
   memwipe(subcredential, 0, sizeof(subcredential));
   if (ret < 0) {
     log_warn(LD_GENERAL, "Could not parse received descriptor as client.");
```
dgoulet-tor: Client *without* any client authorization configured but the service descriptor requires authorization. This means, the client won't be able to decrypt the descriptor and thus ending up here.

We should first bring down that log to `log_info()` in my opinion. This is not something the user needs to know at the warning level, nothing to be concerned about I would say.

Second, and we could do that after the merge, would be to make the error code from `hs_desc_decode_descriptor()` a bit more useful like "Ok descriptor wasn't decryptable" which could make the client print a log message like "We probably do not have access to A.onion. Enable client authorization" kind of thing. And *maybe*, we could also stop the client from refetching a new descriptor from a new HSDir...

src/feature/hs/hs_client.c
```diff
@@ -1393,6 +1397,212 @@ hs_client_receive_rendezvous_acked(origin_circuit_t *circ,
...
+    goto err;
+  }
+
+  auth = tor_malloc_zero(sizeof(hs_client_service_authorization_t));
+  if (base32_decode((char *) auth->enc_seckey.secret_key,
+                    sizeof(auth->enc_seckey.secret_key),
+                    seckey_b32, strlen(seckey_b32)) < 0) {
+    goto err;
+  }
+  memcpy(auth->onion_address, onion_address, HS_SERVICE_ADDR_LEN_BASE32+1);
```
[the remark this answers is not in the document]
dgoulet-tor: Fixup in my branch `ticket20700_035_01` : 3456fa36cc

PR #41, opened 2018-04-05
Postloop callbacks 2

nmathewson:


dgoulet-tor:
So this all lgtm! The only thing I think can be done as an extra commit is to remove the `tell_event_loop_to_run_external_code()` code since it is not used anymore.

PR #43, opened 2018-04-06
Lift the list of default directory servers into their own file.

ahf:
This patch lifts the list of default directory authorities from config.c
into their own auth_dirs.inc file, which is then included in config.c
using the C preprocessor.

Patch by beastr0.

See: https://bugs.torproject.org/24854


teor2345:
This was merged to 0.2.9 and later

PR #45, opened 2018-04-09
Remove old consensus methods 2018

nmathewson:


src/test/test_dir.c
```diff
@@ -3014,7 +3016,7 @@ gen_routerstatus_for_umbw(int idx, time_t now)
   if (vrs) {
     vrs->microdesc = tor_malloc_zero(sizeof(vote_microdesc_hash_t));
     tor_asprintf(&vrs->microdesc->microdesc_hash_line,
-                 "m 9,10,11,12,13,14,15,16,17 "
+                 "m 13,14,15,16,17,18,19,20,21,22,23,24,25,26,27,28 "
```
teor2345: Should this list be 25-28?
21 is not a supported method, should it be added to the list?
After this branch, 13-24 aren't supported, should they be in the list?

src/test/test_dir.c
```diff
@@ -3040,7 +3042,7 @@ vote_tweaks_for_umbw(networkstatus_t *v, int voter, time_t now)
   smartlist_clear(v->supported_methods);
   /* Method 17 is MIN_METHOD_TO_CLIP_UNMEASURED_BW_KB */
   smartlist_split_string(v->supported_methods,
-                         "1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17",
+                         "13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28",
```
teor2345: Should this list also be 25-28?
(See my previous comment)

src/or/networkstatus.c
```diff
@@ -1528,7 +1528,7 @@ networkstatus_consensus_has_ipv6(const or_options_t* options)
     return
        cons->consensus_method >= MIN_METHOD_FOR_A_LINES_IN_MICRODESC_CONSENSUS;
   } else {
-    return cons->consensus_method >= MIN_METHOD_FOR_A_LINES;
+    return 1;
```
teor2345: I think it's ok for us to modify this client code. As of 1 May 2018, we will only support 0.2.9, which creates consensuses with method 25 and later.

And this is edge-case code. It is only useful:
* on networks with consensus methods below 13 (no IPv6 addresses in the full consensus),
* on clients using full descriptors (IPv6 addresses in descriptors),
* with IPv6-supporting relays,
* and IPv6-supporting fallbacks,
* with an IPv6-only client.

src/test/test_dir.c
```diff
@@ -5771,22 +5771,10 @@ test_dir_networkstatus_consensus_has_ipv6(void *arg)
...
-  mock_networkstatus->consensus_method = MIN_METHOD_FOR_A_LINES;
+  mock_networkstatus->consensus_method = MIN_SUPPORTED_CONSENSUS_METHOD;
   has_ipv6 = networkstatus_consensus_has_ipv6(get_options());
   tt_assert(has_ipv6);
 
-  mock_networkstatus->consensus_method = MIN_METHOD_FOR_A_LINES + 1;
-  has_ipv6 = networkstatus_consensus_has_ipv6(get_options());
-  tt_assert(has_ipv6);
-
-  mock_networkstatus->consensus_method = MIN_METHOD_FOR_A_LINES + 20;
```
teor2345: Please don't delete this one test.
Instead, replace `MIN_METHOD_FOR_A_LINES + 20` with `MIN_SUPPORTED_CONSENSUS_METHOD + 20`.

src/test/test_dir.c
nmathewson: 8420898a87f58ddfba28da477465b2f0859870b1 will fix this too.

src/test/test_dir.c
nmathewson: 8420898a87f58ddfba28da477465b2f0859870b1 will fix this

src/test/test_dir.c
nmathewson: I think MAX_SUPPORTED_CONSENSUS_METHOD (for a far-future-method) would be even better?

Going with that in 427732a5d113d190a321819f30cb90aed5b054e9

nmathewson:
This was squashed and merged as 0e8ae82a8757fb8bb2850f1a3d59ad1dfb468556

PR #46, opened 2018-04-10
Token bucket refactor

nmathewson:


src/common/token_bucket.c
```diff
@@ -0,0 +1,172 @@
...
+ * Initialize a token bucket in *<b>bucket</b>, set up to allow <b>rate</b>
+ * bytes per second, with a maximum burst of <b>burst</b> bytes. The bucket
+ * is created such that <b>now_ts</b> is the current timestamp.  The bucket
+ * starts out full.
+ */
+void
+token_bucket_init(token_bucket_t *bucket,
+                  uint32_t rate,
+                  uint32_t burst,
+                  uint32_t now_ts)
```
dgoulet-tor: Within the DoS subsystem, I did something extremely similar (of course bucketting...) but I went for `uint64_t` on purposes so we could address just all use cases in the future. Turned out that we could just have `uint32_t` instead since we use consensus param for some values which are bound to be `int32_t`.

You think we could go to 64 bit there or memory footprint is the idea of 32 bit?

src/common/token_bucket.c
```diff
@@ -0,0 +1,172 @@
...
+ **/
+
+#define TOKEN_BUCKET_PRIVATE
+
+#include "token_bucket.h"
+#include "util_bug.h"
+
+/** Convert a rate in bytes per second to a rate in bytes per step */
+static uint32_t
+rate_per_sec_to_rate_per_step(uint32_t rate)
```
dgoulet-tor: I think we should `inline` this for the small tiny bit of performance we get. Token bucket like that are in fast path often.

src/common/token_bucket.c
```diff
@@ -0,0 +1,172 @@
...
+  tor_assert_nonfatal(burst > 0);
+  if (burst > TOKEN_BUCKET_MAX_BURST)
+    burst = TOKEN_BUCKET_MAX_BURST;
+
+  bucket->rate = rate_per_sec_to_rate_per_step(rate);
+  bucket->burst = burst;
+  if (bucket->read_bucket > (int32_t)burst)
+    bucket->read_bucket = burst;
+  if (bucket->write_bucket > (int32_t)burst)
+    bucket->write_bucket = burst;
```
dgoulet-tor: Could we do a `= MIN(bucket->read_bucket, burst)` here instead? Would just make things a bit clearer in terms of code semantic I believe.

src/common/token_bucket.c
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/* Helper: see token_bucket_refill */
+static int
+refill_single_bucket(int32_t *bucketptr,
+                     const uint32_t rate,
+                     const int32_t burst,
+                     const uint32_t elapsed_steps)
+{
+  const int was_empty = *bucketptr <= 0;
```
dgoulet-tor: Huge nitpick but for my brain to parse, having `(*bucketptr <= 0)` would be great.

src/common/token_bucket.c
```diff
@@ -0,0 +1,172 @@
...
+  /* The casts here prevent an underflow. */
+  const size_t gap = ((size_t)burst) - ((size_t)*bucketptr);
```
dgoulet-tor: `bucketptr = -1` and `burst = INT32_MAX` we wned up with `gap = INT32_MAX + 1` which means we can go above the maximum burst value for the `gap`. It is fine since we normalize it to the "burst" value after but I think its worth commenting that this can happen and it is ok.

src/common/token_bucket.c
```diff
@@ -0,0 +1,172 @@
...
+ * is <b>now_ts</b>.
+ *
+ * Return a bitmask containing TB_READ iff read bucket was empty and became
+ * nonempty, and TB_WRITE iff the write bucket was empty and became nonempty.
+ */
+int
+token_bucket_refill(token_bucket_t *bucket,
+                    uint32_t now_ts)
+{
+  const uint32_t elapsed_ticks = (now_ts - bucket->last_refilled_at_ts);
```
dgoulet-tor: As the commit messages mentions it, it is meant to be use with monotonic time only so then `now_ts` will never be smaller than `last_refilled_at_ts`. Can we make sure of this?

We had issues in the scheduler with Windows not really having monotonic clock so we had to adjust to this reality (see `kist_scheduler_schedule()`.

src/common/token_bucket.c
```diff
@@ -0,0 +1,172 @@
...
+                           bucket->rate, bucket->burst, elapsed_steps))
+    flags |= TB_WRITE;
+
+  bucket->last_refilled_at_ts = now_ts;
+  return flags;
+}
+
+static int
+decrement_single_bucket(int32_t *bucketptr,
+                        ssize_t n)
```
dgoulet-tor: Why `ssize_t` and not `int32_t` which is what the counters uses?

src/common/token_bucket.h
```diff
@@ -0,0 +1,72 @@
...
+#ifndef TOR_TOKEN_BUCKET_H
+#define TOR_TOKEN_BUCKET_H
+
+#include "torint.h"
+
+typedef struct token_bucket_t {
+  uint32_t rate;
+  int32_t burst;
+  int32_t read_bucket;
+  int32_t write_bucket;
```
dgoulet-tor: Is more of a design questions.

I would love to use that in the DoS subsystem which would mean removing bunch of codes. However, this `token_bucket_t` interfaces kind of bakes the fact that you have "read" and "write" thus a connection.

If that struct had only one bucket, we could use it in some other places in a more generic way. Which would mean we would have `token_bucket_t read_bucket` and the write one in the `or_connection_t`. Is that a major inconvenience? Because as it is, I can use that in DoS code but not really semantically.

src/or/main.c
```diff
@@ -2418,21 +2418,23 @@ refill_callback(periodic_timer_t *timer, void *arg)
...
+    token_bucket_get_read(&global_bucket);
 
   stats_n_bytes_read += bytes_read;
   stats_n_bytes_written += bytes_written;
   if (accounting_is_enabled(options) && milliseconds_elapsed >= 0)
     accounting_add_bytes(bytes_read, bytes_written, seconds_rolled_over);
 
   if (milliseconds_elapsed > 0) {
-    connection_bucket_refill(milliseconds_elapsed, (time_t)now.tv_sec,
+    connection_bucket_refill((time_t)now.tv_sec,
```
dgoulet-tor: `refill_callback()` doesn't use monotonic time unless we do adjustments in the previous commit I pointed it out.

src/common/token_bucket.c
```diff
@@ -0,0 +1,172 @@
+/* Copyright (c) 2003-2004, Roger Dingledine
+ * Copyright (c) 2004-2006, Roger Dingledine, Nick Mathewson.
+ * Copyright (c) 2007-2017, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
+ * \file address.c
```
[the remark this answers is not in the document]
nmathewson: or even `token_bucket.c` . ;)
5ce78468cf1f74 has a fix.

src/common/token_bucket.c
nmathewson: Hm.  So, the existing code here assumes it's 32-bit.

Do we ever need to allow a burst of greater than 4GB?  Nobody has actually asked for it so far.  And since the API returns a size_t, it should be comparatively easy to make this bigger later on if we want it.

src/common/token_bucket.c
nmathewson: I think the compiler should be able to figure out that this function should be inlined on its own.  (In my experiments, both gcc and clang inline it.)

Also, note that this function is only used when setting the rate -- not when actually using the buckets.

src/common/token_bucket.c
nmathewson: c752a23eedf31a7 has a fix here.

src/common/token_bucket.c
nmathewson: Done with 3deeec38b7ffcd51f97ecea7bda37902fc9eca33

src/common/token_bucket.c
nmathewson: 4fc779ca1cc19af86d494d6c87913c0322c58131 is a comment rewrite for this.

src/common/token_bucket.c
nmathewson: I've done something similar here with 2c4eef1fa825a1

src/common/token_bucket.c
nmathewson: So that the API is agnostic to the implementation details.

src/common/token_bucket.h
nmathewson: This is what I was asking you about earlier: if we make the change you suggest here, we have to include the `rate`, `burst`, and `last_refilled_at` fields in the or_connection_t twice.

I can try a refactoring to make this change, but I'd rather do it as a separate branch after this is merged, if that's okay with you.

There are also couple of issues with this implementation that make it a little tricky to reuse. Most notably, it assumes that it's working in units of bytes per second, but it converts them to "bytes per step".  That conversion is fine for bandwidths, but the imprecesion might not be okay for all other units: if so, we'd have to configure it.

src/common/token_bucket.c
dgoulet-tor: Ok, after seeing how it is used in the code and tightly coupled with torrc options that are int32_t, this is fine.

src/common/token_bucket.h
dgoulet-tor: Right, I think I realized what you meant when you asked when seeing this.

Hmmm... and it makes sense also with our current options since we only have a single bw rate and not a read bw rate and write bw rate.

If we could make this whole piece of work in a generic way to have token bucket for other things than "bytes per second", it would be really great so we don't have multiple bucket implementation. 

I'm not that concerned about having duplicate values that is 4 bytes (rate), 4 bytes (burst) and 4 bytes (timetamp) which means 16 bytes extra. And scaled to 100k connections let say, it is overhead of ~2MB which for a relay is not that much compared to a mobile client.

An option then is to configure a "bucket param" object statically and attach it to any bucket you create with it. The DoS subsystem does not copy rate/burst to the bucket object and uses the global value each time instead. If we had a way to do this with a "param" object, it would also work out pretty well. Which also could reduce by 8 bytes (rate + burst) the `token_bucket_t` object and make it only have a pointer to a static object (+ 4 bytes). Anyway, we aren't saving much space here either way but what I'm more wanting here is a more generic API so it can be used elsewhere. I know that we'll have to go to hidden service circuit creation rate limitation at some point which a bucket system would be _great_ to have.

src/common/token_bucket.c
```diff
@@ -60,10 +60,8 @@ token_bucket_adjust(token_bucket_t *bucket,
 
   bucket->rate = rate_per_sec_to_rate_per_step(rate);
   bucket->burst = burst;
-  if (bucket->read_bucket > (int32_t)burst)
-    bucket->read_bucket = burst;
-  if (bucket->write_bucket > (int32_t)burst)
-    bucket->write_bucket = burst;
+  bucket->read_bucket = MIN(bucket->read_bucket, (int32_t)burst);
+  bucket->read_bucket = MIN(bucket->write_bucket, (int32_t)burst);
 }
```
dgoulet-tor: `write_bucket` for the second.

nmathewson:
This was squashed and merged with b152d62cee7480ee7b9b68dd9b619db65b6cd112

PR #47, opened 2018-04-13
Token bucket once again

nmathewson:


src/common/token_bucket.h
```diff
@@ -11,15 +11,64 @@
...
+
+/** A raw token bucket, decoupled from its configuration and timestamp. */
+typedef struct token_bucket_raw_t {
+  int32_t bucket;
+} token_bucket_raw_t;
+
+/** A timestamp for a token bucket. The units of this timestamp are
+ * unspecified, but must match with the rate set in the token_bucket_cfg_t. */
+typedef struct token_bucket_timestamp_t {
+  uint32_t last_refilled_at;
```
dgoulet-tor: If this is a Unix timestamp, this means we are stuck in 2038 ... :S  Could we go to uint64_t to avoid that?

src/common/token_bucket.c
```diff
@@ -20,6 +21,51 @@
 #include "token_bucket.h"
 #include "util_bug.h"
 
+/**
+ *Set the <b>rate</b> and <b>burst</b> value in a token_bucket_cfg.
+ *
```
dgoulet-tor: Whitespace missing.

src/common/token_bucket.c
```diff
@@ -20,6 +21,51 @@
...
+
+/**
+ * Initialize a raw token bucket and its associated timestamp to the "full"
+ * state, according to <b>cfg</b>.
+ */
+void
+token_bucket_raw_reset(token_bucket_raw_t *bucket,
+                       token_bucket_timestamp_t *stamp,
+                       const token_bucket_cfg_t *cfg,
+                       uint32_t now)
```
dgoulet-tor: Again, if `now` is a Unix timestamp, we are a bit stuck in 2038.

src/common/token_bucket.c
```diff
@@ -57,55 +103,56 @@ token_bucket_rw_init(token_bucket_rw_t *bucket,
...
-                     const int32_t burst,
-                     const uint32_t elapsed_steps)
+/**
+ * Given an amount of <b>elapsed</b> time units, and a bucket configuration
+ * <b>cfg</b>, refill the level of <b>bucket</b> accordingly.  Note that the
+ * units of time in <b>elapsed</b> must correspond to those used to set the
+ * rate in <b>cfg</b>, or the result will be illogical.
+ */
+int
+token_bucket_raw_refill_steps(token_bucket_raw_t *bucket,
```
dgoulet-tor: This can be made static.

nmathewson: No; it should be public, so that others can use token_bucket_raw.

src/common/token_bucket.c
nmathewson: Fixed in 1be024798e2e2bac499acf8106eb215e3c4c130b

src/common/token_bucket.c
nmathewson: That is to say, the idea of the refactoring here is that the token_bucket_raw() functions can be used by outside code to build token buckets that aren't specifically tied to token_bucket_rw_t.

nmathewson:
This was squashed and merged with 34c2574aa97a05381c1796fb12d82616f96fdeb6

PR #48, opened 2018-04-16
25691/25692: Distinguish "has preferred descriptor" from "has any descriptor"

nmathewson:


src/or/hs_common.c
```diff
@@ -1281,7 +1281,7 @@ node_has_hsdir_index(const node_t *node)
 
   /* A node can't have an HSDir index without a descriptor since we need desc
    * to get its ed25519 key */
-  if (!node_has_descriptor(node)) {
+  if (!node_has_preferred_descriptor(node, 0)) {
```
teor2345: Note: Using for_direct_connect = 0 is always correct for the HSDir hash ring, because we use non-bridge descriptors for the HSDir hash ring. This is true even though Tor2web connects directly to HSDirs.

teor2345: Hmm, does this need a comment?

src/or/hs_common.c
```diff
@@ -1617,7 +1617,7 @@ hs_pick_hsdir(smartlist_t *responsible_dirs, const char *req_key_str)
     time_t last = hs_lookup_last_hid_serv_request(dir, req_key_str, 0, 0);
     const node_t *node = node_get_by_id(dir->identity_digest);
     if (last + hs_hsdir_requery_period(options) >= now ||
-        !node || !node_has_descriptor(node)) {
+        !node || !node_has_preferred_descriptor(node, 0)) {
```
teor2345: Does this need a comment about Tor2web, similar to the last comment?

src/or/nodelist.c
```diff
@@ -1139,6 +1140,27 @@ node_has_descriptor(const node_t *node)
           (node->rs && node->md));
 }
 
+/** DOCDOC */
```
teor2345: Please document the function

src/or/nodelist.c
```diff
@@ -1139,6 +1140,27 @@ node_has_descriptor(const node_t *node)
...
+{
+  const int is_bridge = node_is_a_configured_bridge(node);
+  const int we_use_mds = we_use_microdescriptors_for_circuits(get_options());
+
+  if ((is_bridge && for_direct_connect) || !we_use_mds) {
+    /* We need an ri in this case. */
+    if (!node->ri)
+      return 0;
+  } else {
+    /* Otherwise we need an md. */
```
teor2345: An rs and an md?

src/or/rendservice.c
```diff
@@ -3596,7 +3596,7 @@ directory_post_to_hs_dir(rend_service_descriptor_t *renddesc,
         /* Don't upload descriptor if we succeeded in doing so last time. */
         continue;
       node = node_get_by_id(hs_dir->identity_digest);
-      if (!node || !node_has_descriptor(node)) {
+      if (!node || !node_has_preferred_descriptor(node,0)) {
```
teor2345: Does this need a comment about Tor2web, similar to the previous comments?

src/or/routerlist.c
```diff
@@ -2335,7 +2335,7 @@ router_add_running_nodes_to_smartlist(smartlist_t *sl, int need_uptime,
   SMARTLIST_FOREACH_BEGIN(nodelist_get_list(), const node_t *, node) {
     if (!node->is_running || !node->is_valid)
       continue;
-    if (need_desc && !(node->ri || (node->rs && node->md)))
+    if (need_desc && !node_has_preferred_descriptor(node, direct_conn))
```
teor2345: Oh, good catch. I remember writing this code. I am glad that we are refactoring it and fixing it.

src/or/circuitbuild.c
```diff
@@ -2464,7 +2464,7 @@ count_acceptable_nodes, (smartlist_t *nodes))
     if (! node->is_valid)
 //      log_debug(LD_CIRC,"Nope, the directory says %d is not valid.",i);
       continue;
-    if (! node_has_descriptor(node))
+    if (! node_has_any_descriptor(node))
```
teor2345: I think this code is correct, but it's not obvious:
1. new_route_len() is the only caller
2. new_route_len() will fail if count_acceptable_nodes() is less than the route length.
3. later functions will fail if there is no guard
4. later functions will fail if there are not enough subsequent nodes for the rest of the route

So this check can spuriously succeed when we have no guards, or when we have many bridges, and no subsequent nodes. But it can never fail when there are actually enough nodes for the path. So this is just an optimisation to stop us trying to build lots of circuits when we have no descriptors.

src/or/control.c
```diff
@@ -3498,7 +3498,7 @@ handle_control_extendcircuit(control_connection_t *conn, uint32_t len,
       connection_printf_to_buf(conn, "552 No such router \"%s\"\r\n", n);
       goto done;
     }
-    if (!node_has_descriptor(node)) {
+    if (!node_has_any_descriptor(node)) {
```
teor2345: The log message at line 3525 is now incorrect:
```
                "controller tried to connect to a node that doesn't have any "
                "addresses that are allowed by the firewall configuration; "
                "circuit marked for closing."
```

We can fail because we are missing the preferred descriptor for the node, too.
Should we do `int first_node = zero_circ;` `first_node = 0;` in this loop, too?

src/or/nodelist.c
```diff
@@ -1134,7 +1134,7 @@ node_is_dir(const node_t *node)
 /** Return true iff <b>node</b> has either kind of usable descriptor -- that
  * is, a routerdescriptor or a microdescriptor. */
```
teor2345: Comment: You should probably use node_has_preferred_descriptor() instead of this function.

src/or/circuitbuild.c
teor2345: We could return the number that pass for_direct_connect = 1 and for_direct_connect = 0, but that seems unnecessarily complicated.

src/or/nodelist.c
```diff
@@ -2243,7 +2243,8 @@ compute_frac_paths_available(const networkstatus_t *consensus,
               nu);
 
     SMARTLIST_FOREACH_BEGIN(myexits_unflagged, const node_t *, node) {
-      if (node_has_descriptor(node) && node_exit_policy_rejects_all(node)) {
+      if (node_has_any_descriptor(node) &&
```
teor2345: count_usable_descriptors() counts mds if we have an md consensus, and ris if we have an ns consensus.
So I think we want node_has_preferred_descriptor(node, 0) here.

src/or/routerlist.c
```diff
@@ -2758,15 +2758,15 @@ frac_nodes_with_descriptors(const smartlist_t *sl,
       total <= 0.0) {
     int n_with_descs = 0;
     SMARTLIST_FOREACH(sl, const node_t *, node, {
-      if (node_has_descriptor(node))
+      if (node_has_any_descriptor(node))
```
teor2345: We should pass for_direct_conn into this function, and use node_has_preferred_descriptor().

Then we won't bootstrap unless we have enough actual mid and exit bandwidth, even if we have mids or exits listed as our bridges. (The guard case is unchanged for non-bridge clients.)

The bridge client case could be tricky, because:
1. compute_frac_paths_available() only checks guard-flagged nodes, not bridges
2. even if it did check bridges, they don't have bandwidths
3. even if we used a weight of 1 for each bridge, we don't require 65% of bridges to be up to bootstrap

To workaround this issue, I suggest we make f_guard = 1.0 in compute_frac_paths_available() if we are using bridges, and have at least one bridge with the preferred descriptor.

src/test/test_hs_common.c
```diff
@@ -305,7 +305,7 @@ helper_add_hsdir_to_networkstatus(networkstatus_t *ns,
   node_t *node = node_get_mutable_by_id(ri->cache_info.identity_digest);
   tt_assert(node);
   node->rs = rs;
-  /* We need this to exist for node_has_descriptor() to return true. */
+  /* We need this to exist for node_has_any_descriptor() to return true. */
```
teor2345: hsdirs use node_has_preferred_descriptor(), so I think this comment is wrong

src/or/routerlist.c
teor2345: This change might cause bootstrapping issues. Let's defer it to 0.3.4 in another ticket?

src/or/circuitbuild.c
```diff
@@ -2857,17 +2857,8 @@ extend_info_from_node(const node_t *node, int for_direct_connect)
...
-
-  if ((is_bridge && for_direct_connect) || !we_use_mds) {
-    /* We need an ri in this case. */
-    if (!node->ri)
-      return NULL;
-  } else {
-    /* Otherwise we need an md. */
-    if (node->rs == NULL || node->md == NULL)
-      return NULL;
+  if (!node_has_preferred_descriptor(node, for_direct_connect)) {
```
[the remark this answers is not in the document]
nmathewson: I think we do.  Done with 1d06778e696061.

src/or/entrynodes.c
```diff
@@ -192,7 +192,7 @@ guard_has_descriptor(const entry_guard_t *guard)
   const node_t *node = node_get_by_id(guard->identity);
   if (!node)
     return 0;
-  return node_has_descriptor(node);
+  return node_has_preferred_descriptor(node, 1);
```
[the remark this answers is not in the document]
nmathewson: cea472f2dcce436718eefe601e49ebe000805853 fixes this.

src/or/hs_common.c
nmathewson: 5398bf02fea4a5 has that comment.

src/or/hs_common.c
nmathewson: 03494da6e0be1b0208804850298b2d035a830228 is the fixup for this.

src/or/nodelist.c
nmathewson: 80ac049b6f3d92 does that.

src/or/nodelist.c
nmathewson: Yes. 80ac049b6f3d92 documents that too.

src/or/rendservice.c
nmathewson: I'm content to leave rend* alone as legacy code.

src/or/nodelist.c
nmathewson: f5a472c20da9a0 does that.

src/or/circuitbuild.c
nmathewson: I've opened trac:25885 to change this if we want to do so someday.   I've added a comment to note this defect as df7eaf84439363.

src/or/control.c
nmathewson: e480c11097de79 does it here.

src/or/nodelist.c
nmathewson: looks good; done in 632899c85e919c.

src/or/routerlist.c
nmathewson: Opened trac:25886 for this.

teor2345:
> 5398bf02fea4a5 has that comment.

Two typos: consensus-indexed cnodes are also briges

teor2345:
> I've added a comment to note this defect as df7eaf84439363.

Two typos: thar are note the type we would prefer to use

nmathewson:
This was squashed and merged with 6773102c92e7bd9d949251da75bcc0762e754ab2

PR #50, opened 2018-04-16
Ticket25762 034 01

dgoulet-tor:
Some stuff are still missing but this is a first pass at this. See ticket on trac for details.

### DIFF
--- a/src/or/main.c
+++ b/src/or/main.c
@@ -1419,7 +1419,9 @@ reset_all_main_loop_timers(void)
 {
   int i;
   for (i = 0; periodic_events[i].name; ++i) {
-    periodic_event_reschedule(&periodic_events[i]);
+    if (periodic_event_is_enabled(&periodic_events[i])) {
+      periodic_event_reschedule(&periodic_events[i]);
+    }
   }
 }
 
@@ -1454,7 +1456,9 @@ initialize_periodic_events_cb(evutil_socket_t fd, short events, void *data)
   tor_event_free(initialize_periodic_events_event);
   int i;
   for (i = 0; periodic_events[i].name; ++i) {
-    periodic_event_launch(&periodic_events[i]);
+    if (periodic_event_is_enabled(&periodic_events[i])) {
+      periodic_event_launch(&periodic_events[i]);
+    }
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/or/main.c
```diff
@@ -1458,6 +1458,55 @@ initialize_periodic_events_cb(evutil_socket_t fd, short events, void *data)
   }
 }
 
+/* Helper function: Set up periodic_events[] that only matches the given
```
nmathewson: Documentation should be a doxygen comment, and should explain what "matches" means.

src/or/main.c
```diff
@@ -1458,6 +1458,55 @@ initialize_periodic_events_cb(evutil_socket_t fd, short events, void *data)
...
+ * roles. */
+static void
+setup_periodic_events_by_roles(uint32_t roles)
+{
+  for (int i = 0; periodic_events[i].name; ++i) {
+    periodic_event_item_t *item = &periodic_events[i];
+    if (item->roles & roles) {
+      periodic_event_setup(item);
```
nmathewson: How does this work when we call this function later on?  If we're going to do this, we need to make periodic_event_setup idempotent.

Actually, we can safely just call periodic_event_setup() on all of these events, and use periodic_event_launch to start the ones we really want. We can add a new periodic_event_disable() to turn off the ones we don't want.

src/or/main.c
```diff
@@ -1458,6 +1458,55 @@ initialize_periodic_events_cb(evutil_socket_t fd, short events, void *data)
...
+
+/* Helper function: Set up all events in periodic_events[] and filter them by
+ * roles for which we are configured for. */
+static void
+setup_periodic_events(void)
+{
+  const or_options_t *options = get_options();
+
+  /* At this stage, we are booting up so we need to know for which roles this
+   * tor instance has. */
```
nmathewson: Let's make just one bitfield here, and call setup_periodic_events_by_roles once?

src/or/main.c
```diff
@@ -1551,6 +1551,21 @@ teardown_periodic_events(void)
...
+ * periodic events. */
+void
+periodic_events_on_new_options(void)
+{
+  /* Only if we've already initialized once the events, teardown them all and
+   * reinitialize. It is just simpler that way instead of going through all
+   * currently enabled events and trying to destroy only the one that could be
+   * affected. */
+  if (periodic_events_initialized) {
+    teardown_periodic_events();
```
nmathewson: This is not okay -- if we tear them down and then set them up again, the  actual timers will all be reset.

dgoulet-tor:
This was merged with another PR.

PR #53, opened 2018-04-17
Lazy bucket refill

nmathewson:


src/or/connection.c
```diff
@@ -3186,6 +3186,34 @@ connection_bucket_adjust(const or_options_t *options)
...
+  /* Note that we only check for equality here: the underlying
+   * token bucket functions can handle moving backwards in time if they
+   * need to. */
+  if (now_ts != last_refilled_global_buckets_ts) {
+    token_bucket_rw_refill(&global_bucket, now_ts);
+    token_bucket_rw_refill(&global_relayed_bucket, now_ts);
+    last_refilled_global_buckets_ts = now_ts;
+  }
+
+  if (connection_speaks_cells(conn) && conn->state == OR_CONN_STATE_OPEN) {
```
dgoulet-tor: So hmmm, we don't count bw when negotiating TLS (`state != OPEN`)?

src/or/connection.c
```diff
@@ -3117,7 +3121,8 @@ connection_consider_empty_read_buckets(connection_t *conn)
   if (!connection_is_rate_limited(conn))
     return; /* Always okay. */
 
-  bool is_global = true;
+  int is_global = 1;
+
```
dgoulet-tor: What is going here? Actually, do we need this `is_global` at all? It is still ignored in the `*_bw_exhausted()` functions.

src/or/connection.c
```diff
@@ -3233,55 +3240,76 @@ connection_bucket_refill_single(connection_t *conn, uint32_t now_ts)
...
+ * This event is scheduled after enough time has elapsed to be sure
+ * that the buckets will refill when the connections have something to do.
+ */
+static void
+reenable_blocked_connections_cb(mainloop_event_t *ev, void *arg)
+{
+  (void)ev;
+  (void)arg;
+  SMARTLIST_FOREACH_BEGIN(get_connection_array(), connection_t *, conn) {
+    if (conn->read_blocked_on_bw == 1) {
```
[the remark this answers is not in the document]
nmathewson: So, not every connection was necessarily enabled before -- some might have been off because they had nothing to write, or for some other reason. Re-enabling everything would waste more CPU than it would save.

(Also note that this loop uses *less* CPU than the old loop it replaces.)

src/or/connection.c
nmathewson: (That's right, apparently.)

nmathewson: (At least, I don't think we did before, did we?)

src/or/connection.c
nmathewson: I'd like to leave it in, since we will likely want it in the future if we ever have special handling for the per-connection buckets.

src/or/connection.c
```diff
@@ -138,6 +138,8 @@ static const char *proxy_type_to_string(int proxy_type);
 static int get_proxy_type(void);
 const tor_addr_t *conn_get_outbound_address(sa_family_t family,
                   const or_options_t *options, unsigned int conn_type);
+static void blocked_connection_reenable_init(const or_options_t *options);
+static void schedule_blocked_connection_reenable(void);
```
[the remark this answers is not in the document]
nmathewson: bd3f8260a342b9 fixes this

src/or/connection.c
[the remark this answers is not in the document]
dgoulet-tor: Oh damn, I missed the `start_reading()` part ... so yes agree.

src/or/connection.c
dgoulet-tor: Ok but then why the removal of the `bool` ?

src/or/connection.c
dgoulet-tor: Hmm, yeah I think we do not looking at `connection_buf_read_from_socket()`

PR #56, opened 2018-04-18
Ticket25762 034 04

dgoulet-tor:


src/or/main.c
```diff
@@ -1348,37 +1348,54 @@ CALLBACK(write_stats_file);
...
+  CALLBACK(launch_descriptor_fetches, PERIODIC_EVENT_ROLE_ALL),
+  CALLBACK(reset_padding_counts, PERIODIC_EVENT_ROLE_ALL),
+  CALLBACK(retry_listeners, PERIODIC_EVENT_ROLE_ALL),
+  CALLBACK(rotate_x509_certificate, PERIODIC_EVENT_ROLE_ALL),
+  CALLBACK(write_stats_file, PERIODIC_EVENT_ROLE_ALL),
+
+  /* Routers (bridge and relay) only. */
+  CALLBACK(check_descriptor, PERIODIC_EVENT_ROLE_ROUTER),
+  CALLBACK(check_ed_keys, PERIODIC_EVENT_ROLE_ROUTER),
+  CALLBACK(check_for_reachability_bw, PERIODIC_EVENT_ROLE_ROUTER),
```
nmathewson: Do bridges evaluate their reachability and bw? If so, this is right.

src/or/main.c
```diff
@@ -1348,37 +1348,54 @@ CALLBACK(write_stats_file);
...
+  CALLBACK(check_onion_keys_expiry_time, PERIODIC_EVENT_ROLE_ROUTER),
+  CALLBACK(clean_consdiffmgr, PERIODIC_EVENT_ROLE_ROUTER),
+  CALLBACK(expire_old_ciruits_serverside, PERIODIC_EVENT_ROLE_ROUTER),
+  CALLBACK(retry_dns, PERIODIC_EVENT_ROLE_ROUTER),
```
nmathewson: Bridges shouldn't do this, should they?

src/or/main.c
```diff
@@ -1348,37 +1348,54 @@ CALLBACK(write_stats_file);
...
+  CALLBACK(retry_listeners, PERIODIC_EVENT_ROLE_ALL),
+  CALLBACK(rotate_x509_certificate, PERIODIC_EVENT_ROLE_ALL),
+  CALLBACK(write_stats_file, PERIODIC_EVENT_ROLE_ALL),
+
+  /* Routers (bridge and relay) only. */
+  CALLBACK(check_descriptor, PERIODIC_EVENT_ROLE_ROUTER),
+  CALLBACK(check_ed_keys, PERIODIC_EVENT_ROLE_ROUTER),
+  CALLBACK(check_for_reachability_bw, PERIODIC_EVENT_ROLE_ROUTER),
+  CALLBACK(check_onion_keys_expiry_time, PERIODIC_EVENT_ROLE_ROUTER),
+  CALLBACK(clean_consdiffmgr, PERIODIC_EVENT_ROLE_ROUTER),
```
nmathewson: I think this needs a separate "DirServer" role, unless it is going to be "Everyone".  We have a bug recently where we assumed that all DirServers were relays of some kind, but that doesn't seem to be true.. See trac:23693

src/or/main.c
dgoulet-tor: `retry_dns()` uses `server_mode()` so maybe we should use `public_server_mode()` there if you think it is no good?

src/or/main.c
dgoulet-tor: Again as above, `server_mode()` is used meaning bridge go in that path. And because all relays are `DirServer`, it seems OK to assign the Router role where a bridge also will serve consensus to their clients thus need to clean the cache?

src/or/main.c
dgoulet-tor: I believe so as `server_mode()` is used in that function thus bridges should do it.

dgoulet-tor: Seems fine here from IRC discussion.

src/or/main.c
dgoulet-tor: We'll open a new ticket here since `retry_dns()` only applies to Exit node thus the role should be `Relay` but also the function should use `public_server_mode()` or actually straight `options->ExitRelay`.

src/or/main.c
dgoulet-tor: Per IRC discussion, we need a DirServer role since `DirCache 0` will make this callback pointless for a relay.

nmathewson: I opened #25900 for this

src/or/main.c
nmathewson: I opened #25899 for this.

dgoulet-tor:
This has been merged.

PR #57, opened 2018-04-19
Set number of primaries to the number of configured guards.

asn-d6:


nmathewson:
A version of this was squashed and merged with 346c2eb4e60299f69327b5e63e1e25aaf678f964

PR #58, opened 2018-04-25
Ticket25610 034 01

dgoulet-tor:


src/or/include.am
```diff
@@ -254,11 +263,22 @@ ORHEADERS = \
...
 	src/or/routerparse.h				\
 	src/or/scheduler.h				\
+	src/or/shared_random_common.h			\
 	src/or/statefile.h				\
 	src/or/status.h					\
 	src/or/torcert.h				\
 	src/or/tor_api_internal.h
 
+# We add the headers of the modules even though they are disabled so we can
+# properly compiled the entry points replacedment.
```
ahf: I think `replacedment` is a typo here. Maybe 'stub functions' are better?

src/test/include.am
```diff
@@ -79,7 +79,14 @@ src_test_AM_CPPFLAGS = -DSHARE_DATADIR="\"$(datadir)\"" \
 # This seems to matter nowhere but on Windows, but I assure you that it
 # matters a lot there, and is quite hard to debug if you forget to do it.
 
+DIRAUTH_MOD = \
+  src/or/dirauth/dircollate.c				\
```
ahf: Indentation seems off here? Should be \t.

src/test/include.am
```diff
@@ -79,7 +79,14 @@ src_test_AM_CPPFLAGS = -DSHARE_DATADIR="\"$(datadir)\"" \
...
+  src/or/dirauth/dirvote.c				\
+  src/or/dirauth/shared_random.c				\
+  src/or/dirauth/shared_random_state.c
+
 src_test_test_SOURCES = \
+  $DIRAUTH_MOD \
```
ahf: Same here.

src/or/dirauth/dirvote.h
```diff
@@ -86,90 +84,111 @@
...
+
+static inline struct pending_vote_t *
+dirvote_add_vote(const char *vote_body, const char **msg_out, int *status_out)
+{
+  (void) vote_body;
+  (void) msg_out;
+  (void) status_out;
+  /* If the dirauth module is disabled, this should NEVER be called else we
+   * failed to safeguard the dirauth module. */
+  tor_assert(0);
```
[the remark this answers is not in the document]
dgoulet-tor: Well there is a backtrace with a `tor_assert()` but the main point for me was also to stop the whole thing in that case. It should really NEVER happen.

src/or/include.am
dgoulet-tor: Yes!

src/test/include.am
dgoulet-tor: Yeah that commit doesn't work and will disappear. Sorry about this, it got forgotten :(

PR #64, opened 2018-04-27
Ticket25610 034 01

dgoulet-tor:


src/or/dirauth/shared_random.h
```diff
@@ -101,11 +101,40 @@ typedef struct sr_commit_t {
 
 /* API */
 
-/* Public methods: */
+/* Public methods used _outside_ of the module.
+ *
+ * We need to nullify them if the module is disabled. */
+#ifdef HAVE_MODULE_DIRAUTH
```
nmathewson: We should make sure we have "#ifdef HAVE_MODULE_DIRAUTH" surrounding all of the include files in dirauth.h, except for any parts that should be accessible with dirauths are disabled.

src/or/dirauth/shared_random.h
```diff
@@ -101,11 +101,40 @@ typedef struct sr_commit_t {
...
+  return 0;
+}
+
+static inline void
+sr_save_and_cleanup(void)
+{
+}
+
+static inline void
+sr_act_post_consensus(const networkstatus_t *consensus)
```
nmathewson: I'm not sure that this function is one it makes sense to expose in the case when this module is disabled. Do we really need to do so?  It doesn't seem as logical as the "init" and "cleanup" functions.

src/or/shared_random_common.c
```diff
@@ -0,0 +1,258 @@
+/* Copyright (c) 2018, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
+ * \file shared_random_common.c
```
nmathewson: I don't think "common" is the right name for this kind of module.  This isn't exactly "the parts of the API that are shared by some other things" -- it's "the parts of the API that are used even if you aren't a dirauth."  Maybe "shared_random_user" or "shared_random_client"?  Hm.  Or maybe something else.  This is tricky.

src/or/dirvote_common.h
```diff
@@ -0,0 +1,69 @@
+/* Copyright (c) 2018, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
+ * \file dirvote_common.h
+ * \brief Header file for dirvote_common.c.
```
nmathewson: See comment previously on shared_random_common.c about why I don't like the "common" name here.

This module could probably be split up into the part about consensus schedules; the part about signatures; and the part about authority certs.  But it would make sense to do that _after_ this branch is merged, so let's open a ticket to remember?

src/or/dirvote_common.c
```diff
@@ -200,7 +200,7 @@ dirvote_get_voter_sig_by_alg(const networkstatus_voter_info_t *voter,
 /** Allocate and return a new authority_cert_t with the same contents as
  * <b>cert</b>. */
 authority_cert_t *
-authority_cert_dup(authority_cert_t *cert)
+dirvote_authority_cert_dup(authority_cert_t *cert)
```
nmathewson: I think the original name for this function was correct, since it acts on authority_cert_t.  I think this whole commit should be removed or reverted.

src/or/dirvote_common.c
```diff
@@ -182,8 +182,8 @@ dirvote_recalculate_timing(const or_options_t *options, time_t now)
 /** Return the signature made by <b>voter</b> using the algorithm
  * <b>alg</b>, or NULL if none is found. */
 document_signature_t *
-voter_get_sig_by_algorithm(const networkstatus_voter_info_t *voter,
-                           digest_algorithm_t alg)
+dirvote_get_voter_sig_by_alg(const networkstatus_voter_info_t *voter,
```
nmathewson: Maybe this function should be networkstatus_voter_get_sig_by_alg?  It works  on a networkstatus_voter_info_t.

src/or/dirauth/dirvote.h
```diff
@@ -113,6 +115,13 @@ dirvote_free_all(void)
 {
 }
 
+static inline void
+dirvote_parse_sr_commits(networkstatus_t *ns, smartlist_t *tokens)
```
nmathewson: Why should anything be calling this method if the module is disabled?

src/or/dirauth/dirvote.c
```diff
@@ -3828,3 +3829,73 @@ dirvote_format_all_microdesc_vote_lines(const routerinfo_t *ri, time_t now,
...
+  smartlist_free(chunks);
+  smartlist_free(commits);
+}
+
+/* Using the given directory tokens in tokens, parse the shared random commits
+ * and put them in the given vote document ns.
+ *
+ * This also sets the SR participation flag if present in the vote. */
+void
+dirvote_parse_sr_commits(networkstatus_t *ns, smartlist_t *tokens)
```
nmathewson: Should "tokens" be const?

src/or/dirauth/dirvote.c
```diff
@@ -3899,3 +3899,16 @@ dirvote_parse_sr_commits(networkstatus_t *ns, smartlist_t *tokens)
   }
 }
 
+/* For the given vote, free the shared random commits if any. */
+void
+dirvote_free_commits(networkstatus_t *ns)
```
nmathewson: I think this function should have a name like "clear_commits" or "remove_commits", not "free_commits", since it doesn't take the commits as an argument.

src/or/dirauth/dirvote.h
```diff
@@ -140,16 +148,32 @@ dirvote_dirreq_get_status_vote(const char *url, smartlist_t *items,
...
 
+static inline struct pending_vote_t *
+dirvote_add_vote(const char *vote_body, const char **msg_out, int *status_out)
+{
+  (void) vote_body;
+  (void) msg_out;
+  (void) status_out;
+  /* If the dirauth module is disabled, this should NEVER be called else we
+   * failed to safeguard the dirauth module. */
+  tor_assert(0);
```
nmathewson: This should probably be tor_assert_nonfatal_unreached() -- it's not worth crashing over, is it?

src/or/dirauth/dirvote.h
```diff
@@ -140,16 +148,32 @@ dirvote_dirreq_get_status_vote(const char *url, smartlist_t *items,
...
+static inline int
+dirvote_add_signatures(const char *detached_signatures_body, const char *source,
+                       const char **msg_out)
+{
+  (void) detached_signatures_body;
+  (void) source;
+  (void) msg_out;
+  /* If the dirauth module is disabled, this should NEVER be called else we
+   * failed to safeguard the dirauth module. */
+  tor_assert(0);
```
nmathewson: This should probably be tor_assert_nonfatal_unreached() -- it's not worth crashing over, is it?

src/or/dirauth/shared_random_state.c
```diff
@@ -9,11 +9,13 @@
  **/
 
 #define SHARED_RANDOM_STATE_PRIVATE
+#define TOR_SHARED_RANDOM_COMMON_PRIVATE
```
nmathewson: One C file should never define another C file's "private" macro.  Is there some way we can avoid doing this?  If something is private to shared_random_common, literally nothing else besides the tests should be looking at it.

src/or/dirauth/shared_random.h
dgoulet-tor: Indeed, it should be dirauth only. However, it is used in `directory.c` when a consensus is fetched and if we are acting as a dirauth, we call that post consensus function.

It should be made dirauth only when we refactor the directory.c file for all `authdir_mode_v3()` places that should go in `src/or/dirauth/` at that point.

src/or/shared_random_common.c
dgoulet-tor: Yes tricky, we had a bikeshed moment but I'm still unsure to this moment...

Bottom line is that they are functions that other non-dirauth part of Tor needs. I like the `_client.c` idea. For us developers, it makes sense but looking from another angle, all the functions in those two `_common.h` files are only for HS.

Actually, everything in `dirvote_common.h` is for `shared_random_common.c`  and then that file is for hidden service.

There could be an argument to actually isolate all of this in one hidden service file like `hs_shared_random.c` and `dirvote_common.c` into that file or in a new specific file?

src/or/dirvote_common.h
dgoulet-tor: Yes

src/or/dirauth/shared_random.h
nmathewson: err, whoops.  I meant "all the include files in dirauth/*.h"

src/or/dirauth/dirvote.c
dgoulet-tor: In theory yes, it would be ideal but `find_opt_by_keyword()` doesn't take a const :S ... So as a minimal change, this is it. We could add a commit to make that find function take a const though...

src/or/dirauth/dirvote.h
dgoulet-tor: Hmmm dunno, if your Tor is voting but without a dirauth module, I believe its worth stopping... I mean in theory it just can't happen with the amount of checks there is but if we happen to be in that path, I would prefer that tor stop instead of starting spamming dirauth with vote.

src/or/dirauth/shared_random_state.c
dgoulet-tor: Yes actually... I believe this is an artifact and it builds perfectly without it!

Fixup commit e614ae1a03

src/or/dirvote_common.c
dgoulet-tor: Yeah actually, that function is not longer needed outside `dirauth/dirvote.c` so it should go back there and thus untouched.

I can't revert the commit as it is with other things but I made an extra commit putting it back in `dirvote.c` with original name: bc393fc737

src/or/dirvote_common.c
dgoulet-tor: The `dirvote_` is simply to namespace it to the file which `networkstatus_` would fail that with the current filename.

However, I do see `networkstatus_get_voter_by_id()` so maybe we should simply move that function into `networkstatus.c` ?

src/or/dirauth/dirvote.c
dgoulet-tor: Done in fixup: 3300369875

src/or/dirauth/shared_random.h
dgoulet-tor: We can't ... header file in `dirauth/` have symbols used outside of the dirauth module which are NOP if no dirauth module.

If we ifdef the include statement, it will fail to compile by not finding the symbols.

src/or/shared_random_common.c
nmathewson: I think we can rename these post-merge, but we should figure out what we're renaming these to.

src/or/dirauth/shared_random.h
nmathewson: Okay.  I'll look for things we can #ifdef postmerge...

src/or/dirauth/dirvote.c
nmathewson: okay, I'll fix postmerge

src/or/dirauth/dirvote.h
nmathewson: The counterargument here is that it's better to spam relays logs than to take relays down off the network because we have some stupid bug we forgot about.

src/or/dirvote_common.c
nmathewson: Yes, let's.  We can do it postmerge.

src/or/shared_random_common.c
dgoulet-tor: Agree. The more we talk about it and the more I don't like `_common.c`.

src/or/dirauth/dirvote.h
dgoulet-tor: Fair enough.

See fixup commit 0765bf37fc

src/or/dirvote_common.c
dgoulet-tor: Ack.

src/or/dirauth/dirvote.h
dgoulet-tor: Should not! But it is used in `routerparse.c` within `networkstatus_parse_vote_from_string()` which parses ALL THE THINGS including votes which are dirauth only.

So this is one of the example that once we can extract the vote parsing in some ways into the dirauth module, this function won't be public anymore.

nmathewson:
We merged a squashed version of this with d018bf199c9e566202eef0bb7fb68c2567c4a25e

PR #65, opened 2018-04-27
Add Travis CI builds with modules disabled (Trac: #25953)

ahf:
(#64 needs to land before this should land)

This patch adds a GCC and Clang build to Travis where we build with the dirauth module disabled.

See: https://bugs.torproject.org/25953


nmathewson:
The relevant part of this ticket, dgoulet says on trac:25953,  was cherry-picked as 60a2c92d105927a782dad16eb75b710757db61b2

PR #66, opened 2018-04-27
Eliminate gettimeofday cached

nmathewson:


src/or/circuitmux_ewma.c
```diff
@@ -597,24 +599,48 @@ cell_ewma_to_circuit(cell_ewma_t *ewma)
...
+/**
+ * Initialize the system that tells which ewma tick we are in.
+ */
+STATIC void
+cell_ewma_initialize_ticks(void)
+{
+  if (ewma_ticks_initialized)
+    return;
+  monotime_coarse_get(&start_of_current_tick);
+  crypto_rand((char*)&current_tick_num, sizeof(current_tick_num));
```
dgoulet-tor: Oh wow, I guess this is an upgraded security feature that we didn't have? That is starting the EWMA tick number at a random value! :) +1

nmathewson: I don't *think* it's a security feature, since we don't actually expose tick numbers.

My motivation here was to make sure that we would observe tick rollover sometimes, even on relays that aren't up for long enough to overflow the tick counter.  Otherwise, overflow would be very difficult to test.

PR #72, opened 2018-05-01
Improve test coverage for addressmap_get_virtual_address.

nmathewson:


src/test/test_addr.c
```diff
@@ -941,6 +942,156 @@ test_virtaddrmap(void *data)
...
+
+  // Try a hostname.
+  a = addressmap_register_virtual_address(RESOLVED_TYPE_HOSTNAME,
+                                          tor_strdup("foobar.baz"));
+  tt_assert(a);
+  tt_assert(!strcmpend(a, ".virtual"));
+
+  // mock crypto_rand to repeat the same result twice; make sure we get
+  // different outcomes.  (Because even though it's only 2^40 for this
+  // to happen, 2^40 isn't all that low.)
```
tlyu: Maybe say something like "for this to happen with uniformly distributed 80-bit outputs" to make it more clear it's referring to the birthday problem? Also suggest "2^40 is still more iterations than is reasonable for a test suite".

nmathewson:
A squashed version of this branch is now merged as `ticket25993_squashed`

PR #77, opened 2018-05-02
Add bw lines to test with order variations

teor2345:


src/test/test_dir.c
```diff
@@ -1497,8 +1497,15 @@ test_dir_measured_bw_kb(void *arg)
...
                 "bw=1024 junk=007\n",
+    /* check whether node_id must be the first */
     "misc=junk node_id=$557365204145532d32353620696e73746561642e  "
                 "bw=1024 junk=007\n",
+    /* check whether a key_value can be in the middle of node_id and bw */
+    "node_id=$557365204145532d32353620696e73746561642e foo=bar bw=1024\n",
+    "misc=junk node_id=$557365204145532d32353620696e73746561642e foo=bar "
+    "bw=1024\n",
+    /* check whether a key_value can be after bw */
+    "node_id=$557365204145532d32353620696e73746561642e bw=1024 foo=bar\n",
```
teor2345: node_id can't be at the end of the line, but it can be after bw.

For example, this line should succeed:
"bw=1024 node_id=$557365204145532d32353620696e73746561642e foo=1\n"

juga0: That's what i expected, but it doesn't!!, so i had to put it in the "lines_fail" (https://github.com/torproject/tor/pull/77/commits/3b88ea4e56e05e1ca948823d33b32cb34d5ef9bb#diff-08742ca5bf9a35b909d5e8080b6f725bR1543), and that's why i say in @tor-dev that i checked and the only order required is bw before node_id

teor2345: The test you linked to is:
"bw=1024 node_id=$557365204145532d32353620696e73746561642e\n"
which is not the same.

juga0: Oh, i realize now, you're right!, and i realize you created also #26004.
Should i open a new ticket in trac for these extra lines, or would this be part of #26004 or #25947?

teor2345: #25947, please, it's the unit test task.

juga0:
Closing because this is now solved with the patch to allow node_id in the end #26004 and merged in master https://gitweb.torproject.org/tor.git/commit/?id=dbc80ad19b8a27aab557fe7930e18eef8f415133.

juga0:
Oh, i can't close because it was you that opened it

PR #78, opened 2018-05-02
Add a test for dirserv_read_measured_bandwidths,

juga0:
@teor2345 this is the branch with the test where i was using document files, but they fail as soon as the hardcoded timestamp is not valid. That's why i didn't create a PR before, was going to wait for the additional headers test, but i should probably work on this without the additional headers.


juga0:
It'd be easy to follow the same approach and create a branch that read an empty file as a test for #26007, but version 0.2.9 does not include ``SRCDIR``, that was added recently.
Would that be a valid test or better continue with #25947 and mock reading a file?.

teor2345:
> It'd be easy to follow the same approach and create a branch that read an empty file as a test for #26007, but version 0.2.9 does not include SRCDIR, that was added recently.
Would that be a valid test or better continue with #25947 and mock reading a file?.

Filesystem tests are really hard to do correctly, particularly on Windows.
It is better to mock reading a file.

Tests do not need to be backported if it is too hard.

juga0:
Then we can close this PR, as it adds the files to the filesystem.

PR #82, opened 2018-05-03
Remove second-counting responsibility from second_elapsed_callback

nmathewson:
See Tor trac ticket 26009.

### DIFF
--- a/src/or/main.c
+++ b/src/or/main.c
@@ -2495,8 +2495,40 @@ hs_service_callback(time_t now, const or_options_t *options)
 static periodic_timer_t *second_timer = NULL;
 /** Number of libevent errors in the last second: we die if we get too many. */
 static int n_libevent_errors = 0;
-/** Last time that second_elapsed_callback was called. */
+
+/** Last time that update_current_time was called. */
 static time_t current_second = 0;
+
+/**
+ * Set the current time to "now", which should be the value returned by
+ * time().  Check for clock jumps and track the total number of seconds we
+ * have been running.
+ */
+void
+update_current_time(time_t now)
+{
+  if (PREDICT_LIKELY(now == current_second)) {
+    /* We call this function a lot.  Most frequently, the current second
+     * will not have changed, so we just return. */
+    return;
+  }
+
+  const time_t seconds_elapsed = current_second ? (now - current_second) : 0;
+
+/** If more than this many seconds have elapsed, probably the clock
+ * jumped: doesn't count. */
+#define NUM_JUMPED_SECONDS_BEFORE_WARN 100
+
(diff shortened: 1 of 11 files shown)


src/or/main.c
```diff
@@ -2515,13 +2517,42 @@ update_current_time(time_t now)
...
-      seconds_elapsed >= NUM_JUMPED_SECONDS_BEFORE_WARN) {
-    circuit_note_clock_jumped(seconds_elapsed);
+  /** How much idleness do we tolerate? */
+#define NUM_IDLE_SECONDS_BEFORE_WARN 3600
+
+  if (seconds_elapsed < -NUM_JUMPED_SECONDS_BEFORE_WARN) {
+    // moving back in time is always a bad sign.
+    circuit_note_clock_jumped(seconds_elapsed, false);
+  } else if (seconds_elapsed >= NUM_JUMPED_SECONDS_BEFORE_WARN) {
+    /* Compare the monotonic clock to the */
```
dgoulet-tor: This comment is incomplete.

src/or/main.c
```diff
@@ -2515,13 +2517,42 @@ update_current_time(time_t now)
...
+    /* If the monotonic clock deviates from time(NULL), we have a couple of
+     * possibilities.  On some systems, this means we have been suspended or
+     * sleeping.  Everywhere, it can mean that the wall-clock time has
+     * been changed -- for example, with settimeofday().
+     *
+     * On the other hand, if the monotonic time matches with the wall-clock
+     * time, we've probably just been idle for a while, with no events firing.
+     * we tolerate much more of that.
+     */
+    const bool clock_jumped = abs(discrepency) > 2;
```
dgoulet-tor: If I understand correctly, we tolerate 2 second discrepancy with this?

(Also Google is telling me that it is `discrepancy`)

src/or/main.c
nmathewson: bbf547b79e7995f9e8de1616a97cc777050dcbba fixes this

src/or/main.c
nmathewson: That's right.  Actually, it's a little more complicated than that because of the "divide by 1000" logic: we're actually tolerating a 2999 ms discrepancy.  Do you think it should be less?

(spelling fix in b90c1189ddc85b675e1b149ee21356877141c8f9)

dgoulet-tor: Nea, I think it is fine. My comment was also mostly asking why 2 sec and 5 sec or 1 sec... so if you really have something specific in mind, I would recommend just an extra comment explaining the choice of "2 sec". If it is really just "well we are winging it", fair enough :).

nmathewson: Totally winging it. :)

nmathewson:
Squashed and merged with f684b48b5be5515395846d089a386167e76f895a

PR #85, opened 2018-05-04
Bug25647

isislovecruft:
https://bugs.torproject.org/25647


src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
   return 0;
 }
 
+/*****************************************************************************
```
nmathewson: We don't generally use this header format in the rest of our codebase.  Then again, we don't necessarily have a consistent way to indicate sections of a module.

src/or/onion.h
```diff
@@ -66,6 +74,25 @@ typedef struct create_cell_t {
...
+   * parsing by handshake code until this struct's <b>finished</b> bit is set.
+   */
+  create2v_cell_body_t *body;
+  /**
+   * Whether or not we're done parsing incoming fragments of this cell.  If the
+   * bit is set to 1, then we've collected all the data that the first cell
+   * fragment specified as the length of all fragments combined.  Otherwise, if
+   * 0, we're still waiting on incoming data.
+   */
+  unsigned int finished: 1;
```
nmathewson: Could use bool here

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+ *
+ * All variable length fields in the CREATE2V cell's body will be set
+ * to NULL, and all their corresponding lengths set to 0x00.
+ */
+create2v_cell_t*
+create2v_cell_new(void)
+{
+  create2v_cell_t *cell;
+
+  cell = tor_malloc(sizeof(create2v_cell_t));
```
nmathewson: Should either use tor_malloc_zero() or initialize "finished" explicitly.

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+ * Conduct various length checks on some handshake and padding data lengths for
+ * a given <b>handshake_type</b>.
+ *
+ * If <b>is_initiator</b> is true, then lengths are checked for the Alice side
+ * of the handshake, otherwise, if false, the checks are run for Bob's
+ * response.
+ */
+// XXXisis should this instead return bool and report errors rather than
+// asserting?
+static void
```
nmathewson: Yes, that would be a good idea.

nmathewson: And if this is going to be an assert at some point in the code, we should try to make it a nonfatal assertion if possible: those are less likely to turn into remote crash bugs.

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+  if (padding_len)
+    memcpy(create2v_cell_body_getarray_ignored(cell_out->body),
+           padding_data, padding_len);
+}
+
+/**
+ * Deallocate a CREATE2V cell.
+ */
+void
+create2v_cell_free(create2v_cell_t *cell)
```
nmathewson: Please use the same conventions as elsewhere in the code: name this function ending with an underscore, and #define create2v_cell_free() as a macro using FREE_AND_NULL.

Also, make sure this function accepts NULL as an input.

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+
+/**
+ * Given the Create2VMaximumData consensus parameter, calculate the
+ * total maximum allowable size of an incoming CREATE(D)2V buffer.
+ */
+static int
+create2v_cell_calculate_maximum_total(void)
+{
+  return networkstatus_get_create2v_maximum_data(NULL)
+    // XXXisis what about these extra bits of stuff? are we only counting
```
nmathewson: I think it's fine to decide either way, but we should patch the proposal to clarify.  I say we go with whatever makes things clearer.

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+}
+
+/**
+ * Check that the encoded length of an incoming CREATE2V or CREATED2V cell is
+ * less than the maximum allowed.
+ *
+ * Returns true if it's okay, and false otherwise.
+ */
+static bool
+create2v_cell_check_encoded_length(const size_t len)
```
nmathewson: Let's try to use predicate names for boolean functions.  Instead of "check_encoded_length" maybe "encoded_length_is_ok"?

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+{
+  const uint16_t max_len = create2v_cell_calculate_maximum_total();
+
+  /* If the incoming buffer is reportedly way larger than that which would fill
+   * the allowable space for hdata and the ignored padding, then fail parsing
+   * early.
+   */
+  if (len > max_len) {
+    log_info(LD_OR, "Received a suspicious CREATE(D)2V cell whose reported "
```
nmathewson: This should use LOG_PROTOCOL_WARN as its severity, not LOG_INFO.  We should also _maybe_ log where the cell came from?

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+/**
+ * Check that the encoded length of an incoming CREATE2V or CREATED2V cell is
+ * less than the maximum allowed.
+ *
+ * Returns true if it's okay, and false otherwise.
+ */
+static bool
+create2v_cell_check_encoded_length(const size_t len)
+{
+  const uint16_t max_len = create2v_cell_calculate_maximum_total();
```
nmathewson: We should use an int or a uint or a size_t here, since the function can techincally return a value larger than uint16_max

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+}
+
+/**
+ * Parse the CREATE2V payload at <b>p</b>, which could be up to <b>p_len</b>
+ * bytes long, and use it to fill the fields of <b>cell_out->body</b>.
+ *
+ * Note that part of the body of an EXTEND2V cell is a CREATE2V payload, so
+ * this function is also used for parsing those.
+ *
+ * Returns true if parsing was successful, and false otherwise.
```
nmathewson: Hmmmmm.  This is very different than the 0 on success, negative on failure convention we use most everywhere else.  Are we sure we want to start using bools like this?  If we do, it will be hard to read code that checks a function's output for success.....

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+                       const uint8_t *p,
+                       const size_t p_len)
+{
+  create2v_cell_body_t *body;
+
+  /* Check the length before even attempting to parse it. */
+  if (!create2v_cell_check_encoded_length(p_len)) {
+    return false;
+  }
+  create2v_cell_body_parse(&body, p, p_len);
```
nmathewson: This has a return value, I think.  And you really do need to check it, or else body might not be valid.

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+ * If <b>unknown_ok</b> is true, then handshake types we don't know about yet
+ * are okay.
+ *
+ * In the future, checks such as verifying that the handshake data (for a
+ * particular handshake) should go here.
+ *
+ * Returns true if the <b>cell</b> is okay, and false otherwise.
+ */
+bool
+create2v_cell_check(const create2v_cell_t *cell, bool unknown_ok)
```
nmathewson: Here a bool return is okay if we name this function something like "...is_ok"

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+{
+  const char *not_okay;
+
+  not_okay = create2v_cell_body_check(cell->body);
```
nmathewson: I think this trunnel function can't actually fail on a parsed body ... only on a body we're thinking of parsing.

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+    return false;
+  if (body_len > payload_len)
+    return false;
+
+  memset(cell_out->payload, 0, payload_len);
+  cell_out->headers.command = CELL_CREATE2V;
+  cell_out->payload_len = payload_len;
+
+  if (create2v_cell_body_encode(cell_out->payload,
+                                (size_t)payload_len,
```
nmathewson: No need for a cast here

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+
+/**
+ * Fill <b>cell_out</b> with up to <b>payload_len</b> bytes of a correctly
+ * formatted version of the CREATE2V cell in <b>cell_in</b>.
+ *
+ * If <b>relayed</b> is true, then this is a cell not originating from us.
+ *
+ * Returns true if the formatting was successful, and false otherwise.
+ */
+static bool
```
nmathewson: See notes above on true = success

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+}
+
+/**
+ * Fill <b>cell_out</b> with up to <b>payload_len</b> bytes of a correctly
+ * formatted version of a CREATE2V cell originating from here in
+ * <b>cell_in</b>.
+ *
+ * Returns true if the formatting was successful, and false otherwise.
+ */
+bool
```
nmathewson: See notes above on true = success.

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+ * If <b>relayed</b> is true, then this is a cell not originating from us.
+ *
+ * Returns true if the formatting was successful, and false otherwise.
+ */
+static bool
+create2v_cell_format_impl(var_cell_t *cell_out,
+                          const size_t payload_len,
+                          const create2v_cell_t *cell_in,
+                          const bool relayed)
+{
```
nmathewson: Would it be cleaner for this function to allocate and return the var_cell_t ?  That way you don't need to worry about heap overruns if payload_len is not the allocated payload length of cell_out.

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+  if (padding_len)
+    memcpy(created2v_cell_body_getarray_ignored(cell_out->body),
+           padding_data, padding_len);
+}
+
+/**
+ * Deallocate a CREATED2V cell.
+ */
+void
+created2v_cell_free(created2v_cell_t *cell)
```
nmathewson: See notes about create2v_cell_free above

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+}
+
+/**
+ * Run various checks to ensure that a created2v_cell_t, <b>cell</b>, is
+ * well-formed.
+ *
+ * In the future, checks such as verifying that the handshake data (for a
+ * particular handshake) should go here.
+ */
+bool
```
nmathewson: See notes on bool=success

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+created2v_cell_check(const created2v_cell_t *cell)
+{
+  const created2v_cell_body_t *body = cell->body;
+  const char *not_okay = created2v_cell_body_check(body);
+
+  if (not_okay != NULL) {
+    log_warn(LD_BUG, "Unable to parse a CREATED2V cell: %s", not_okay);
```
nmathewson: This should be a LOG_PROTOCOL_WARN.

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+/**
+ * Parse the CREATED2V payload at <b>p</b>, which could be up to <b>p_len</b>
+ * bytes long, and use it to fill the fields of <b>cell_out->body</b>.
+ *
+ * Note that part of the body of an EXTENDED2V cell is a CREATED2V payload, so
+ * this function is also used for parsing those.
+ *
+ * Returns true if parsing was successful, and false otherwise.
+ */
+static bool
```
nmathewson: See earlier note on true=success

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+    return true;
+  return false;
+}
+
+/**
+ * Parse a CREATED2V cell from <b>cell_in</b> into <b>cell_out</b>.
+ *
+ * Returns true if parsing was successful, and false otherwise.
+ */
+bool
```
nmathewson: ibid

src/or/onion.c
```diff
@@ -1339,3 +1342,466 @@ extended_cell_format(uint8_t *command_out, uint16_t *len_out,
...
+ * Fill <b>cell_out</b> with a correctly formatted version of the CREATED2V
+ * cell in <b>cell_in</b>, up to <b>payload_len</b> bytes long.
+ *
+ * Returns true upon success, otherwise false if something went wrong.
+ */
+bool
+created2v_cell_format(var_cell_t *cell_out,
+                      const size_t payload_len,
+                      const created2v_cell_t *cell_in)
+{
```
nmathewson: See notes on create2v_cell_format

src/or/onion.h
```diff
@@ -121,5 +160,50 @@ int extend_cell_format(uint8_t *command_out, uint16_t *len_out,
 int extended_cell_format(uint8_t *command_out, uint16_t *len_out,
                          uint8_t *payload_out, const extended_cell_t *cell_in);
 
+// TODO: Remove these defines when we implement the first handshake that uses
+// CREATE(D)2V cells. --isis
+#ifndef ONION_HANDSHAKE_TYPE_NTOR2_NULL
+#define ONION_HANDSHAKE_TYPE_NTOR2_NULL 0x0003
```
nmathewson: Let's not use 3 even as a temporary value; instead let's reserve some range for temporary use

src/or/onion.h
```diff
@@ -121,5 +160,50 @@ int extend_cell_format(uint8_t *command_out, uint16_t *len_out,
 int extended_cell_format(uint8_t *command_out, uint16_t *len_out,
                          uint8_t *payload_out, const extended_cell_t *cell_in);
 
+// TODO: Remove these defines when we implement the first handshake that uses
+// CREATE(D)2V cells. --isis
+#ifndef ONION_HANDSHAKE_TYPE_NTOR2_NULL
```
nmathewson: These ifndef wrappers shouldn't be necessary -- instead, only define this constant in one place.

src/or/or.h
```diff
@@ -959,6 +959,8 @@ typedef enum {
 #define CELL_CREATE2 10
 #define CELL_CREATED2 11
 #define CELL_PADDING_NEGOTIATE 12
+#define CELL_CREATE2V 13
```
nmathewson: These values should be in the 128..255 range, since they are new variable-length cells.

src/or/networkstatus.c
```diff
@@ -2444,6 +2444,33 @@ networkstatus_get_weight_scale_param(networkstatus_t *ns)
...
+  if (!ns)
+    ns = networkstatus_get_latest_consensus();
+
+  if (!ns) {
+    log_info(LD_DIR, "We appear to have no consensus and thus don't know "
+             "Create2VMaximumData! Defaulting to the recommended value of %d.",
+             NETWORKSTATUS_DEFAULT_CREATE2V_MAXIMUM_DATA);
+  }
+
+  return (uint16_t)networkstatus_get_param(ns, "Create2VMaximumData",
```
nmathewson: I think networkstatus_get_param already knows to default to the latest consensus if its argument is NULL, and to default to the default if there is no networkstatus.  You can probably remove the earlier statements in this function.

PR #87, opened 2018-05-05
Ticket25947 029 01 Add bw lines to test with order variations

juga0:
Same as #77 based on maint-0.2.9.


juga0:
Closing because this is now solved with the patch to allow node_id in the end #26004 and merged in master https://gitweb.torproject.org/tor.git/commit/?id=dbc80ad19b8a27aab557fe7930e18eef8f415133.

PR #90, opened 2018-05-07
Bug17873 take3 squashed

rl1987:


src/or/connection.c
```diff
@@ -1146,7 +1150,8 @@ static connection_t *
 connection_listener_new(const struct sockaddr *listensockaddr,
```
nmathewson: If we're adding new arguments here, we need to document them.

src/or/connection.c
```diff
@@ -1162,6 +1167,8 @@ connection_listener_new(const struct sockaddr *listensockaddr,
   tor_addr_t addr;
   int exhaustion = 0;
 
+  if (addr_in_use) *addr_in_use = 0;
```
nmathewson: I'm assuming that this formatting problem (and other similar ones) gets fixed in a later commit

src/or/connection.c
```diff
@@ -2604,25 +2621,37 @@ connection_read_proxy_handshake(connection_t *conn)
   return ret;
 }
 
+struct replacement_s
```
nmathewson: We need to document this type.  What's it for?  What do the members mean?

src/or/connection.c
```diff
@@ -2604,25 +2621,37 @@ connection_read_proxy_handshake(connection_t *conn)
...
  * port_cfg_t entries in <b>ports</b>, open a new listener for every port in
  * <b>ports</b> that does not already have a listener in <b>old_conns</b>.
  *
  * Remove from <b>old_conns</b> every connection that has a corresponding
  * entry in <b>ports</b>.  Add to <b>new_conns</b> new every connection we
- * launch.
+ * launch. If we may need to perform socket rebind when creating new
+ * listener that replaces old one, create a <b>replacement_s</b> struct
+ * for affected pair  and add it to <b>replacements</b>. For more
+ * information, see ticket #17873.
```
nmathewson: We try not to use this style: the documentation for the code should stand by itself, and not require the bugtracker for help understanding why it is that way.

src/or/connection.c
```diff
@@ -2658,16 +2687,27 @@ retry_listener_ports(smartlist_t *old_conns,
           break;
         }
       } else {
-        int port_matches;
-        if (wanted->port == CFG_AUTO_PORT) {
```
nmathewson: What happens to this CFG_AUTO_PORT check?  it seems to get removed here, but not added somewhere else?

src/or/connection.c
```diff
@@ -2658,16 +2687,27 @@ retry_listener_ports(smartlist_t *old_conns,
...
+        int port_matches_exact = (wanted->port == conn->port);
+        int port_matches = wanted->port || port_matches_exact;
+
         if (port_matches && tor_addr_eq(&wanted->addr, &conn->addr)) {
           found_port = wanted;
           break;
         }
+
+        int may_need_rebind = port_matches_exact &
+                              (tor_addr_is_null(&wanted->addr) ^
```
nmathewson: Please use "bool_neq" here to avoid bugs.

src/or/connection.c
```diff
@@ -2658,16 +2687,27 @@ retry_listener_ports(smartlist_t *old_conns,
           break;
         }
       } else {
-        int port_matches;
-        if (wanted->port == CFG_AUTO_PORT) {
-          port_matches = 1;
-        } else {
-          port_matches = (wanted->port == conn->port);
-        }
+        int port_matches_exact = (wanted->port == conn->port);
```
nmathewson: Let's make these variables const, and document what we mean by them.

src/or/connection.c
```diff
@@ -2658,16 +2687,27 @@ retry_listener_ports(smartlist_t *old_conns,
...
+
+        int may_need_rebind = port_matches_exact &
+                              (tor_addr_is_null(&wanted->addr) ^
+                               tor_addr_is_null(&conn->addr));
+        if (replacements && may_need_rebind) {
+          struct replacement_s *replacement =
+            tor_malloc(sizeof(struct replacement_s));
+
+          replacement->old_conn = conn;
+          replacement->new_port = (port_cfg_t *)wanted;
```
nmathewson: Instead of casting here, it would be safer to make the struct member into a const pointer.

src/or/connection.c
nmathewson: Also let's give this a typedef name, and use that instead: we tend not to use "struct foo" in Tor.

src/or/connection.c
```diff
@@ -2727,9 +2768,40 @@ retry_all_listeners(smartlist_t *replaced_conns,
...
+    tor_assert(r->new_port);
+    tor_assert(r->old_conn);
+
+    connection_t *new_conn =
+     connection_listener_new_for_port(r->new_port, &skip, &addr_in_use);
+
+    if (skip)
+      continue;
+
+    // XXX: replaced_conns
```
nmathewson: What does this mean?

src/or/connection.c
```diff
@@ -112,6 +112,10 @@
 #include <sys/un.h>
 #endif
 
+#if defined(__linux__) || defined(_WIN32)
+#define ENABLE_LISTENER_REBIND
```
nmathewson: We should document here what makes these platforms different.

src/or/connection.c
rl1987: Yes, in ee3fbf1195052019c73c8c5e43b64bf5f17ab4a2 and 7cd5c0004eb4c55a598e0f918e6861d8293b6d70.

src/or/connection.c
rl1987: Fixed in 714ebd1e2b1878d17cc703f19f62c39c206b9473.

src/or/connection.c
rl1987: Oops. Fixed in 6975f6d8cb2724f7039938350497d8ec3f36ecc5.

src/or/connection.c
rl1987: Did so in 118cf220f6c81f54beeff3d31ecf0d2dfb3e83fc.

src/or/connection.c
rl1987: We don't really have a way to walk back to old listener configuration in case this rebinding business fails. Then again, we're not implementing any walking back in general case either (when rebinding is not needed) in `options_act_reversible`. I guess this is fine, as long as we're not getting stuck in some bad state where we cannot just set the old torrc back?

rl1987: Should I delete this line?

src/or/connection.c
rl1987: Fixed in ee3fbf1195052019c73c8c5e43b64bf5f17ab4a2 and 88de87e5d90ddaeaac21e8910428f6c3c9f68e98 .

teor2345:
Apparently this pull request should have been closed.

PR #91, opened 2018-05-07
Bug25852 take2

rl1987:
https://trac.torproject.org/projects/tor/ticket/25852

### DIFF
--- a/src/or/policies.c
+++ b/src/or/policies.c
@@ -2999,9 +2999,8 @@ getinfo_helper_policies(control_connection_t *conn,
     smartlist_free(private_policy_strings);
   } else if (!strcmp(question, "exit-policy/reject-private/relay")) {
     const or_options_t *options = get_options();
-    const routerinfo_t *me = NULL;
-
-    int err = router_get_my_routerinfo_with_err((routerinfo_t **)&me);
+    int err = 0;
+    const routerinfo_t *me = router_get_my_routerinfo_with_err(&err);
 
     if (!me) {
       *errmsg = routerinfo_errno_to_string(err);
@@ -3043,8 +3042,13 @@ getinfo_helper_policies(control_connection_t *conn,
     int include_ipv4 = 0;
     int include_ipv6 = 0;
 
-    const routerinfo_t *me = NULL;
-    int err = router_get_my_routerinfo_with_err((routerinfo_t **)&me);
+    int err = 0;
+    const routerinfo_t *me = router_get_my_routerinfo_with_err(&err);
+
+    if (!me) {
+      *errmsg = routerinfo_errno_to_string(err);
+      return routerinfo_err_is_transient(err) ? -1 : 0;
+    }
 
     if (!strcmp(question, "exit-policy/ipv4")) {
       include_ipv4 = 1;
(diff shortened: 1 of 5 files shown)


src/or/router.c
```diff
@@ -114,6 +114,56 @@ const char *format_node_description(char *buf,
                                     const tor_addr_t *addr,
                                     uint32_t addr32h);
 
+/** Return a readonly string with human readable description
+ * of <b>err</b>.
+ */
+const char *
+routerinfo_errno_to_string(int err)
```
dgoulet-tor: Hmm, `errno` is usually quite specific to the libc error values.

I would recommend just using `err`.

src/or/router.c
```diff
@@ -114,6 +114,56 @@ const char *format_node_description(char *buf,
...
+      return "Not running in server mode";
+    case TOR_ROUTERINFO_ERROR_DIGEST_FAILED:
+      return "Key digest failed";
+    case TOR_ROUTERINFO_ERROR_CANNOT_GENERATE:
+      return "Cannot generate descriptor";
+    case TOR_ROUTERINFO_ERROR_NOT_SO_FAST:
+      return "Too soon; not ready yet";
+  }
+
+  log_warn(LD_BUG, "unknown errno %d", err);
```
dgoulet-tor: We probably want more context on this error that is to inform that it is a `routerinfo` error here. I would also couple this with a `tor_assert_unreached()` so we can get the backtrace of how we got in here.

src/or/router.h
```diff
@@ -14,6 +14,13 @@
 
 #include "testsupport.h"
 
+#define TOR_ROUTERINFO_ERROR_NO_EXT_ADDR     (-1)
+#define TOR_ROUTERINFO_ERROR_CANNOT_PARSE    (-2)
+#define TOR_ROUTERINFO_ERROR_NOT_A_SERVER    (-3)
+#define TOR_ROUTERINFO_ERROR_DIGEST_FAILED   (-4)
+#define TOR_ROUTERINFO_ERROR_CANNOT_GENERATE (-5)
+#define TOR_ROUTERINFO_ERROR_NOT_SO_FAST     (-6)
```
dgoulet-tor: Seems this one is used when we learn when our IP address has changed and we have to rebuild the descriptor. I would call it something around those lines instead of a the vague `not so fast` ;).

src/or/policies.c
```diff
@@ -3038,11 +3040,12 @@ getinfo_helper_policies(control_connection_t *conn,
     SMARTLIST_FOREACH(configured_addresses, tor_addr_t *, a, tor_free(a));
     smartlist_free(configured_addresses);
   } else if (!strcmpstart(question, "exit-policy/")) {
-    const routerinfo_t *me = router_get_my_routerinfo();
-
     int include_ipv4 = 0;
     int include_ipv6 = 0;
 
+    const routerinfo_t *me = NULL;
+    int err = router_get_my_routerinfo_with_err((routerinfo_t **)&me);
```
dgoulet-tor: Hmmm I would go the other way around here so we don't have this cast violating the `const`.

Make `router_get_my_routerinfo_with_err()` return the `me` const pointer and pass a `int` pointer for the err. If your `me` is NULL, then we know to look at the `err`.

src/or/router.c
rl1987: Fixed in 8ad97b7dcf143fc488422f9866613fea3d900119.

src/or/router.c
rl1987: Did so in 036df13a03f6345913fbe5e7806fd00312d41c4d.

src/or/router.h
rl1987: Fixed.

src/or/policies.c
rl1987: Fixed in 36f7d0a940bbe8695d5343fccc1a0560a529883a.

PR #96, opened 2018-05-09
Ticket25947 029 squashed

juga0:
See comments in https://trac.torproject.org/projects/tor/ticket/25947#comment:11

PR #97, opened 2018-05-09
Test read bandwidth measurements with empty file

juga0:


juga0:
Will need rebase if #96 is merged

PR #99, opened 2018-05-09
Ticket26063: Don't run per-second timer when we don't have to.

nmathewson:


src/common/compat_libevent.c
```diff
@@ -253,10 +253,39 @@ periodic_timer_new(struct event_base *base,
   }
```
dgoulet-tor: Comment on the commit message. What does `objections` means in that context?

src/or/main.c
```diff
@@ -2491,6 +2491,36 @@ hs_service_callback(time_t now, const or_options_t *options)
...
+  if (! second_timer) {
+    second_timer = periodic_timer_new(tor_libevent_get_base(),
+                                      &one_second,
+                                      second_elapsed_callback,
+                                      NULL);
+    tor_assert(second_timer);
+  }
+
+  const bool run_per_second_events =
+    control_any_per_second_event_enabled() || ! net_is_completely_disabled();
```
dgoulet-tor: This seems that we also run if we have control event per second enabled. So we should fix the commit message to say that as well.

teor2345:
Closed on trac

PR #100, opened 2018-05-09
Give responsibility for waking up from DORMANT to a mainloop event

nmathewson:
Closes ticket 26064.

### DIFF
--- a/src/or/hibernate.c
+++ b/src/or/hibernate.c
@@ -52,6 +52,10 @@ static time_t hibernate_end_time = 0;
  * we aren't shutting down. */
 static time_t shutdown_time = 0;
 
+/** A timed event that we'll use when it's time to wake up from
+ * hibernation. */
+static mainloop_event_t *wakeup_event = NULL;
+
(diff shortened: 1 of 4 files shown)


src/or/hibernate.c
```diff
@@ -936,6 +942,63 @@ hibernate_go_dormant(time_t now)
...
+hibernate_schedule_wakeup_event(time_t now, time_t end_time)
+{
+  struct timeval delay = { 0, 0 };
+
+  if (now >= end_time) {
+    // In these cases we always wait at least a second, to avoid running
+    // the callback in a tight loop.
+    delay.tv_sec = 1;
+  } else {
+    delay.tv_sec = (end_time - now);
```
dgoulet-tor: I do think we are fairly sure that `end_time` can't be smaller than `now` but I would probably still check it since if it ever happens, that sleep time will be enormous.

dgoulet-tor: Yes you do. This is was clearly a big brain confusion on my part! :P

PR #104, opened 2018-05-11
Ticket25947 032

juga0:
Commits cherry-picked from branch https://github.com/juga0/tor/tree/ticket25947_029_squashed but based on maint-0.3.2, as i think we don't need to backport it to 0.2.9
Fix check-spaces and check-changes.
Not sure changes format is correct.
Not sure if instead of 4 commits as commented in https://trac.torproject.org/projects/tor/ticket/25947#comment:6, i should create 2 branches for the 2 different tickets and make only 1 commit on each.


juga0:
Following in #106

PR #105, opened 2018-05-11
Bug24658 dh stream

isislovecruft:
Creating a PR for @ffmancera's changes for the next part of #24658 so that I can review.


isislovecruft:
I apparently cannot request a review from myself if I'm the one who created the PR, even if none of the commits are mine.

src/common/crypto_util.c
```diff
@@ -103,5 +106,22 @@ memwipe(void *mem, uint8_t byte, size_t sz)
   memset(mem, byte, sz);
 }
 
+void
```
isislovecruft: We should be sure to move the corresponding documentation this function had as well.

src/tools/tor-gencert.c
```diff
@@ -78,29 +78,6 @@ show_help(void)
...
-  unsigned long err;
-  const char *msg, *lib, *func;
-  while ((err = ERR_get_error()) != 0) {
-    msg = (const char*)ERR_reason_error_string(err);
-    lib = (const char*)ERR_lib_error_string(err);
-    func = (const char*)ERR_func_error_string(err);
-    if (!msg) msg = "(null)";
-    if (!lib) lib = "(null)";
-    if (!func) func = "(null)";
-    if (doing) {
```
isislovecruft: This function was slightly different in that it didn't require `doing`… was this used anywhere important in `tor-gencert.c`?

isislovecruft: Nope, it wasn't, `doing` is set every time.

src/common/crypto_dh.h
```diff
@@ -0,0 +1,58 @@
...
+
+int crypto_expand_key_material_TAP(const uint8_t *key_in,
+                                   size_t key_in_len,
+                                   uint8_t *key_out, size_t key_out_len);
+int crypto_expand_key_material_rfc5869_sha256(
+                                    const uint8_t *key_in, size_t key_in_len,
+                                    const uint8_t *salt_in, size_t salt_in_len,
+                                    const uint8_t *info_in, size_t info_in_len,
+                                    uint8_t *key_out, size_t key_out_len);
+
```
isislovecruft: `crypto_expand_key_material_TAP` and `crypto_expand_key_material_TAP` should probably go into their own `crypto_hkdf.[ch]` module, it wouldn't really occur to me to look in the DH module and they don't need any of the types defined here.

src/common/crypto_s2k.c
```diff
@@ -15,6 +15,7 @@
 #include "compat.h"
 #include "crypto.h"
 #include "crypto_digest.h"
+#include "crypto_dh.h"
```
isislovecruft: If we make the `crypto_hkdf.[ch]` module I suggested, we won't need this include.

PR #116, opened 2018-05-20
Bug17873 take3 squashed2

rl1987:
https://trac.torproject.org/projects/tor/ticket/17873

### DIFF
--- a/src/or/connection.c
+++ b/src/or/connection.c
@@ -117,7 +117,11 @@ static connection_t *connection_listener_new(
                                const struct sockaddr *listensockaddr,
                                socklen_t listensocklen, int type,
                                const char *address,
-                               const port_cfg_t *portcfg);
+                               const port_cfg_t *portcfg,
+                               int *addr_in_use);
+static connection_t *connection_listener_new_for_port(
+                               const port_cfg_t *port,
+                               int *defer, int *addr_in_use);
 static void connection_init(time_t now, connection_t *conn, int type,
                             int socket_family);
 static int connection_handle_listener_read(connection_t *conn, int new_type);
@@ -1147,7 +1151,8 @@ static connection_t *
 connection_listener_new(const struct sockaddr *listensockaddr,
                         socklen_t socklen,
                         int type, const char *address,
-                        const port_cfg_t *port_cfg)
+                        const port_cfg_t *port_cfg,
+                        int *addr_in_use)
 {
   listener_connection_t *lis_conn;
   connection_t *conn = NULL;
@@ -1162,6 +1167,8 @@ connection_listener_new(const struct sockaddr *listensockaddr,
   static int global_next_session_group = SESSION_GROUP_FIRST_AUTO;
   tor_addr_t addr;
   int exhaustion = 0;
+
+  if (addr_in_use) *addr_in_use = 0;
 
   if (listensockaddr->sa_family == AF_INET ||
       listensockaddr->sa_family == AF_INET6) {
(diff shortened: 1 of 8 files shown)


rl1987:
https://trac.torproject.org/projects/tor/ticket/17873

teor2345:
Obsoleted by #90

src/or/connection.c
```diff
@@ -2659,16 +2688,31 @@ retry_listener_ports(smartlist_t *old_conns,
...
         }
+        const int may_need_rebind =
+          port_matches_exact & bool_neq(tor_addr_is_null(&wanted->addr),
+                                        tor_addr_is_null(&conn->addr));
+        if (replacements && may_need_rebind) {
+          struct replacement_s *replacement =
+            tor_malloc(sizeof(struct replacement_s));
+
+          replacement->old_conn = conn;
+          replacement->new_port = (port_cfg_t *)wanted;
```
nmathewson: It would be better to have the new_port field be a const pointer than to cast the const away here.

src/or/connection.c
```diff
@@ -2659,16 +2688,31 @@ retry_listener_ports(smartlist_t *old_conns,
...
+         */
+        const int port_matches = (wanted->port == CFG_AUTO_PORT ||
+                                  port_matches_exact);
+
         if (port_matches && tor_addr_eq(&wanted->addr, &conn->addr)) {
           found_port = wanted;
           break;
         }
+        const int may_need_rebind =
+          port_matches_exact & bool_neq(tor_addr_is_null(&wanted->addr),
```
nmathewson: Should this also check whether the addresses are different?  For example: 1.2.3.4 vs 1.2.3.5.  Also for example: [::] vs 0.0.0.0

src/test/test_rebind.py
```diff
@@ -0,0 +1,64 @@
...
+        if s in l:
+            return
+
+if not os.path.exists(sys.argv[1]):
+    sys.exit('ERROR: cannot find tor at %s' % sys.argv[1])
+
+tor_path = sys.argv[1]
+
+tor_process = subprocess.Popen([tor_path,
+                               '-ControlPort', '127.0.0.1:9053',
```
nmathewson: Somewhat risky to use fixed ports here: using fixed ports means that we can't actually run two of these in parallel, which can be a problem in some testing environments (eg running "make check" on two Tor versions at once).

src/or/connection.c
nmathewson: (We should also test for these cases)

teor2345:
Oops, #90 was obsoleted by this pull request.

src/or/connection.c
teor2345: I think you want '&&' here semantically, even though it happens to be equivalent to '&'.

src/or/connection.c
```diff
@@ -113,11 +113,33 @@
...
+ * specific address and the same port.
+ *
+ * To address this problem on these two platforms we implement a
+ * routine that:
+ * 1) Checks if first attempt to bind() a new socket  failed with
+ * EADDRINUSE.
+ * 2) If so, it will close the appropriate old listener connection and
+ * 3) Attempts bind()'ing the new listener socket again.
+ */
+#if defined(__linux__) || defined(_WIN32)
```
teor2345: If rebinding is a no-op when it's not required, let's do it on all platforms?

src/or/connection.c
```diff
@@ -1443,6 +1477,68 @@ connection_listener_new(const struct sockaddr *listensockaddr,
   return NULL;
 }
 
+/**
+ * Create a new listener connection for a given <b>port</b>. In case we
+ * cannot create listener now, but will be able later, set <b>defer</b>
+ * to true. If we cannot bind listening socket because address is already
```
teor2345: I'm not sure that this explanation of defer makes sense.
We unconditionally set defer whenever NoListen is set.
It has nothing to do with whether we can create a listener now, or whether we will be able to later.

src/or/connection.c
rl1987: Fixed in 03e4465cfa8bdc4b65951e1b74bb95226d41745a.

src/or/connection.c
rl1987: Did so in 7c34404e3bd832d5ab5d5a2de6255dc05a3b790c.

rl1987: Fixed - 7c34404e3bd832d5ab5d5a2de6255dc05a3b790c

PR #117, opened 2018-05-21
Improve openssl_version tests with better messages

nmathewson:
These tests would report errors, but wouldn't report the offending
strings.

### DIFF
--- a/src/test/test_crypto.c
+++ b/src/test/test_crypto.c
@@ -152,8 +152,13 @@ test_crypto_openssl_version(void *arg)
   const char *h_version = crypto_openssl_get_header_version_str();
   tt_assert(version);
   tt_assert(h_version);
-  tt_assert(!strcmpstart(version, h_version)); /* "-fips" suffix, etc */
-  tt_assert(!strstr(version, "OpenSSL"));
+  if (strcmpstart(version, h_version)) { /* "-fips" suffix, etc */
+    TT_DIE(("OpenSSL library version %s did not begin with header version %s.",
+            version, h_version));
+  }
+  if (strstr(version, "OpenSSL")) {
+    TT_DIE(("assertion failed: !strstr(\"%s\", \"OpenSSL\")", version));
+  }
   int a=-1,b=-1,c=-1;
   if (!strcmpstart(version, "LibreSSL") || !strcmpstart(version, "BoringSSL"))
     return;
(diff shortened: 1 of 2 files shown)


saper:
Would that be possible to add something to [autoconf tests](https://github.com/torproject/tor/blob/943291d7ae3d246c50fa55caa96c9bb04ee577eb/configure.ac#L833)? It was quite annoying to wait for a long time for the build to finish only to find out in the end. Maybe we could even loop in `[TOR_SEARCH_LIBRARY](https://github.com/torproject/tor/blob/943291d7ae3d246c50fa55caa96c9bb04ee577eb/configure.ac#L818)` to reject non-compliant combinations right away and try other ones.

src/test/test_crypto.c
```diff
@@ -152,8 +152,13 @@ test_crypto_openssl_version(void *arg)
   const char *h_version = crypto_openssl_get_header_version_str();
   tt_assert(version);
   tt_assert(h_version);
-  tt_assert(!strcmpstart(version, h_version)); /* "-fips" suffix, etc */
-  tt_assert(!strstr(version, "OpenSSL"));
+  if (strcmpstart(version, h_version)) { /* "-fips" suffix, etc */
+    TT_DIE(("OpenSSL library version %s did not begin with header version %s.",
+            version, h_version));
+  }
+  if (strstr(version, "OpenSSL")) {
```
isislovecruft: I think you want `!strstr` here.

src/test/test_crypto.c
```diff
@@ -152,8 +152,13 @@ test_crypto_openssl_version(void *arg)
...
+    TT_DIE(("assertion failed: !strstr(\"%s\", \"OpenSSL\")", version));
+  }
   int a=-1,b=-1,c=-1;
   if (!strcmpstart(version, "LibreSSL") || !strcmpstart(version, "BoringSSL"))
     return;
```
isislovecruft: This has nothing to do with your patch since it was already a bug in the unittest, but if `!strstr(version, "OpenSSL")` passes above, then this would only be true here if someone named their "OpenSSL" library `LibreSSLOpenSSL-6.6.6` or `BoringSSL-OpenSSL-42` or something, which… would be very weird. I think we're pretty unlikely to hit this line.

src/test/test_crypto.c
nmathewson: I don't think so -- the previous code did `assert(!strstr(...))`, but the new code does `if (strstr(...)) { DIE }`.  The `!` goes away because the sense of the test is inverted.

isislovecruft: Oh sorry, you're correct! Carry on.

PR #118, opened 2018-05-21
Appveyor min 034

nmathewson:
Here's a minimal attempt to get github working. It's based on work from other hackers who know appveyor better, and my own brute-force attempts.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,88 @@
+version: 1.0.{build}
+
+clone_depth: 50
+
+environment:
+  compiler: mingw
+
+  matrix:
+  - target: i686-w64-mingw32
+    compiler_path: mingw32
+    openssl_path: /c/OpenSSL-Win32
+  - target: x86_64-w64-mingw32
+    compiler_path: mingw64
+    openssl_path: /c/OpenSSL-Win64
+
+install:
+- ps: >-
+    Function Execute-Command ($commandPath)
+    {
+        & $commandPath $args 2>&1
+        if ( $LastExitCode -ne 0 ) {
+            $host.SetShouldExit( $LastExitCode )
+        }
+    }
+    Function Execute-Bash ()
+    {
+        Execute-Command 'c:\msys64\usr\bin\bash' '-e' '-c' $args
+    }
+    if ($env:compiler -eq "mingw") {
+            $oldpath = ${env:Path} -split ';'
+            $buildpath = @("C:\msys64\${env:compiler_path}\bin", "C:\msys64\usr\bin") + $oldpath
+            $env:Path = @($buildpath) -join ';'
+            $env:libeventversionfile = @("${env:APPVEYOR_BUILD_FOLDER}", ".appveyor-libevent-version") -join '\'
+            $env:libeventversion = Get-Content "${env:libeventversionfile}" | Out-String
+            Set-Location ".."
+            Execute-Bash "git clone https://github.com/libevent/libevent"
+            Set-Location "libevent"
+            $env:buildlibevent = (Get-Item -Path ".\").FullName
+            Execute-Bash "git checkout -q ${env:libeventversion}"
+            Execute-Bash "./autogen.sh"
+            Execute-Bash "./configure --prefix=/${env:compiler_path} --host=${env:target} --disable-libevent-regress --disable-samples --disable-openssl"
+            Execute-Bash "make -j2 install"
+    }
+
+build_script:
+- ps: >-
+    if ($env:compiler -eq "mingw") {
+            $oldpath = ${env:Path} -split ';'
+            $buildpath = @("C:\msys64\${env:compiler_path}\bin", "C:\msys64\usr\bin") + $oldpath
+            $env:Path = @($buildpath) -join ';'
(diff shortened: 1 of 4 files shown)


.appveyor.yml
```diff
@@ -0,0 +1,88 @@
...
+
+on_failure:
+- cmd: C:\Python27\python.exe %APPVEYOR_BUILD_FOLDER%\scripts\test\appveyor-irc-notify.py irc.oftc.net:6697 tor-ci failure
+
+## Uncomment to print Remote Desktop Connection details for the build worker,
+## and pause the build indefinitely.
+on_finish:
+#- ps: $blockRdp = $true; iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
+
+## Disabled: The builds failed until I commented out this section. If you
```
saper: I would drop this completely. We might also consider going back to installation of libevent from the package.

.appveyor.yml
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+on_success:
+- cmd: C:\Python27\python.exe %APPVEYOR_BUILD_FOLDER%\scripts\test\appveyor-irc-notify.py irc.oftc.net:6697 tor-ci success
+
+on_failure:
+- cmd: C:\Python27\python.exe %APPVEYOR_BUILD_FOLDER%\scripts\test\appveyor-irc-notify.py irc.oftc.net:6697 tor-ci failure
+
+## Uncomment to print Remote Desktop Connection details for the build worker,
+## and pause the build indefinitely.
```
saper: "indefinitely" seems to be currently [60 minutes](https://www.appveyor.com/docs/how-to/rdp-to-build-worker/) - according to AppVeyor docs. This was 30 minutes some time ago.

I'm fine with removing all of this altogether or just linking to AppVeyor docs.

.appveyor.yml
```diff
@@ -0,0 +1,88 @@
+version: 1.0.{build}
```
saper: do want/need something better here, resembling actual to branches, maybe?

changes/ticket25549
```diff
@@ -0,0 +1,4 @@
+  o Minor features (continuous integration):
```
saper: This script still has some UTF-8 trouble, my name comes out sometimes quite strange.

.appveyor.yml
```diff
@@ -0,0 +1,88 @@
...
+            $env:libeventversionfile = @("${env:APPVEYOR_BUILD_FOLDER}", ".appveyor-libevent-version") -join '\'
+            $env:libeventversion = Get-Content "${env:libeventversionfile}" | Out-String
+            Set-Location ".."
+            Execute-Bash "git clone https://github.com/libevent/libevent"
+            Set-Location "libevent"
+            $env:buildlibevent = (Get-Item -Path ".\").FullName
+            Execute-Bash "git checkout -q ${env:libeventversion}"
+            Execute-Bash "./autogen.sh"
+            Execute-Bash "./configure --prefix=/${env:compiler_path} --host=${env:target} --disable-libevent-regress --disable-samples --disable-openssl"
+            Execute-Bash "make -j2 install"
```
saper: Do we want to keep building it from scratch? libevent packaged seemed to work previously?

saper:
Ouch, I have pushed your branch onto [my AppVeyor](https://ci.appveyor.com/project/saper/appveyor/build/job/rwwgddo0k41tc2c8) and I have run, again, into https://trac.torproject.org/projects/tor/ticket/26076 

Is it caused by the order make chooses to run the test scripts?

nmathewson:
I don't know why #26076 came back -- perhaps the changes in "directories.sh" were necessary after all?

.appveyor.yml
nmathewson: I'm confused about what to revert there.  It looks like the commit (a507a62dd637f2b1b7a18bc88c6af402a765e44a) that started building libevent from scratch also removed a bunch of other pacman targets?

changes/ticket25549
nmathewson: Hm. Any idea how to fix that?  I'm afraid I don't know.  If neither of us can figure it out, then we can open a ticket and hope somebody else gets it working.

.appveyor.yml
nmathewson: I don't know -- if there's something that other projects do, we can do it too?  The current version setting is not great, but it seems "good enough for v1".

nmathewson:
Actually, I think I believe your idea about directory creation sometime failing.

saper:
Oh, I just noticed that you managed to get it work without ba663097fe0dbf35c472f2ad8b5e311e6651f9e1 
I was getting the error with it as well so I don't think it improves things. 

The tests heavily depend on a temporary directory being unique, really empty or in expected state. I guess simple change in the order make decides to run targets breaks things.

.appveyor.yml
saper: Yeah it is basically reverting https://github.com/torproject/tor/commit/a507a62dd637f2b1b7a18bc88c6af402a765e44a but fortunately we need only `pacman libevent libevent-dev` without all the OpenSSL and (non-working) zstd stuff.

teor2345:
Closed on trac https://trac.torproject.org/projects/tor/ticket/25549

PR #122, opened 2018-05-25
Feature 8323 - GETINFO md/all

rl1987:
https://trac.torproject.org/projects/tor/ticket/8323

### DIFF
--- a/src/or/control.c
+++ b/src/or/control.c
@@ -2207,6 +2207,25 @@ getinfo_helper_dir(control_connection_t *control_conn,
         return -1;
       }
     }
+  } else if (!strcmp(question, "md/all")) {
+    const smartlist_t *nodes = nodelist_get_all_nodes();
+    if (!nodes) {
+      *errmsg = "Internal error";
+      return -1;
+    }
+
+    smartlist_t *microdescs = smartlist_new();
+
+    SMARTLIST_FOREACH_BEGIN(nodes, node_t *, n) {
+      if (n->md && n->md->body) {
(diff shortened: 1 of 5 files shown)


src/or/nodelist.c
```diff
@@ -891,6 +891,15 @@ node_get_by_hex_id(const char *hex_id, unsigned flags)
   return NULL;
 }
 
+const smartlist_t *
```
asn-d6: Hmm why do we need to define a new function for this? We already have `nodelist_get_list()` that does this.
See how it's used in other parts of the code. :)

src/or/control.c
```diff
@@ -2207,6 +2207,25 @@ getinfo_helper_dir(control_connection_t *control_conn,
         return -1;
       }
     }
+  } else if (!strcmp(question, "md/all")) {
+    const smartlist_t *nodes = nodelist_get_all_nodes();
+    if (!nodes) {
```
asn-d6: Hmm, I don't think we should return `Internal error` if there are no nodes/consensus. Instead we should do what we do for `ns/all` in `getinfo_helper_networkstatus()` and return an empty string.

That's also an artficat of the way your `nodelist_get_all_nodes()` function works. If you use the `nodelist_get_list()` function, you are not gonna have this problem.

rl1987: Fixed in 51641533b2c565e22e24b27bf5782d70802a051.

src/or/nodelist.c
rl1987: Fixed.

PR #123, opened 2018-05-27
Bug26214

mikeperry-tor:
Needs changes tweak. I think the stream-level sendme stuff has been there since the first Tor release but I'm not sure. If so, do I win the oldest Tor bugfix prize? ;)

### DIFF
--- a/src/or/relay.c
+++ b/src/or/relay.c
@@ -1752,8 +1752,7 @@ connection_edge_process_relay_cell(cell_t *cell, circuit_t *circ,
           circuit_resume_edge_reading(circ, layer_hint);
 
           /* We count circuit-level sendme's as valid delivered data because
-           * they are rate limited. Note that we cannot count stream
-           * sendme's because the other end could send as many as they like.
+           * they are rate limited.
            */
           if (CIRCUIT_IS_ORIGIN(circ)) {
             circuit_read_valid_data(TO_ORIGIN_CIRCUIT(circ),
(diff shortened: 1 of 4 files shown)


src/or/relay.c
```diff
@@ -1783,6 +1782,21 @@ connection_edge_process_relay_cell(cell_t *cell, circuit_t *circ,
...
+               "Unexpected stream sendme cell. Closing circ (window %d).",
+               conn->package_window);
+        return -END_CIRC_REASON_TORPROTOCOL;
+      }
+
+      /* At this point, the stream sendme is valid */
+      if (CIRCUIT_IS_ORIGIN(circ)) {
+        circuit_read_valid_data(TO_ORIGIN_CIRCUIT(circ),
+                                rh.length);
+      }
```
dgoulet-tor: Hmmm I wonder why we need to do this here. Considering that we can make this call a bit earlier within the circuit window check, won't that increment twice the bw counters on the circuit for the same cell?

mikeperry-tor: The first call to circuit_read_valid_data() is inside the code block for the !rh.stream_id. That code block handles circuit level SENDMEs (which have stream ID 0), and it returns before the code that handles stream-level SENDMEs executes (which have a stream ID corresponding to the connection they are operating on). This patch only alters the stream-level SENDME branch.

teor2345:
Waiting for backport on trac

PR #124, opened 2018-05-30
Recompute some consensus information after detecting clock skew.

asn-d6:
Fixes bug #24977 and potentially other clock-skew bugs too.

The issue is that when we start up with a skewed clock in the past, we first
fetch a consensus and then we ignore it because it's not live. Afterwards, when
the clock gets fixed, we never re-read the consensus to calculate metadata
based on the right time.

This was causing troubles with HSv3 because we were left with a broken voting
schedule (that was calculated with a skewed clock).

### DIFF
--- a/src/or/main.c
+++ b/src/or/main.c
@@ -109,6 +109,7 @@
 #include "tor_api.h"
 #include "tor_api_internal.h"
 #include "util_process.h"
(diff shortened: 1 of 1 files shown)


teor2345:
Closed on trac

PR #126, opened 2018-05-31
Ticket3723

juga0:


src/or/dirserv.c
```diff
@@ -2621,6 +2623,8 @@ dirserv_read_measured_bandwidths(const char *from_file,
     return -1;
   }
 
+  smartlist_add_asprintf(bwlist_headers, "timestamp=%ld", file_time);
```
teor2345: bwlist_headers is sometimes NULL, but it can't be NULL in this function call. This might cause other unit tests to crash.

src/or/dirserv.c
```diff
@@ -2632,6 +2636,14 @@ dirserv_read_measured_bandwidths(const char *from_file,
         dirserv_cache_measured_bw(&parsed_line, file_time);
         if (measured_bw_line_apply(&parsed_line, routerstatuses) > 0)
           applied_lines++;
+      } else {
+        /* If line does not contain the header separator and it is key_value,
+         * it is probably a KeyValue header.*/
+        if (strcmp(line, "====\n") != 0 &&
```
teor2345: The terminator is 5 =, but this line has 4 =
https://gitweb.torproject.org/torspec.git/tree/bandwidth-file-spec.txt#n88

The header ends with the terminator, or the first valid relay bandwidth entry:
https://gitweb.torproject.org/torspec.git/tree/bandwidth-file-spec.txt#n235
But this code will also add any malformed relay entries to the bandwidth-file line, which is not what we want.
Please stop adding header lines when the terminator is read, or when the first valid relay bandwidth line is read.

Please also add a limit to the length of bwlist_headers. I don't expect we will ever have more than 50 headers. Stop adding headers when the limit is reached.

src/or/dirauth/dirvote.c
```diff
@@ -248,8 +249,12 @@ format_networkstatus_vote(crypto_pk_t *private_signing_key,
       params = smartlist_join_strings(v3_ns->net_params, " ", 0, NULL);
     else
       params = tor_strdup("");
-
     tor_assert(cert);
+    if (v3_ns->bwlist_headers)
+      bwlist_headers = smartlist_join_strings(v3_ns->bwlist_headers, " ", 0,
+                                              NULL);
```
teor2345: Please add a limit to the length of the bwlist_headers string. I don't expect we will ever have more than 1024 (~50*20) bytes of headers. Truncate the string when the limit is reached.

src/or/dirauth/dirvote.c
```diff
@@ -4518,6 +4527,7 @@ dirserv_generate_networkstatus_vote_obj(crypto_pk_t *private_key,
                            options->ConsensusParams, NULL, 0, 0);
     smartlist_sort_strings(v3_out->net_params);
   }
+  v3_out->bwlist_headers = bwlist_headers;
```
teor2345: v3_out->bwlist_headers is set, but never freed. When the vote is freed, you must free both the smartlist and the strings in it.

src/test/test_dir.c
```diff
@@ -1566,18 +1566,55 @@ test_dir_measured_bw_kb(void *arg)
...
+
+  /* And test bwlist_headers generation for dirvote.c */
+  tor_asprintf(&content, "%ld\n%s", timestamp, v110_header_lines);
+  write_str_to_file(fname, content, 0);
+  tor_free(content);
+  tt_int_op(0, OP_EQ, dirserv_read_measured_bandwidths(fname, NULL,
+                                                       bwlist_headers));
+
+  /* The bwlist_headers str that should get generated by the previous
+   * v110v110_header_lines */
```
teor2345: There is an extra "v110" on this line.

src/test/test_dir.c
```diff
@@ -1566,18 +1566,55 @@ test_dir_measured_bw_kb(void *arg)
...
   expect_log_msg("Empty bandwidth file\n");
 
+  /* Test v1.1.0 headers */
+  const char *v110_header_lines=
+    "version=1.1.0\n"
+    "software=sbws\n"
+    "software_version=0.1.0\n"
+    "generator_started=2018-05-08T16:13:25\n"
+    "earliest_bandwidth=2018-05-08T16:13:26\n"
+    "====\n";
```
teor2345: This terminator is wrong, see above.

src/test/test_dir.c
```diff
@@ -1566,18 +1566,55 @@ test_dir_measured_bw_kb(void *arg)
...
 
-  fname = tor_strdup(get_fname("V3BandwidthsFile"));
   /* Test an empty file */
   write_str_to_file(fname, "", 0);
   setup_capture_of_logs(LOG_WARN);
-  tt_int_op(-1, OP_EQ, dirserv_read_measured_bandwidths(fname, NULL));
+  tt_int_op(-1, OP_EQ, dirserv_read_measured_bandwidths(fname, NULL, NULL));
   expect_log_msg("Empty bandwidth file\n");
 
+  /* Test v1.1.0 headers */
```
teor2345: Please add another test for v1.0.0 headers. (That is, just a timestamp.)
Please add a test for v1.1.0 headers with no terminator (it is a SHOULD in the spec).
Please add a test for all 3 kinds of headers (v100, v110-terminator, v110-no-terminator) with relay lines.

You can also add tests with malformed relay lines if you want: any malformed lines before the terminator (if present) or the first valid relay line will end up being treated as headers.

src/or/dirserv.c
```diff
@@ -2632,6 +2636,14 @@ dirserv_read_measured_bandwidths(const char *from_file,
         dirserv_cache_measured_bw(&parsed_line, file_time);
         if (measured_bw_line_apply(&parsed_line, routerstatuses) > 0)
           applied_lines++;
+      } else {
```
teor2345: Also check if bwlist_headers is NULL here.

src/test/test_dir.c
```diff
@@ -1566,18 +1566,55 @@ test_dir_measured_bw_kb(void *arg)
...
+  smartlist_t *bwlist_headers = smartlist_new();
+  char *bwlist_headers_str = NULL;
+  char *out_bwlist_headers_str = NULL;
 
-  fname = tor_strdup(get_fname("V3BandwidthsFile"));
   /* Test an empty file */
   write_str_to_file(fname, "", 0);
   setup_capture_of_logs(LOG_WARN);
-  tt_int_op(-1, OP_EQ, dirserv_read_measured_bandwidths(fname, NULL));
+  tt_int_op(-1, OP_EQ, dirserv_read_measured_bandwidths(fname, NULL, NULL));
```
teor2345: Please also add tests for a non-empty file with a NULL bwlist_headers, because the code calls:
dirserv_read_measured_bandwidths(options->V3BandwidthsFile, NULL, NULL);

juga0:
Reminder to myself: s/bandwidth-file/bandwidth-file-headers

src/test/test_dir.c
juga0: I think i should move the 1st comparison of this test (bandwidth file empty) to https://gitweb.torproject.org/tor.git/tree/src/test/test_dir.c#n1649, that has been recently merged, and the name of this test to test_dir_dirserv_read_measured_bandwidths_bwlist_headers, since what we want to check here is the generation of the bwlist_headers.

src/or/dirauth/dirvote.c
juga0: Where is the vote freed?, this sounds like the place https://gitweb.torproject.org/tor.git/tree/src/or/dirauth/dirvote.c#n3616, but not to free bwlist_headers

teor2345: The vote is a networkstatus_t, which is freed by networkstatus_vote_free_ at  https://gitweb.torproject.org/tor.git/tree/src/or/networkstatus.c#n321

src/test/test_dir.c
teor2345: Sounds like a good idea to me!

juga0:
Closing because rebased to master and created #194

PR #130, opened 2018-06-02
Move dirauth module from src/or/dirauth to src/dirauth.

ahf:
This patch moves the directory authority module from src/or/dirauth to
src/dirauth to avoid having too deeply nested directory structures.

See: https://bugs.torproject.org/26270


teor2345:
Merge ready on trac

PR #136, opened 2018-06-04
#26284: Avoid out-of-bounds smartlist access in protover_compute_vote()

rl1987:
https://trac.torproject.org/projects/tor/ticket/26284

### DIFF
--- a/src/or/protover.c
+++ b/src/or/protover.c
@@ -637,6 +637,9 @@ char *
 protover_compute_vote(const smartlist_t *list_of_proto_strings,
                       int threshold)
 {
+  if (smartlist_len(list_of_proto_strings) == 0)
+    return tor_strdup("");
+
   smartlist_t *all_entries = smartlist_new();
 
   // First, parse the inputs and break them into singleton entries.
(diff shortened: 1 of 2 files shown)


src/or/protover.c
```diff
@@ -668,7 +671,9 @@ protover_compute_vote(const smartlist_t *list_of_proto_strings,
 
   // Now find all the strings that appear at least 'threshold' times.
   smartlist_t *include_entries = smartlist_new();
-  const char *cur_entry = smartlist_get(all_entries, 0);
+  const char *cur_entry = smartlist_len(all_entries) > 0 ?
```
asn-d6: Perhaps instead of setting cur_entry to `NULL` here, we could just return `tor_strdup("")` like you did above? I think that would simplify analysis, since now we need to analyze what happens with `contract_protocol_list()` and an empty smartlist? Or alternatively you could add another check in`contract_protocol_list()` which returns an allocated empty string if it receives an empty smartlist?

I think the first approach is easier but you would need to rearrange the error checking a bit, so that you can free memor yin case where `smartlist_len(all_entries)` is not strictly positive. I would do that by introducing a `done` label and doing a `goto done` if an empty smartlist is found.

What do you think?

rl1987: Sounds good. I did something like this in 57cdfb2a2d8ebca302351758735d46a578f4f7e5 - return empty string if `all_entries` turn out to be empty and also early bailout in `contract_protocol_list()` with empty string if argument turns out to be empty smartlist.

PR #140, opened 2018-06-09
Trac #21346: Only choose IPv6 exits when NoIPv4Traffic is set as a SocksPort flag

neelchauhan:
Pull request for [Tor Trac #21346](https://trac.torproject.org/projects/tor/ticket/21346).

### DIFF
--- a/src/or/circuituse.c
+++ b/src/or/circuituse.c
@@ -94,17 +94,19 @@ circuit_matches_with_rend_stream(const edge_connection_t *edge_conn,
  */
 static int
 circuit_is_acceptable(const origin_circuit_t *origin_circ,
-                      const entry_connection_t *conn,
+                      const entry_connection_t *entry_conn,
                       int must_be_open, uint8_t purpose,
                       int need_uptime, int need_internal,
                       time_t now)
 {
   const circuit_t *circ = TO_CIRCUIT(origin_circ);
   const node_t *exitnode;
+  const connection_t *conn = ENTRY_TO_CONN(entry_conn);
   cpath_build_state_t *build_state;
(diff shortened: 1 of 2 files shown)


src/or/circuituse.c
```diff
@@ -187,35 +190,47 @@ circuit_is_acceptable(const origin_circuit_t *origin_circ,
...
+    /* If we can't exit on IPv4, check if the node supports exiting on IPv6. */
+    if (!entry_conn->entry_cfg.ipv4_traffic &&
+        conn->purpose == CONN_TYPE_EXIT) {
+      tor_addr_t addr6;
+      tor_addr_make_null(&addr6, AF_INET6);
+      int r = compare_tor_addr_to_node_policy(&addr6,
+                                              entry_conn->socks_request->port,
+                                              exitnode);
+      if (r == ADDR_POLICY_REJECTED)
+        return 0;
```
dgoulet-tor: I think all this should go in `connection_ap_can_use_exit()`. But also, how is this different from what there is already in that function? Maybe we don't do the check correctly?

src/or/circuituse.c
```diff
@@ -94,17 +94,19 @@ circuit_matches_with_rend_stream(const edge_connection_t *edge_conn,
  */
 static int
 circuit_is_acceptable(const origin_circuit_t *origin_circ,
-                      const entry_connection_t *conn,
+                      const entry_connection_t *entry_conn,
```
dgoulet-tor: This is a fine change but if you do it, please do it in its separate commit so we can distinguish refactoring/renaming with code change. Thanks!

PR #146, opened 2018-06-13
Bug24977 final

asn-d6:


src/or/voting_schedule.c
```diff
@@ -138,14 +142,32 @@ voting_schedule_t voting_schedule;
...
                       sizeof(voting_schedule))) {
-    voting_schedule_recalculate_timing(get_options(), time(NULL));
+    need_to_recalculate_voting_schedule = true;
+  }
+
+  /* Also make sure we are not using an outdated voting schedule. If we have a
+   * newer consensus, make sure we recalculate the voting schedule. */
+  networkstatus_t *consensus = networkstatus_get_live_consensus(approx_time());
+  if (consensus &&
+      consensus->valid_after != voting_schedule.live_consensus_valid_after) {
```
dgoulet-tor: Maybe we want to use a `else if` somehow so we avoid all this code if the first condition was hit which would recalculate the voting schedule in the first place?

src/or/voting_schedule.c
```diff
@@ -138,14 +142,32 @@ voting_schedule_t voting_schedule;
...
    * voting schedule can lead to bugs. */
   if (tor_mem_is_zero((const char *) &voting_schedule,
                       sizeof(voting_schedule))) {
-    voting_schedule_recalculate_timing(get_options(), time(NULL));
+    need_to_recalculate_voting_schedule = true;
+  }
+
+  /* Also make sure we are not using an outdated voting schedule. If we have a
+   * newer consensus, make sure we recalculate the voting schedule. */
+  networkstatus_t *consensus = networkstatus_get_live_consensus(approx_time());
```
dgoulet-tor: I would `const` that if possible.

Also, `approx_time()` is used twice in this function so probably a `time_t now` is desirable?

src/or/voting_schedule.c
```diff
@@ -138,14 +142,32 @@ voting_schedule_t voting_schedule;
...
+  if (consensus &&
+      consensus->valid_after != voting_schedule.live_consensus_valid_after) {
+    log_info(LD_DIR, "Voting schedule is outdated: recalculating (%d/%d)",
+             (int) consensus->valid_after,
+             (int) voting_schedule.live_consensus_valid_after);
+    need_to_recalculate_voting_schedule = true;
+  }
+
+  if (need_to_recalculate_voting_schedule) {
+    voting_schedule_recalculate_timing(get_options(), approx_time());
```
dgoulet-tor: Hold on a sec. Don't we have a circular loop here? `get_voting_schedule()` can reach this point and then `voting_schedule_recalculate_timing()` calls `get_voting_schedule()` again so we do this whole path over and over again?

I actually wonder if the right approach would be to have a getter for the `voting_schedule` static object that makes sure it is not out of date. So no where we should use that object directly but rather get the ref through the getter assuring us that it is updated?

src/or/nodelist.c
```diff
@@ -854,6 +865,29 @@ nodelist_assert_ok(void)
...
+ensure_nodelist_freshness(void)
+{
+  /* We don't even have a nodelist: this is a NOP. */
+  if (!the_nodelist) {
+    return;
+  }
+
+  /* We don't even have a recent consensus: this is a NOP */
+  networkstatus_t *ns = networkstatus_get_live_consensus(approx_time());
+  if (!ns) {
```
dgoulet-tor: Seems we can't call this without a live consensus. Should we BUG() here?

src/or/nodelist.h
```diff
@@ -29,6 +29,7 @@ const node_t *node_get_by_hex_id(const char *identity_digest,
 node_t *nodelist_set_routerinfo(routerinfo_t *ri, routerinfo_t **ri_old_out);
 node_t *nodelist_add_microdesc(microdesc_t *md);
 void nodelist_set_consensus(networkstatus_t *ns);
+void ensure_nodelist_freshness(void);
```
dgoulet-tor: I propose we namespace this one starting with `nodelist_`.

src/or/voting_schedule.c
asn-d6: Hmmm, trying to understand this better: How can `get_voting_schedule()` reach `voting_schedule_get_next_valid_after_time()`? They seem disconnected to me.

No?

src/or/voting_schedule.c
asn-d6: Good point. Took the goto approach in 82bb4f34 to avoid the extra indentation of `else if`. Let me know if you hate it, and I will just use `else if`.

src/or/voting_schedule.c
asn-d6: Done in 222c5cf5.

src/or/nodelist.c
asn-d6: Good point. Done in aa951c72dc.

src/or/nodelist.h
asn-d6: Good point. Done in aa951c72dc.

src/or/voting_schedule.c
dgoulet-tor: ACK

src/or/voting_schedule.c
dgoulet-tor: Oh oops... my brain failed to parse that `voting_schedule_get_start_of_next_interval()` (in get_voting_schedule()) `!= voting_schedule_get_next_valid_after_time()` :P...

NVM.

src/or/nodelist.c
dgoulet-tor: ACK.

src/or/nodelist.h
dgoulet-tor: ACK.

PR #148, opened 2018-06-13
test_entrynodes: fix a GCC warning

nmathewson:
Some versions of GCC complain that the bfn_mock_node_get_by_id
function might return NULL, but we're assuming that it won't.
(We're assuming it won't return NULL because we know in the tests
that we're passing it valid IDs.)

To make GCC happy, tt_assert() that each node_t is set before using
it.

Fixes a second case of bug26269; bugfix on 0.3.0.1-alpha.


nmathewson:
Upstream ticket closed; this patch and its cousin were merged into 0.2.9 and 0.3.2

PR #150, opened 2018-06-14
Bug25552 ope draft

asn-d6:


src/or/hs_service.c
```diff
@@ -2506,16 +2511,82 @@ increment_descriptor_revision_counter(hs_descriptor_t *hs_desc)
...
+   * guaranteed to be a positive value since we need the SRV to start making a
+   * descriptor (so that we know where to upload it).
+   *
+   * Depending on whether we are building the current or the next descriptor,
+   * services use a different SRV value. See [SERVICEUPLOAD] of
+   * rend-spec-v3.txt for more details. */
+  if (is_current) {
+    srv_start = sr_state_get_start_time_of_previous_protocol_run(now);
+  } else {
+    srv_start = sr_state_get_start_time_of_current_protocol_run(now);
```
dgoulet-tor: Is there a mix up here? Seems counter intuitive to have `is_current` using the _previous_ protocol run start time? If that is correct, we should explain why this is correct.

src/or/hs_service.c
```diff
@@ -1755,7 +1755,6 @@ service_desc_schedule_upload(hs_service_descriptor_t *desc,
 /* Update the given descriptor from the given service. The possible update
```
dgoulet-tor: General comment: This commit is VERY satisfying :).

src/or/hs_service.c
```diff
@@ -2506,16 +2511,82 @@ increment_descriptor_revision_counter(hs_descriptor_t *hs_desc)
...
+    crypto_digest_add_bytes(digest, (char*)eph_privkey->seckey,
+                            sizeof(eph_privkey->seckey));
+    crypto_digest_get_digest(digest, (char *)key, sizeof(key));
+    crypto_digest_free(digest);
+  }
+
+  { /* Now encrypt the revision counter! */
+    crypto_ope_t *ope = NULL;
+    ope = crypto_ope_new(key);
+    rev_counter = crypto_ope_encrypt(ope, (int) seconds_since_start_of_srv);
```
dgoulet-tor: `UINT64_MAX` is sent back on an "invalid" input. I would `BUG()` on it for extra safety because if we start doing that but not noticing, we'll wonder for a long time why descriptor are getting refused at the HSDir :S...

src/or/hs_service.c
asn-d6: Not a mixup. It's just the way prop224 spec works. See [SERVICEUPLOAD]:

For is_current (aka first descriptor), Tor always uses previous SRV and hence
the previous protocol run start time:

```
When a service is in the time segment between a new time period a new SRV
(i.e. the segments drawn with "-"), it uses the previous time period and
previous SRV for uploading its first descriptor

Now if a service is in the time segment between a new SRV and a new time
period (i.e. the segments drawn with "=") it uses the current time period
and the previous SRV for its first descriptor
```

whereas for `!is_current` (aka second descriptor), it always uses current SRV and
hence start time of current protocol run.

src/or/hs_service.c
asn-d6: Done in 2902c2d9d0

src/common/crypto_ope.h
```diff
@@ -14,8 +14,16 @@
...
-/** Largest value that can be passed to crypto_ope_encrypt() */
-#define OPE_INPUT_MAX 131072
+/** Largest value that can be passed to crypto_ope_encrypt().
+
+ *  The current max value (220k seconds) stands for about 61 hours. The
+ *  rationale here is as follows: The rev counter is the time of seconds since
+ *  the start of an SRV period. SRVs are useful for about 48 hours (that's how
+ *  long they stick on the consensus). Let's also add 12 hours of drift for
+ *  clock skewed clients stuck with an old consensus, and we arrive to 60
+ *  hours.
```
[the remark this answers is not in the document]
asn-d6: Hmm, that comment might have been a bit misleading. I meant to say "clock
skewed services" and not "clock skewed clients". I pumped it to 60 hours
because I'm concerned about clock skewed services using outdated consensus. In particular, an SRV stays for 48 hours in consensuses, and a service could potentially have an outdated consensus and keep on using it.

In general, 60 hours VS 48 hours doesn't make a huge difference in terms of
OPE, so perhaps we can leave it at 60 and be on the safe side.

nmathewson:
A version of this was squashed and merged

PR #169, opened 2018-06-21
Extract easy common libs

nmathewson:


.gitignore
```diff
@@ -169,10 +166,14 @@ uptime-*.json
 # /src/lib
 /src/lib/libtor-compress.a
 /src/lib/libtor-compress-testing.a
+/src/lib/libtor-crypt-ops.a
+/src/lib/libtor-crypt-ops-testing.a
```
dgoulet-tor: Hmmm nitpick but why not `crypto` here? Do we plan to have something else than crypto operation a.k.a `crypto_ops` ?

I'm asking because if I go into tor source tree looking for the crypto library, I'm not sure `crypto_ops` will stand out as "oh yes of course, crypto operations"...

PR #172, opened 2018-06-22
#19506: Tool to inspect id signing certs (print expiration date)

rl1987:
https://trac.torproject.org/projects/tor/ticket/19506


rl1987:
Obsoleted by #216, which was merged.

PR #173, opened 2018-06-22
#21349: Refactor long functions in entrynodes.c

rl1987:
https://trac.torproject.org/projects/tor/ticket/21349

### DIFF
--- a/changes/ticket21349
+++ b/changes/ticket21349
@@ -1,0 +1,6 @@
+  o Code simplification and refactoring:
+    - Split sampled_guards_update_from_consensus() and
+      select_entry_guard_for_circuit() into subfunctions.
+      In entry_guards_update_primary() unite
+      three smartlist enumerations into one and move smartlist
+      comparison code out of the function. Closes ticket 21349.
--- a/src/common/container.h
+++ b/src/common/container.h
@@ -93,6 +93,24 @@ static inline void smartlist_swap(smartlist_t *sl, int idx1, int idx2)
   }
 }
 
+/** Return true if there is shallow equality between smartlists -
+ * i.e. all indices correspond to exactly same object (pointer
+ * values are matching). Otherwise, return false.
+ */
+static inline int smartlist_shallow_eq(const smartlist_t *s1,
+                                       const smartlist_t *s2)
+{
+  if (smartlist_len(s1) != smartlist_len(s2))
+    return 0;
+
+  for (int i = 0; i < smartlist_len(s1); i++) {
+    if (smartlist_get(s1, i) != smartlist_get(s2, i))
+      return 0;
+  }
+
+  return 1;
+}
+
 void smartlist_del(smartlist_t *sl, int idx);
 void smartlist_del_keeporder(smartlist_t *sl, int idx);
 void smartlist_insert(smartlist_t *sl, int idx, void *val);
--- a/src/common/container.h
+++ b/src/common/container.h
@@ -93,23 +93,8 @@ static inline void smartlist_swap(smartlist_t *sl, int idx1, int idx2)
   }
 }
 
-/** Return true if there is shallow equality between smartlists -
- * i.e. all indices correspond to exactly same object (pointer
- * values are matching). Otherwise, return false.
- */
-static inline int smartlist_shallow_eq(const smartlist_t *s1,
-                                       const smartlist_t *s2)
-{
-  if (smartlist_len(s1) != smartlist_len(s2))
-    return 0;
-
-  for (int i = 0; i < smartlist_len(s1); i++) {
-    if (smartlist_get(s1, i) != smartlist_get(s2, i))
-      return 0;
-  }
-
-  return 1;
-}
+int smartlist_shallow_eq(const smartlist_t *s1,
+                         const smartlist_t *s2);
 
 void smartlist_del(smartlist_t *sl, int idx);
 void smartlist_del_keeporder(smartlist_t *sl, int idx);
--- a/src/or/entrynodes.c
+++ b/src/or/entrynodes.c
@@ -142,9 +142,6 @@
 #include "or/node_st.h"
 #include "or/origin_circuit_st.h"
 
-#define REMOVE_UNLISTED_GUARDS_AFTER \
-  ((time_t)(get_remove_unlisted_guards_after_days() * 86400))
-
 /** A list of existing guard selection contexts. */
 static smartlist_t *guard_contexts = NULL;
 /** The currently enabled guard selection context. */
@@ -404,6 +401,13 @@ get_remove_unlisted_guards_after_days(void)
                                  DFLT_REMOVE_UNLISTED_GUARDS_AFTER_DAYS,
                                  1, 365*10);
 }
+
+static time_t
+get_remove_unlisted_guards_after_seconds(void)
+{
+  return get_remove_unlisted_guards_after_days() * 24 * 60 * 60;
+}
+
 /**
  * We remove unconfirmed guards from the sample after this many days,
  * regardless of whether they are listed or unlisted.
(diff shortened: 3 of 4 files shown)


src/or/entrynodes.c
```diff
@@ -142,6 +142,9 @@
 #include "or/node_st.h"
 #include "or/origin_circuit_st.h"
 
+#define REMOVE_UNLISTED_GUARDS_AFTER \
```
asn-d6: Hmm, not sure if I like this macro doing multiplications and casting to time_t. Suggestion: Let's modify `get_remove_unlisted_guards_after_days()` to return exactly what we want, and let's do the multiplication in the log of `sampled_guards_prune_obsolete_entries()`. What do you think?

src/or/entrynodes.c
```diff
@@ -1231,43 +1234,38 @@ entry_guard_is_listed,(guard_selection_t *gs, const entry_guard_t *guard))
   }
 }
 
-/**
- * Update the status of all sampled guards based on the arrival of a
- * new consensus networkstatus document.  This will include marking
- * some guards as listed or unlisted, and removing expired guards. */
-STATIC void
-sampled_guards_update_from_consensus(guard_selection_t *gs)
+/** Enumerate <b>sampled_entry_guards</b> smartlist in <b>gs</b>.
```
asn-d6: Not exactly the right function documentation style. For multi-line comments perhaps leave the `/**` line empty. See how other functions do it.

src/or/entrynodes.c
```diff
@@ -1231,43 +1234,38 @@ entry_guard_is_listed,(guard_selection_t *gs, const entry_guard_t *guard))
...
-  // It's important to use only a live consensus here; we don't want to
-  // make changes based on anything expired or old.
-  if (live_consensus_is_missing(gs)) {
-    log_info(LD_GUARD, "Not updating the sample guard set; we have "
-             "no live consensus.");
-    return;
-  }
-  log_info(LD_GUARD, "Updating sampled guard status based on received "
-           "consensus.");
+  tor_assert(n_changes);
```
asn-d6: Suggestion: Instead of taking an `n_changes` pointer argument, let's switch this function (and the other one) from `void` to `int`, and have it return the value of `n_changes`. I think that will be cleaner, since there won't be a need to do pointer operations.

src/or/entrynodes.c
```diff
@@ -1343,12 +1357,51 @@ sampled_guards_update_from_consensus(guard_selection_t *gs)
...
+             "no live consensus.");
+    return;
+  }
+  log_info(LD_GUARD, "Updating sampled guard status based on received "
+           "consensus.");
+
+  int n_changes = 0;
+
+  /* First: Update listed/unlisted. */
+  sampled_guards_update_listedness(gs, &n_changes);
```
asn-d6: Suggestion: Rename `sampled_guards_update_listedness()` to `sampled_guards_update_consensus_presence()`, since 'listedness' is not actually a word. Let me know if you like this, or you prefer something else, or prefer to leave it as is.

src/or/entrynodes.c
```diff
@@ -1343,12 +1357,51 @@ sampled_guards_update_from_consensus(guard_selection_t *gs)
...
+
+  const time_t remove_if_unlisted_since =
```
asn-d6: Perhaps we can do these computations inside `sampled_guards_prune_obsolete_entries()`? Or you like them here?

src/common/container.h
```diff
@@ -93,6 +93,24 @@ static inline void smartlist_swap(smartlist_t *sl, int idx1, int idx2)
   }
 }
 
+/** Return true if there is shallow equality between smartlists -
```
asn-d6: Perhaps we should put this in `container.c`? That's where the other `_eq()` functions are. I'm not sure if I like functions in .h files, but it seems to be the case in `container.h`, so if you like it like this feel free to leave it as is.

src/common/container.h
```diff
@@ -93,6 +93,24 @@ static inline void smartlist_swap(smartlist_t *sl, int idx1, int idx2)
...
+ * i.e. all indices correspond to exactly same object (pointer
+ * values are matching). Otherwise, return false.
+ */
+static inline int smartlist_shallow_eq(const smartlist_t *s1,
+                                       const smartlist_t *s2)
+{
```
asn-d6: Since we opted for a generic function here we should also check that `s1` and `s2` are non-NULL and `return 0` if any of them are NULL.

src/common/container.h
```diff
@@ -93,6 +93,24 @@ static inline void smartlist_swap(smartlist_t *sl, int idx1, int idx2)
   }
 }
 
+/** Return true if there is shallow equality between smartlists -
+ * i.e. all indices correspond to exactly same object (pointer
+ * values are matching). Otherwise, return false.
+ */
+static inline int smartlist_shallow_eq(const smartlist_t *s1,
```
asn-d6: Optional suggestion: We can make this function return `bool` instead of `int`.

src/or/entrynodes.c
```diff
@@ -2082,34 +2086,86 @@ select_entry_guard_for_circuit(guard_selection_t *gs,
...
+  if (need_descriptor)
+    flags |= SAMPLE_EXCLUDE_NO_DESCRIPTOR;
+  chosen_guard = sample_reachable_filtered_entry_guards(gs,
+                                                 rst,
+                                                 SAMPLE_EXCLUDE_CONFIRMED |
+                                                 SAMPLE_EXCLUDE_PRIMARY |
+                                                 SAMPLE_EXCLUDE_PENDING |
+                                                 flags);
+  if (chosen_guard == NULL) {
+    log_info(LD_GUARD, "Absolutely no sampled guards were available. "
```
asn-d6: Suggestion: Perhaps we can pull this `mark all guards as maybe reachable` logic out of this function, and into the end of `select_entry_guard_for_circuit()` to make it clear that this is the end of the algorithm. We could do this in a separate commit in the end, if you like the idea.

changes/ticket21349
```diff
@@ -0,0 +1,6 @@
+  o Code simplification and refactoring:
```
asn-d6: General comment: This branch needs to be rebased to the latest master, which contains the various modularization structure improvements. Let me know if you don't want to do this, and I will do it myself.

src/or/entrynodes.c
rl1987: In 64685c37b670cd539848377f8c6febc22fd75cb4 I made a function that returns the exact number we need here.

changes/ticket21349
rl1987: I will do this once I address all the things you mentioned.

src/or/entrynodes.c
rl1987: Renamed in 72da8b4be0a896bd90d2127ac67919cdbf383ead.

src/common/container.h
rl1987: Moved to container.c in cda8adc876b87443859577e54f080f56b8f5e583.

src/common/container.h
rl1987: Introduced some extra checks in cda8adc876b87443859577e54f080f56b8f5e583.

src/or/entrynodes.c
rl1987: Did so (with `size_t`) in eea7c816335766b7bb15b07b060ae9df7cf20e2e.

src/or/entrynodes.c
rl1987: Did so in d02b2f1b0cfe7bd251bd347ba293cf11cd7ea670.

src/common/container.h
rl1987: We generally don't use `bool` in the codebase, so I'll rather use `int`, like we do pretty much always.

src/common/container.h
asn-d6: LGTM

src/common/container.h
asn-d6: stdbool was only recently added to the codebase that's why it's not yet used extensively. In any case, I'm fine with int as well.

src/common/container.h
asn-d6: LGTM

src/or/entrynodes.c
asn-d6: LGTM.

src/or/entrynodes.c
asn-d6: LGTM.

src/or/entrynodes.c
asn-d6: LGTM.

src/or/entrynodes.c
asn-d6: Sounds good!

src/or/entrynodes.c
asn-d6: LGTM!

tlyu:
Superseded by #246.

PR #174, opened 2018-06-22
Log dependencies

nmathewson:


teor2345:
What's the trac ticket here?

src/lib/container/bloomfilt.c
```diff
@@ -0,0 +1,49 @@
...
+ * Copyright (c) 2007-2018, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
+ * \file container.c
+ * \brief Implements a smartlist (a resizable array) along
+ * with helper functions to use smartlists.  Also includes
+ * hash table implementations of a string-to-void* map, and of
+ * a digest-to-void* map.
+ **/
```
dgoulet-tor: This whole section is wrong.

src/lib/container/bloomfilt.h
```diff
@@ -0,0 +1,58 @@
...
+#include "lib/cc/torint.h"
+#include "lib/container/bitarray.h"
+#include "siphash.h"
+
+/** A set of digests, implemented as a Bloom filter. */
+typedef struct {
+  int mask; /**< One less than the number of bits in <b>ba</b>; always one less
+             * than a power of two. */
+  bitarray_t *ba; /**< A bit array to implement the Bloom filter. */
+} digestset_t;
```
dgoulet-tor: I think, this can be done after, we should namespace these with the `bloomfilt` name even thought "digest set" is technically valid.

src/lib/container/order.c
```diff
@@ -0,0 +1,51 @@
...
+ * Copyright (c) 2007-2018, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
+ * \file container.c
+ * \brief Implements a smartlist (a resizable array) along
+ * with helper functions to use smartlists.  Also includes
+ * hash table implementations of a string-to-void* map, and of
+ * a digest-to-void* map.
+ **/
```
dgoulet-tor: Bad comments, I guess copy pasting :). Seems many files in `lib/container/` suffer from this.

src/include.am
```diff
@@ -5,6 +5,7 @@ include src/lib/ctime/include.am
 include src/lib/compress/include.am
 include src/lib/container/include.am
 include src/lib/crypt_ops/include.am
+include src/lib/defs/include.am
```
dgoulet-tor: What does `defs/` stands for here? "Definitions" I presume? Would we put all global constant in there or ?...

src/lib/string/include.am
```diff
@@ -0,0 +1,23 @@
+
+noinst_LIBRARIES += src/lib/libtor-string.a
+
+if UNITTESTS_ENABLED
+noinst_LIBRARIES += src/lib/libtor-string-testing.a
+endif
+
+src_lib_libtor_string_a_SOURCES =			\
+	src/lib/string/compat_ctype.c			\
+	src/lib/string/util_string.c			\
```
dgoulet-tor: I might have the same comment as `malloc` module here with this `util_string.c`.

.gitignore
```diff
@@ -173,6 +173,8 @@ uptime-*.json
 /src/lib/libtor-ctime-testing.a
 /src/lib/libtor-err.a
 /src/lib/libtor-err-testing.a
+/src/lib/libtor-intmath.a
```
dgoulet-tor: What about a `math` module instead and we put everything related to "math" in there (not only int stuff). I mean with this name, does the floating math functions apply to go in?

src/lib/wallclock/include.am
```diff
@@ -7,6 +7,7 @@ endif
 
 src_lib_libtor_wallclock_a_SOURCES =			\
 	src/lib/wallclock/approx_time.c			\
+	src/lib/wallclock/tm_cvt.c			\
```
dgoulet-tor: I'm not familiar with this naming, what does `tm_cvt` means here?

Makefile.am
```diff
@@ -42,6 +42,7 @@ TOR_UTIL_LIBS = \
 	src/common/libor.a \
         src/lib/libtor-log.a \
         src/lib/libtor-lock.a \
+        src/lib/libtor-fdio.a \
```
dgoulet-tor: Quick question: Is the goal of this library to also get the `tor_open_*` family of functions (currently mostly in `compat.c`)?

If yes, I'm wondering if `libtor-io.a` wouldn't be better and we could also bring in all the network low level functions? Or for that you had in mind a `libtor-network.` kind of separation?

src/include.am
nmathewson: yes, definitions.  But I think only top-level definitions that need to be really global.

src/common/util.h
```diff
@@ -23,6 +23,7 @@
 #include <sys/stat.h>
 #endif
 #include "lib/err/torerr.h"
+#include "lib/malloc/util_malloc.h"
```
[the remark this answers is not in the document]
nmathewson: I think malloc.h is fine, but not util.h -- that would mean we had a bunch of files named util.h, and I don't want to have two headers or c files with the same name.

.gitignore
nmathewson: no, higher-level stuff should go in another module.

This one is for low-level integer math.

src/lib/wallclock/include.am
nmathewson: tm is "struct tm"; cvt is convert.

Makefile.am
nmathewson: No, tor_open_* needs to be at a higher level, or else it will cause a circular dependency.  Those functions have a lot of complexity, and have the potential to log, and require the sandbox code.

PR #178, opened 2018-06-25
Update tor-rust-dependencies from isislovecruft

liang55:


teor2345:
Hi, thanks for this fix!

But we need to merge @isislovecruft's https://github.com/isislovecruft/tor-rust-dependencies into https://git.torproject.org/tor-rust-dependencies instead.

Is there a track ticket for this fix, or should we open one?

PR #180, opened 2018-06-26
#3569 part 1: Rework SOCKS5 server code with trunnel

rl1987:
https://trac.torproject.org/projects/tor/ticket/3569


src/or/proto_socks.c
```diff
@@ -218,6 +218,119 @@ process_socks4_request(socks_request_t *req, int is_socks4a,
...
+
+  *have_no_auth = 0;
+  *have_user_pass = 0;
+
+  for (size_t i = 0; i < n_methods; i++) {
+    uint8_t method = socks5_client_version_get_methods(trunnel_req,
+                                                       i);
+
+    if (method == SOCKS_USER_PASS) {
+      *have_user_pass = 1;
```
nmathewson: I would be more comfortable if this function also set these values to 0 when they are supposed to be false.

src/or/proto_socks.c
```diff
@@ -218,6 +218,119 @@ process_socks4_request(socks_request_t *req, int is_socks4a,
...
+    } else if (method == SOCKS_NO_AUTH) {
+      *have_no_auth = 1;
+    }
+  }
+  
+  end:
+  *drain_out = (size_t)parsed;
```
nmathewson: This shouldn't happen when parsed < 0.

src/or/proto_socks.c
```diff
@@ -287,6 +403,45 @@ fetch_from_buf_socks(buf_t *buf, socks_request_t *req,
...
     res = 1;
     goto end;
+  } else if (socks_version == 5) {
+    if (datalen < 2) { /* version and another byte */
+      res = 0;
+      goto end;
+    }
+
+    if (!req->got_auth) {
+
```
nmathewson: I'm not sure what's going on here; perhaps it will become clear later in the branch.

src/or/proto_socks.c
```diff
@@ -287,6 +403,45 @@ fetch_from_buf_socks(buf_t *buf, socks_request_t *req,
...
+    }
+
+    if (req->socks_version != 5) {
+      int have_user_pass, have_no_auth;
```
nmathewson: Shouldn't these be initialized to 0, given how they aren't set above?  I would suggest doing _both_.  (Initialize to 0, and set to 0 in parse function)

src/or/proto_socks.c
```diff
@@ -250,6 +256,8 @@ parse_socks5_methods_request(const uint8_t *raw_data, socks_request_t *req,
     goto end;
   }
 
+  *drain_out = (size_t)parsed;
```
nmathewson: Better!  But let's assert that parsed >= 0 in this case.

src/or/proto_socks.c
```diff
@@ -116,6 +119,8 @@ parse_socks4_request(const uint8_t *raw_data, socks_request_t *req,
...
```
nmathewson: We should assert that this value is >= 0. 

Also, what about the bytes from the socks4a hostname?  I think we should add them to drain_out too.

src/or/proto_socks.c
```diff
@@ -88,13 +88,16 @@ socks_request_free_(socks_request_t *req)
...
+                     size_t datalen, int *is_socks4a, size_t *drain_out)
 {
   // http://ss5.sourceforge.net/socks4.protocol.txt
   // http://ss5.sourceforge.net/socks4A.protocol.txt
   int res = 1;
   tor_addr_t destaddr;
 
+  tor_assert(is_socks4a);
+  tor_assert(drain_out);
+
```
nmathewson: We should initialize drain_out to 0, so that it can't wind up un-set.

src/or/proto_socks.c
nmathewson: (And let's make sure drain_out is initialized to 0 in this function too)

src/or/proto_socks.c
```diff
@@ -331,49 +338,13 @@ process_socks5_methods_request(socks_request_t *req, int have_user_pass,
...
 
-  raw_data = tor_malloc(datalen);
-  memset(raw_data, 0, datalen);
-
-  buf_peek(buf, (char *)raw_data, datalen);
-
-  raw_ptr = raw_data;
-
-  socks_version = (uint8_t)raw_data[0];
+  uint8_t socks_version = (uint8_t)raw_data[0];
```
nmathewson: As before: we can't do this unless we check datalen first.

src/or/proto_socks.c
```diff
@@ -331,49 +338,13 @@ process_socks5_methods_request(socks_request_t *req, int have_user_pass,
...
- *
- * If <b>safe_socks</b> is true, then reject unsafe socks protocols.
- *
- * If returning 0 or -1, <b>req->address</b> and <b>req->port</b> are
- * undefined.
- */
-int
-fetch_from_buf_socks(buf_t *buf, socks_request_t *req,
-                     int log_sockstype, int safe_socks)
+static int
```
nmathewson: Needs documentation.

src/or/proto_socks.c
```diff
@@ -540,6 +544,15 @@ parse_socks(const char *data, size_t datalen, socks_request_t *req,
     return 0;
   }
 
+  socksver = *data;
```
nmathewson: In this case, we _should_ need to cast to `uint8_t *` first.  Or better, we could use the get_uint8() function.

src/or/proto_socks.c
```diff
@@ -338,6 +338,101 @@ process_socks5_methods_request(socks_request_t *req, int have_user_pass,
...
+    req->got_auth = 1;
+  }
+
+  end:
+  socks5_client_userpass_auth_free(trunnel_req);
+  return res;
+}
+
+static int
+process_socks5_userpass_auth(socks_request_t *req)
```
nmathewson: Needs documentation

src/or/proto_socks.c
```diff
@@ -338,6 +338,101 @@ process_socks5_methods_request(socks_request_t *req, int have_user_pass,
...
+    log_warn(LD_APP, "socks5: parsing failed - invalid user/pass "
+                     "authentication message.");
+    res = -1;
+    goto end;
+  } else if (parsed == -2) {
+    res = 0;
+    goto end;
+  }
+
+  *drain_out = (size_t)parsed;
```
nmathewson: Let's add an assertion, and let's initialize drain_out to 0.

src/or/proto_socks.c
```diff
@@ -338,6 +338,101 @@ process_socks5_methods_request(socks_request_t *req, int have_user_pass,
...
+   socks5_client_userpass_auth_get_username_len(trunnel_req);
+  uint8_t passwordlen =
+   socks5_client_userpass_auth_get_passwd_len(trunnel_req);
+  const char *username = 
+   socks5_client_userpass_auth_getconstarray_username(trunnel_req);
+  const char *password =
+   socks5_client_userpass_auth_getconstarray_passwd(trunnel_req);
+
+  if (usernamelen && username) {
+    req->username = tor_strdup(username);
```
nmathewson: No, this value is not guaranteed to be nul-terminated.  Let's use tor_memdup_nulterm instead.

src/or/proto_socks.c
```diff
@@ -338,6 +338,101 @@ process_socks5_methods_request(socks_request_t *req, int have_user_pass,
...
+    req->usernamelen = (unsigned char)usernamelen;
```
nmathewson: This cast shouldn't be necessary?

src/or/proto_socks.c
```diff
@@ -338,6 +338,101 @@ process_socks5_methods_request(socks_request_t *req, int have_user_pass,
...
+
+    req->got_auth = 1;
+  }
+
+  if (passwordlen && password) {
+    req->password = tor_strdup(password);
```
nmathewson: Same comments as above for this block: memdup_nulterm, no cast?

src/or/proto_socks.c
```diff
@@ -338,6 +338,101 @@ process_socks5_methods_request(socks_request_t *req, int have_user_pass,
...
+    goto end;
+  }
+
+  if (req->auth_type != SOCKS_USER_PASS &&
+      req->auth_type != SOCKS_NO_AUTH) {
+    res = -1;
+    goto end;
+  }
+
+  socks5_server_userpass_auth_set_version(trunnel_resp, 1);
```
nmathewson: These should be constants?

src/or/proto_socks.c
```diff
@@ -378,12 +476,26 @@ handle_socks_message(const uint8_t *raw_data, size_t datalen, socks_request_t *r
       res = 0;
       goto end;
     }
+    /* RFC1929 SOCKS5 username/password subnegotiation. */
+    if ((!req->got_auth && raw_data[0] == 1) ||
```
nmathewson: Should be a constant?

src/or/proto_socks.c
```diff
@@ -433,11 +433,134 @@ process_socks5_userpass_auth(socks_request_t *req)
   return res;
 }
 
+static int
```
nmathewson: Needs documentation.

src/or/proto_socks.c
```diff
@@ -433,11 +433,134 @@ process_socks5_userpass_auth(socks_request_t *req)
...
+  if (parsed == -1) {
+    log_warn(LD_APP, "socks5: parsing failed - invalid client request");
+    res = -1;
+    goto end;
+  } else if (parsed == -2) {
+    res = 0;
+    goto end;
+  }
+
+  *drain_out = (size_t)parsed;
```
nmathewson: Add assertion; initialize to 0.

src/or/proto_socks.c
```diff
@@ -433,11 +433,134 @@ process_socks5_userpass_auth(socks_request_t *req)
...
+    } break;
+  }
+
+  end:
+  socks5_client_request_free(trunnel_req);
+  return res;
+}
+
+static int
+process_socks5_client_request(socks_request_t *req,
```
nmathewson: Can this argument become const?

Needs docs.

src/or/proto_socks.c
```diff
@@ -477,8 +600,8 @@ handle_socks_message(const uint8_t *raw_data, size_t datalen, socks_request_t *r
       goto end;
     }
     /* RFC1929 SOCKS5 username/password subnegotiation. */
-    if ((!req->got_auth && raw_data[0] == 1) ||
-        req->auth_type == SOCKS_USER_PASS) {
+    if (!req->got_auth && (raw_data[0] == 1 ||
```
nmathewson: I don't understand the reasoning behind this change?

src/or/proto_socks.c
```diff
@@ -719,13 +716,26 @@ fetch_from_buf_socks(buf_t *buf, socks_request_t *req,
...
       buf_drain(buf, n_drain);
+
+    switch (socks_res) {
+      case SOCKS_RESULT_INVALID:
+        res = -1;
+        break;
+      case SOCKS_RESULT_DONE:
+        res = 1;
+        break;
+      case SOCKS_RESULT_TRUNCATED:
```
nmathewson: This should have a "falls through" comment so GCC doesn't complain.

src/or/proto_socks.c
```diff
@@ -85,6 +86,138 @@ socks_request_free_(socks_request_t *req)
...
+                         MAX_SOCKS_ADDR_LEN, 0)) {
+      res = -1;
+      goto end;
+    }
+  }
+
+  end:
+  socks4_client_request_free(trunnel_req);
+
+  return res;
```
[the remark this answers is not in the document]
rl1987: It does it through `drain_out`.

src/or/proto_socks.c
```diff
@@ -85,6 +86,138 @@ socks_request_free_(socks_request_t *req)
...
+  // http://ss5.sourceforge.net/socks4A.protocol.txt
+  int res = 1;
+  tor_addr_t destaddr;
+
+  req->socks_version = 4;
+
+  socks4_client_request_t *trunnel_req;
+
+  ssize_t parsed =
+  socks4_client_request_parse(&trunnel_req, (const uint8_t *)raw_data,
```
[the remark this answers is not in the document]
rl1987: Fixed (45882114e6eecf053e47d51758017f72472fdf9b).

src/or/proto_socks.c
```diff
@@ -85,6 +86,138 @@ socks_request_free_(socks_request_t *req)
...
+
+    req->got_auth = 1;
+    req->username = tor_strdup(username);
+    req->usernamelen = usernamelen;
+  }
+
+  if (*is_socks4a) {
+    // We cannot rely on trunnel here, as we want to detect if
+    // we have abnormally long hostname field.
+    char *hostname = (char *)raw_data + SOCKS4_NETWORK_LEN +
```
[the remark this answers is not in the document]
rl1987: To detect hostnames that are overly long (one of the testcases is test_socks.c), we cannot generally rely on trunnel, as it will report failure on entire message. That's the reason for doing this workaround here.

Added `const` in 5785007ed5e0d950c6e7e614637531452f47a1c3.

src/or/proto_socks.c
```diff
@@ -85,6 +86,138 @@ socks_request_free_(socks_request_t *req)
...
+        strlcpy(req->address, trunnel_hostname, sizeof(req->address));
+    } else {
+      log_warn(LD_APP, "socks4: Destaddr too long. Rejecting.");
+      res = -1;
+      goto end;
+    }
+  } else {
+    tor_addr_from_ipv4h(&destaddr, dest_ip);
+
+    if (!tor_addr_to_str(req->address, (const tor_addr_t *)&destaddr,
```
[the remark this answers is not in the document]
rl1987: Fixed (07e470c1fe3d4880f54b55f91024975ecc4bb69b).

src/or/proto_socks.c
rl1987: It does so before the loop. Not sure if this was the same commit.

src/or/proto_socks.c
rl1987: Fixed (e16a55ac2f689cc8b5e52a988173825bbbfebba4).

src/or/proto_socks.c
rl1987: Fixed (a15af2920987f1cf047cd001b9b9ef6a4da2e43c).

src/or/proto_socks.c
rl1987: Fixed together with above.

src/or/proto_socks.c
rl1987: Fixed in 90f61531dc9880cbbc1ec2c07816090abd7bfaf3.

src/trunnel/socks5.trunnel
```diff
@@ -0,0 +1,115 @@
...
+}
+
+struct socks4_server_reply {
+   u8 version IN [4];
+   u8 status;
+   u16 port;
+   u32 addr;
+}
+
+// And here's the extended stuff from proposal 229
```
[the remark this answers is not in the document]
rl1987: Removed in 055840a09ac520d148a9954f9f43187a9e9b4428.

src/or/proto_socks.c
```diff
@@ -85,6 +86,138 @@ socks_request_free_(socks_request_t *req)
...
+  }
+
+  end:
+  socks4_client_request_free(trunnel_req);
+
+  return res;
+}
+
+static int
+process_socks4_request(socks_request_t *req, int is_socks4a,
```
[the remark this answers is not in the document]
rl1987: Yes. efe4b8c19fb81b9aaeb1d675df37b69faa09b47c

src/or/proto_socks.c
```diff
@@ -114,14 +247,56 @@ int
...
                      int log_sockstype, int safe_socks)
 {
-  int res;
+  int res = 0;
+  size_t datalen = buf_datalen(buf);
+  uint8_t *raw_data;
+  uint8_t socks_version;
+
+  raw_data = tor_malloc(datalen);
+  memset(raw_data, 0, datalen);
```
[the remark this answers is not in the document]
rl1987: This was fixed later in the branch. We always pull-up at most 512 bytes of data, to make sure that at least one SOCKS message is in head (if enough data is available). `memset` is also no longer used here.

src/or/proto_socks.c
```diff
@@ -114,14 +247,56 @@ int
...
+
+  buf_peek(buf, (char *)raw_data, datalen);
+
+  socks_version = (uint8_t)raw_data[0];
```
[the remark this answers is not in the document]
rl1987: Removed cast in 78458d8fae85366499ce6bccd52604c6c05d4526. Checks earlier in codepath do not allow `datalen` to be 0. I might have fixed it in different commit.

src/or/proto_socks.c
```diff
@@ -114,14 +247,56 @@ int
...
+    int process_status = process_socks4_request(req, is_socks4a,
+                                                log_sockstype,
+                                                safe_socks);
+
+    if (process_status != 1) {
+      res = process_status;
+      goto end;
+    }
+
+    buf_clear(buf);
```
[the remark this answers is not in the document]
rl1987: I'm pretty sure I did it properly further in the branch. See commits 7d174330a1f9f9ff67935f748cea3b6a77373a7c and 44d7f634a5415065def93bcc57aaeafeaaf0c855. Testcases that try to feed our SOCKS implementation multiple SOCKS messages back-to-back do pass now.

src/or/proto_socks.c
```diff
@@ -114,14 +247,56 @@ int
...
+  }
+
   ssize_t n_drain;
   size_t want_length = 128;
   const char *head = NULL;
-  size_t datalen = 0;
 
-  if (buf_datalen(buf) < 2) /* version and another byte */
-    return 0;
+  if (buf_datalen(buf) < 2) { /* version and another byte */
```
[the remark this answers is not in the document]
rl1987: Fixed later in the branch.

src/or/proto_socks.c
rl1987: Fixed both in bfa4a7da261b4acb75878522d70c305378c1753e.

src/or/proto_socks.c
rl1987: Added more asserts in 83ba4ab5f9160d3ef77801f7dd280d199e42e7cc, 15d23330ac962f8eb13a5b88a7fb77ac6e6071a2 and 6b915edf00724c77ef1612a85c895e9eb30add01.

src/or/proto_socks.c
rl1987: Assert added in bfa4a7da261b4acb75878522d70c305378c1753e. 

Doesn't code from trunnel already include SOCKS4a hostname length into return value? 
```
1648       /* Parse nulterm socks4a_addr_hostname */
1649       {
1650         uint8_t *eos = (uint8_t*)memchr(ptr, 0, remaining);
1651         size_t memlen;
1652         if (eos == NULL)
1653           goto truncated;
1654         trunnel_assert(eos >= ptr);
1655         trunnel_assert((size_t)(eos - ptr) < SIZE_MAX - 1);
1656         memlen = ((size_t)(eos - ptr)) + 1;
1657         if (!(obj->socks4a_addr_hostname = trunnel_malloc(memlen)))
1658           goto fail;
1659         memcpy(obj->socks4a_addr_hostname, ptr, memlen);
1660         remaining -= memlen; ptr += memlen;
1661       }
1662       break;
1663 
1664     default:
1665       break;
1666   }
1667   trunnel_assert(ptr + remaining == input + len_in);
1668   return len_in - remaining;
```

src/or/proto_socks.c
rl1987: Fixed in bfa4a7da261b4acb75878522d70c305378c1753e.

src/or/proto_socks.c
rl1987: Fixed in d71766f462eb2630087f51a474a5f28a09f0f949.

src/or/proto_socks.c
rl1987: Fixed.

src/or/proto_socks.c
rl1987: But we may call `socks_request_set_socks5_error` on this `req`... Doesn't that violate semantics of `const`?

src/or/proto_socks.c
rl1987: Yes, made changes in 805484e2be5ded2736fd3ace42aa804da4f3e638.

src/or/proto_socks.c
rl1987: 805484e2be5ded2736fd3ace42aa804da4f3e638

src/or/proto_socks.c
rl1987: Fixed.

src/or/proto_socks.c
rl1987: I think I did that in some later commit.

src/or/proto_socks.c
rl1987: I don't remember exactly, but it has something to do with how our implicit SOCKS FSM works. If I change it back, `5_authenticate_with_data` starts failing.

This code will go away when we rework our SOCKS implementation to function like an explicit state machine.

src/or/proto_socks.c
```diff
@@ -85,6 +86,138 @@ socks_request_free_(socks_request_t *req)
...
+  socks4_client_request_parse(&trunnel_req, (const uint8_t *)raw_data,
+                              datalen);
+
+  if (parsed == -1) {
+    log_warn(LD_APP, "socks4: parsing failed - invalid request.");
+    res = -1;
+    goto end;
+  } else if (parsed == -2) {
+    res = 0;
+    if (datalen > 1024) { // XXX
```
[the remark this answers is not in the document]
rl1987: No longer comparing `datalen` to 1024 anywhere in proto_socks.c. Comparing to `MAX_SOCKS_MESSAGE_LEN` (512 bytes) instead.

PR #185, opened 2018-06-26
Bug25552 ope

asn-d6:


src/lib/crypt_ops/crypto_ope.h
```diff
@@ -13,8 +13,16 @@
...
+/** Largest value that can be passed to crypto_ope_encrypt().
+
+ *  The current max value (220k seconds) stands for about 61 hours. The
+ *  rationale here is as follows: The rev counter is the time of seconds since
+ *  the start of an SRV period. SRVs are useful for about 48 hours (that's how
+ *  long they stick around on the consensus). Let's also add 12 hours of drift
+ *  for clock skewed services that might be using an old consensus and we
+ *  arrive to 60 hours.
+ */
+#define OPE_INPUT_MAX 220000
```
nmathewson: I'm not completely certain, but I think the OPE code I wrote prefers a power of two here. Can we increase this to 1<<18 (262144)

src/or/hs_service.c
```diff
@@ -1515,9 +1520,9 @@ build_descriptors_for_new_service(hs_service_t *service, time_t now)
 
   /* Build descriptors. */
   build_service_descriptor(service, now, current_desc_tp,
-                           &service->desc_current);
+                           &service->desc_current, 1);
```
nmathewson: If we're using a bool, let's use "true" and "false", not "1" and "0".

src/or/hs_service.c
```diff
@@ -2514,16 +2519,94 @@ increment_descriptor_revision_counter(hs_descriptor_t *hs_desc)
...
+  /* Compute seconds elapsed since the start of the time period. That's the
+   * number of seconds of how long this blinded key has been active. */
+  time_t seconds_since_start_of_srv = now - srv_start;
+
+  /* Increment by one so that we are definitely sure this is strictly
+   * positive and not zero. */
+  seconds_since_start_of_srv++;
+
+  /* Check for too big inputs. */
+  if (BUG(seconds_since_start_of_srv > OPE_INPUT_MAX)) {
```
nmathewson: I think this should be >= ?  Or we should add a -1 to OPE_INPUT_MAX.

src/or/hs_service.c
```diff
@@ -2514,16 +2519,94 @@ increment_descriptor_revision_counter(hs_descriptor_t *hs_desc)
...
+    crypto_digest_add_bytes(digest, ope_key_prefix, sizeof(ope_key_prefix));
+    crypto_digest_add_bytes(digest, (char*)eph_privkey->seckey,
+                            sizeof(eph_privkey->seckey));
+    crypto_digest_get_digest(digest, (char *)key, sizeof(key));
+    crypto_digest_free(digest);
+  }
+
+  { /* Now encrypt the revision counter! */
+    crypto_ope_t *ope = NULL;
+    ope = crypto_ope_new(key);
```
nmathewson: We should keep this key in memory, associated with the service: it is a little costly to compute, and we shouldn't do it over and over.

src/or/hs_service.c
```diff
@@ -2514,16 +2519,94 @@ increment_descriptor_revision_counter(hs_descriptor_t *hs_desc)
...
+    rev_counter = crypto_ope_encrypt(ope, (int) seconds_since_start_of_srv);
+    crypto_ope_free(ope);
+  }
+
+  /* The OPE module returns UINT64_MAX in case of errors. */
```
nmathewson: We should make this a symbolic constant then, like CRYPTO_OPE_ERROR.

src/lib/crypt_ops/crypto_ope.h
asn-d6: Done in 6cabdd0845.

src/or/hs_service.c
asn-d6: Done in 286c8e2e34.

src/or/hs_service.c
asn-d6: Hm, I might need some help here. In general, I like `MAX` types to mean "This is the max allowed value" and hence I like the `>` check.

That also seems to agree with the `crypto_ope_encrypt()` check on top:
```
  if (plaintext <= 0 || plaintext > OPE_INPUT_MAX)
    return CRYPTO_OPE_ERROR;
```

What about the samples optimization? IIUC the `samples` array will contain up to `SAMPLE_INTERVAL*N_SAMPLES` which should be `OPE_INPUT_MAX`. 

Does this make sense? I might need some help here.

src/or/hs_service.c
asn-d6: Done in d5d91e8fc2. Hope that makes sense.

src/or/hs_service.c
asn-d6: Done in 6866acd4c8.

src/or/hs_service.c
asn-d6: This is not needed anymore, because of the bugfix in f13d629db0 .

src/or/hs_service.c
asn-d6: Done in 18dae0bf5b .

src/or/hs_service.c
asn-d6: Done in 62cda39678 .

src/or/hs_service.c
asn-d6: Done in 087e76a35a .

src/or/hs_service.c
asn-d6: Done in b9ae510aa8.

nmathewson:
This branch, or a later version of it, was squashed and merged.

PR #186, opened 2018-06-27
Net refactor

nmathewson:


src/lib/net/ipv6.h
```diff
@@ -0,0 +1,86 @@
...
+  sa_family_t sin6_family;
+  uint16_t sin6_port;
+  // uint32_t sin6_flowinfo;
+  struct in6_addr sin6_addr;
+  // uint32_t sin6_scope_id;
+};
+#endif /* !defined(HAVE_STRUCT_SOCKADDR_IN6) */
+
+const char *tor_inet_ntop(int af, const void *src, char *dst, size_t len);
+int tor_inet_pton(int af, const char *src, void *dst);
```
dgoulet-tor: Those two aren't IPv6 only which makes me wonder if that is the right place to be? Why not in `address.h`? I'm still seeing IPvX specific functions there?

src/lib/net/nettypes.h
```diff
@@ -0,0 +1,39 @@
+/* Copyright (c) 2003-2004, Roger Dingledine
+ * Copyright (c) 2004-2006, Roger Dingledine, Nick Mathewson.
+ * Copyright (c) 2007-2018, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+#ifndef TOR_NET_TYPES_H
+#define TOR_NET_TYPES_H
```
dgoulet-tor: There is a possible argument here that would argue about this file being called `types.h` since it is now in the "net" module by design.

src/common/sandbox.c
```diff
@@ -1458,183 +1459,6 @@ sandbox_cfg_allow_openat_filename(sandbox_cfg_t **cfg, char *file)
   return 0;
 }
```
dgoulet-tor: So the move below of network sandbox code, is the idea to move the per-module sandbox code *into* the module or a specific sandbox module is coming?

src/lib/net/ipv6.h
nmathewson: Hm.  They're IPv6 in the sense that they were introduced at the same time as IPv6, and they are the only way to handle IPv6 addresses.  But you're right that they can also handle IPv4.  Hm....

src/lib/net/nettypes.h
nmathewson: I'm trying hard to avoid having two headers with the same name (like net/types.h and int/types.h).

src/common/sandbox.c
nmathewson: A sandbox module is coming. :)

PR #193, opened 2018-06-29
Remove util and compat

nmathewson:


.gitignore
```diff
@@ -195,6 +195,8 @@ uptime-*.json
 /src/lib/libtor-meminfo-testing.a
 /src/lib/libtor-net.a
 /src/lib/libtor-net-testing.a
+/src/lib/libtor-osinfo.a
+/src/lib/libtor-osinfo-testing.a
```
dgoulet-tor: So here is an idea. How about we have an "os" library for which we could put all the things we use to interact with the operating system.

I'm thinking about this `get_uname()` but also, set file desc, get available memory, etc... ?

src/common/include.am
```diff
@@ -54,6 +54,7 @@ COMMONHEADERS = \
   src/common/compat_libevent.h			\
   src/common/handles.h				\
   src/common/procmon.h				\
+  src/common/socks5_status.h                    \
```
dgoulet-tor: To respond to the commit message, my guts feeling tells me that we probably want a `socks5` library at some point for which it should be (I think!) only visible to the "tor client" component.

.gitignore
nmathewson: Hm. I'm not so sure there.  Isn't *everything* that ends in a syscall "something that interacts with the operating system?"

src/common/include.am
nmathewson: Yeah, I agree

PR #194, opened 2018-06-30
Ticket3723 02

juga0:
Follows #126 after rebase to master.
@teor2345 I think i addressed all your comments.
In the end i thought it was better to add all the tests to the same function to don't duplicate all the variables.
Instead of truncating the string with the bandwidth file headers, i thought it was better to truncate the smartlist. I created a function for that but something is wrong since there's memory leaks when i use it in the test.  See the last 3 commits and the FIXMEs in the code. Not ready to merge because of that.


src/or/dirserv.h
```diff
@@ -32,6 +32,13 @@
 /** Maximum allowable length of a version line in a networkstatus. */
 #define MAX_V_LINE_LEN 128
 
+/** Maximum allowable length of bandwidth headers in a bandwidth file */
+#define MAX_BW_FILE_HEADERS_LEN 50
+
+/** Terminatore that separates bandwidth file headers from bandwidth file
+ * relay lines */
+#define BW_FILE_TERMINATOR "=====\n"
```
teor2345: Please rename to BW_FILE_HEADERS_TERMINATOR to avoid confusion.

src/or/dirserv.c
```diff
@@ -2673,15 +2676,22 @@ dirserv_read_measured_bandwidths(const char *from_file,
...
+      } else if (strcmp(line, BW_FILE_TERMINATOR) == 0)
+        line_is_after_headers = 1;
+      /* if the line was not a correct relay line nor the terminator and
+       * the end of the header lines has not been detected yet
+       * and it is key_value and bw_file_headers did not reach the maximum
+       * number of headers,
+       * then assume this line is a header and add it to bw_file_headers */
+      else if (bw_file_headers &&
+              (line_is_after_headers == 0) &&
+              string_is_key_value(LOG_DEBUG, line) &&
```
teor2345: string_is_key_value returns true for relay bandwidth lines, because it allows spaces:
https://github.com/torproject/tor/blob/be40ad51b68fa87bafb88e55db7856d373274a4d/src/lib/encoding/keyval.c#L19

Please also check that there are no spaces in the line.

src/lib/container/smartlist.c
```diff
@@ -79,6 +79,25 @@ smartlist_string_remove(smartlist_t *sl, const char *element)
...
+{
+  int i, diff;
+  /* Difference between the number of elements in sl and the maximum size */
+  diff = sl->num_used - size;
+  /* Remove elements at position size, difference times, the next element
+   * index will be always size. */
+  for (i=0; i < diff; i++) {
+    log_debug(LD_DIRSERV, "sl->list[size] %s\n", (char *)sl->list[size]);
+    log_debug(LD_DIRSERV, "sl len %d\n", smartlist_len(sl));
+    smartlist_del(sl, size);
```
teor2345: Use smartlist_del_keeporder() to avoid reordering the bandwidth headers.
Use smartlist_del_keeporder(sl, i) to delete the correct item.
Use tor_free(smartlist_get(sl, i)) to free the item before deleting it.

Also, you should probably delete from the end of the smartlist, rather than the start.

src/or/dirauth/dirvote.c
```diff
@@ -267,10 +267,13 @@ format_networkstatus_vote(crypto_pk_t *private_signing_key,
...
+    if (v3_ns->bw_file_headers) {
       bw_file_headers = smartlist_join_strings(v3_ns->bw_file_headers, " ", 0,
-                                              NULL);
-    else
+                                               NULL);
+      if (smartlist_len(v3_ns->bw_file_headers) > MAX_BW_FILE_HEADERS_LEN)
+        /* FIXME */
+        smartlist_truncate(v3_ns->bw_file_headers, MAX_BW_FILE_HEADERS_LEN);
+    } else
       bw_file_headers = tor_strdup("");
```
teor2345: If any block in a group of if/else blocks has braces, please put braces around every block.

src/or/dirauth/dirvote.c
```diff
@@ -267,10 +267,13 @@ format_networkstatus_vote(crypto_pk_t *private_signing_key,
...
     tor_assert(cert);
-    if (v3_ns->bw_file_headers)
+    if (v3_ns->bw_file_headers) {
       bw_file_headers = smartlist_join_strings(v3_ns->bw_file_headers, " ", 0,
-                                              NULL);
-    else
+                                               NULL);
+      if (smartlist_len(v3_ns->bw_file_headers) > MAX_BW_FILE_HEADERS_LEN)
+        /* FIXME */
+        smartlist_truncate(v3_ns->bw_file_headers, MAX_BW_FILE_HEADERS_LEN);
```
teor2345: We already limit the number of headers in earlier code.

So you can just log a BUG() warning here, and vote an empty string:
```
if (BUG(smartlist_len(v3_ns->bw_file_headers) > MAX_BW_FILE_HEADERS_LEN)) {
    bw_file_headers = tor_strdup("");
}
```

You should do this check before joining the string together, to avoid a memory leak, and to avoid allocating large amounts of memory.

src/lib/container/smartlist.c
teor2345: Actually, I think we can delete this entire commit.
See my comments on the next commit.

src/or/dirauth/dirvote.c
```diff
@@ -267,10 +267,13 @@ format_networkstatus_vote(crypto_pk_t *private_signing_key,
     else
       params = tor_strdup("");
     tor_assert(cert);
-    if (v3_ns->bw_file_headers)
+    if (v3_ns->bw_file_headers) {
       bw_file_headers = smartlist_join_strings(v3_ns->bw_file_headers, " ", 0,
```
teor2345: Please limit the length of the bw_file_headers string, not just the number of headers:
https://github.com/torproject/tor/pull/126#discussion_r198677005

(Limiting the number of headers to 50 header is not enough to limit the vote line to 1024 bytes, because each header can be 512 bytes long.)

teor2345: You could log a BUG() warning and use the empty string if the string is too long.
(It's very unlikely that the string will be too long.)
Remember to free the string before replacing it with "".

teor2345:
> Right now there is also this problem with last fixup:

key_value.h isn't included in util.h:
https://github.com/torproject/tor/blob/master/src/common/util.h#L19

Please make a separate commit with this fix.

src/or/dirauth/dirvote.c
juga0: > (Limiting the number of headers to 50 header is not enough to limit the vote line to 1024 bytes, because each header can be 512 bytes long.)

Is there already a constant limiting the size of the vote line?, i don't find something like that, maybe i should add it?.

src/or/dirauth/dirvote.c
juga0: so then i would do 2 checks?, 1 with the smartlist and 1 with the string?.
The code then would be:
```
      if (BUG(smartlist_len(v3_ns->bw_file_headers) > MAX_BW_FILE_HEADERS_LEN)) {
        bw_file_headers = tor_strdup("");
      }
      bw_file_headers = smartlist_join_strings(v3_ns->bw_file_headers, " ", 0,
                                               NULL);
      if (BUG(strlen(bw_file_headers) > MAX_VOTE_LINE_LEN)) {
        tor_free(bw_file_headers);
        bw_file_headers = tor_strdup("");
      }
```

src/or/dirauth/dirvote.c
juga0: It would be:
```
/** Maximum size of a line in a vote. */
MAX_VOTE_LINE_LEN 1024
```
In ``dirvote.h``

teor2345: There are similar line length constants in diserv.h:
https://github.com/torproject/tor/blob/master/src/or/dirserv.h#L29
So I think this new constant should go in dirserv.h, even though it's only used in votes.

You said:
> /** Maximum size of a line in a vote. */
> MAX_VOTE_LINE_LEN 1024
But this constant does not limit all lines in the vote, just the bandwidth-file-headers line.

Please name the line limit constant after the kind of line, something like:
MAX_BANDWIDTH_HEADERS_LINE_LEN

teor2345: Or MAX_BW_FILE_HEADERS_LINE_LEN to match MAX_BW_FILE_HEADERS_LEN.

teor2345: Now that we have two similar constants, we should change MAX_BW_FILE_HEADERS_LEN to MAX_BW_FILE_HEADER_COUNT_IN_VOTE. Otherwise people might get confused.

src/or/dirauth/dirvote.c
teor2345: Almost, but you can't overwrite bw_file_headers with valid data after an error:
```
      if (BUG(smartlist_len(v3_ns->bw_file_headers) > MAX_BW_FILE_HEADERS_LEN)) {
        bw_file_headers = tor_strdup("");
      } else {
        bw_file_headers = smartlist_join_strings(v3_ns->bw_file_headers, " ", 0,
                                                 NULL);
        if (BUG(strlen(bw_file_headers) > MAX_VOTE_LINE_LEN)) {
          tor_free(bw_file_headers);
          bw_file_headers = tor_strdup("");
        }
    }
```

Here's another alternative that might make the code simpler:
```
      char *bw_file_headers = NULL;
      …
      if (! BUG(smartlist_len(v3_ns->bw_file_headers) > MAX_BW_FILE_HEADERS_LEN)) {
        bw_file_headers = smartlist_join_strings(v3_ns->bw_file_headers, " ", 0,
                                                 NULL);
        if (BUG(strlen(bw_file_headers) > MAX_VOTE_LINE_LEN)) {
          /* Free and set to NULL, so the vote header line is empty */
          tor_free(bw_file_headers);
        }
      }
      …
      bw_file_headers ? bw_file_headers : "");
```

Then the final tor_free(bw_file_headers) works because free(NULL) does nothing.

juga0:
Continuing in #195

PR #202, opened 2018-07-03
#21349: Refactor long functions in entrynodes.c (3)

rl1987:
https://trac.torproject.org/projects/tor/ticket/21349

### DIFF
--- a/src/lib/container/smartlist.c
+++ b/src/lib/container/smartlist.c
@@ -188,6 +188,31 @@ smartlist_ints_eq(const smartlist_t *sl1, const smartlist_t *sl2)
   return 1;
 }
 
+/**
+ * Return true if there is shallow equality between smartlists -
+ * i.e. all indices correspond to exactly same object (pointer
+ * values are matching). Otherwise, return false.
+ */
+int
+smartlist_shallow_eq(const smartlist_t *s1, const smartlist_t *s2)
+{
+  if (s1 == NULL || s2 == NULL)
+    return 0;
+
+  if (s1 == s2)
+    return 1;
+
+  if (smartlist_len(s1) != smartlist_len(s2))
+    return 0;
+
+  for (int i = 0; i < smartlist_len(s1); i++) {
+    if (smartlist_get(s1, i) != smartlist_get(s2, i))
+      return 0;
+  }
+
+  return 1;
+}
+
 /** Return true iff <b>sl</b> has some element E such that
  * tor_memeq(E,<b>element</b>,DIGEST_LEN)
  */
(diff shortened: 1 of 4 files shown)


src/lib/container/smartlist.c
```diff
@@ -188,6 +188,31 @@ smartlist_ints_eq(const smartlist_t *sl1, const smartlist_t *sl2)
   return 1;
 }
 
+/**
+ * Return true if there is shallow equality between smartlists -
+ * i.e. all indices correspond to exactly same object (pointer
+ * values are matching). Otherwise, return false.
+ */
+int
+smartlist_shallow_eq(const smartlist_t *s1, const smartlist_t *s2)
```
nmathewson: For consistency with smartlist_ints_eq and smartlist_strings_eq, this should probably be called smartlist_ptrs_eq

src/lib/container/smartlist.c
```diff
@@ -188,6 +188,31 @@ smartlist_ints_eq(const smartlist_t *sl1, const smartlist_t *sl2)
...
+{
+  if (s1 == NULL || s2 == NULL)
```
nmathewson: if `s1 == NULL && s2 == NULL` I think this function should return 1.

src/or/entrynodes.c
```diff
@@ -404,6 +404,17 @@ get_remove_unlisted_guards_after_days(void)
                                  DFLT_REMOVE_UNLISTED_GUARDS_AFTER_DAYS,
                                  1, 365*10);
 }
+
+/**
+ * If a guard has been unlisted for this many seconds continuously, we
```
nmathewson: Usually we'll document a function by using a verb to describe what it does, like "Return the number of seconds after which we remove a guard that..."

src/or/entrynodes.c
```diff
@@ -1235,42 +1246,40 @@ entry_guard_is_listed,(guard_selection_t *gs, const entry_guard_t *guard))
...
+  const time_t unlisted_since_slop =
+    get_remove_unlisted_guards_after_seconds() /  5;
 
-  /* First: Update listed/unlisted. */
   SMARTLIST_FOREACH_BEGIN(gs->sampled_entry_guards, entry_guard_t *, guard) {
     /* XXXX #20827 check ed ID too */
     const int is_listed = entry_guard_is_listed(gs, guard);
 
     if (is_listed && ! guard->currently_listed) {
-      ++n_changes;
```
nmathewson: What's the point of these changes?

src/lib/container/smartlist.c
rl1987: Fixed in ac40d0e22ed097fb5cca5eb31b7d308ffd545947.

src/or/entrynodes.c
rl1987: Reworded in be9401d703eadc675cf9782efd9a7750b4adfb2e.

src/or/entrynodes.c
rl1987: None really, just a byproduct of some now-squashed fixup commits. Changed it back in 4a82c906e36552b98e5daa6874aeb0ef63942650 .

tlyu:
Superseded by #246.

PR #214, opened 2018-07-07
#26502: Stop using fallback blacklist in "updateFallbackDirs.py"

imnotbad:
https://trac.torproject.org/projects/tor/ticket/26502


imnotbad:
This PR wasn't based on master and included other changes, oops!

PR #215, opened 2018-07-07
#26502: Stop using fallback blacklist

imnotbad:
https://trac.torproject.org/projects/tor/ticket/26502


teor2345:
Thanks for this patch, it looks great!

We don't have CI on the fallback list, so I'm going to ask Phoul to run your branch manually before we merge.

imnotbad:
Sounds good, thanks.

PR #219, opened 2018-07-10
Ticket26223

nmathewson:


configure.ac
```diff
@@ -585,7 +585,9 @@ AC_CHECK_FUNCS(
         ftime \
         get_current_dir_name \
         getaddrinfo \
+	getdelim \
         getifaddrs \
+	getline \
```
teor2345: Why are these lines misaligned?
Is it a tabs and spaces thing?

src/feature/dircache/dirserv.c
```diff
@@ -2679,11 +2676,17 @@ dirserv_read_measured_bandwidths(const char *from_file,
...
-  fclose(fp);
   log_info(LD_DIRSERV,
            "Bandwidth measurement file successfully read. "
            "Applied %d measurements.", applied_lines);
-  return 0;
+  rv = 0;
+
+ err:
+  if (line)
+    raw_free(line);
```
teor2345: Let's add a comment saying that we use raw_free() because getdelim() uses raw_malloc().

src/lib/fs/files.h
```diff
@@ -103,4 +103,38 @@ char *read_file_to_str_until_eof(int fd, size_t max_bytes_to_read,
                                  size_t *sz_out)
   ATTR_MALLOC;
 
+#if !defined(HAVE_GETDELIM) || defined(TOR_UNIT_TESTS)
+ssize_t compat_getdelim_(char **lineptr, size_t *n, int delim, FILE *stream);
```
teor2345: There's no function comment for getdelim() - do we expect developers to look it up in the relevant man page?

configure.ac
nmathewson: Yup; added a commit to retabify configure.ac's lists.

src/feature/dircache/dirserv.c
nmathewson: ok

src/lib/fs/files.h
nmathewson: I think that's best -- since this is a POSIX standard, if our function differs from the manpage, one of them is wrong.

PR #223, opened 2018-07-11
Tor Trac #18642: Teach the OOM handler about the DNS cache

neelchauhan:
For the bug [here](https://trac.torproject.org/projects/tor/ticket/18642).

### DIFF
--- a/src/feature/relay/dns.c
+++ b/src/feature/relay/dns.c
@@ -2076,14 +2076,22 @@ dns_cache_entry_count(void)
    return HT_SIZE(&cache_root);
 }
 
+/* Return the total size in bytes of the DNS cache. */
+size_t
+dns_cache_total_allocation(void)
+{
+  return sizeof(struct cached_resolve_t) * dns_cache_entry_count() +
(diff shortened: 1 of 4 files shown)


src/feature/relay/dns.c
```diff
@@ -2101,6 +2101,35 @@ dump_dns_mem_usage(int severity)
...
+    if (!dns_cache_entry_count())
+      break;
+
+    /* Get cutoff interval and remove entries. */
+    time_t cutoff = now + time_inc;
+    purge_expired_resolves(cutoff);
+
+    /* Update amount of bytes removed and array size. */
+    bytes_removed += current_size - dns_cache_total_allocation();
+    current_size -= bytes_removed;
```
dgoulet-tor: I think there is a problem here. Since `bytes_removed` increment every loop, is it possible you'll get offset as you go like this (considering initial state to be `bytes_removed = 0` and `current_size = 100` and every loop, let say we remove 1 entry of size `10`:

```
bytes_removed - current_size
0 - 100
10 - 90
20 - 70
30 - 40
40 - 0
```

PR #224, opened 2018-07-11
Remove the openbsd-malloc code

nmathewson:
Patch from icanhasaccount; closes ticket 20424.


nmathewson:
This code has been removed; see #20424

PR #225, opened 2018-07-11
Add a --with-malloc option.

nmathewson:
Based on a patch from Hello71 on ticket 20424.

This patch additionally fixes openbsd-malloc support, switches
our tcmalloc support to use pkgconfig, and tells the compiler to
omit system malloc implementations as appropriate.


nmathewson:
Squashed and merged as part of #20424

PR #228, opened 2018-07-11
Warn the directory authority operator if their versions list is bogus

nmathewson:
Prevents bug 26485; bugfix on 0.1.1.6-alpha.

### DIFF
--- a/src/or/dirserv.c
+++ b/src/or/dirserv.c
@@ -1043,7 +1043,7 @@ format_recommended_version_list(const config_line_t *ln, int warn)
   }
 
   /* Handle the case where a dirauth operator has accidentally made some
-   * versions comma-separated instead of space-separated. */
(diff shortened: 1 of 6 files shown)


src/or/dirserv.c
```diff
@@ -1042,6 +1041,37 @@ format_versions_list(config_line_t *ln)
     smartlist_split_string(versions, ln->value, ",",
                            SPLIT_SKIP_SPACE|SPLIT_IGNORE_BLANK, 0);
   }
+
+  /* Handle the case where a dirauth operator has accidentally made some
+   * versions comma-separated instead of space-separated. */
```
tlyu: This comment seems backward. The code looks like it checks for and warns about spaces.

tlyu:
This was squashed and merged to 0.2.9 in a321d72401.

PR #234, opened 2018-07-16
Nss symmetric crypto support

nmathewson:


nmathewson:
This was squashed and merged as part of a parent ticket.

PR #239, opened 2018-07-17
Tor Trac #18642: Teach the OOM handler about the DNS cache (Revision 2)

neelchauhan:
For the bug report [here](https://trac.torproject.org/projects/tor/ticket/18642).

### DIFF
--- a/src/core/or/relay.c
+++ b/src/core/or/relay.c
@@ -64,6 +64,7 @@
 #include "lib/crypt_ops/crypto_rand.h"
 #include "lib/crypt_ops/crypto_util.h"
 #include "feature/dircache/directory.h"
+#include "feature/relay/dns.h"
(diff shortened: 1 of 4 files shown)


src/core/or/relay.c
```diff
@@ -2537,6 +2538,7 @@ cell_queues_check_size(void)
   const size_t geoip_client_cache_total =
     geoip_client_cache_total_allocation();
   alloc += geoip_client_cache_total;
+  const size_t dns_cache_total = dns_cache_total_allocation();
```
nmathewson: At this point, shouldn't we also add dns_cache_total to alloc? 

"alloc" is the total amount of bytes allocated that we might be able to free, so I think we should include the DNS bytes here, especially since we subtract them from "alloc" below.

PR #243, opened 2018-07-19
Tor Trac #25152: Try to call less circuitmux_find_map_entry()

neelchauhan:
For the bug [here](https://trac.torproject.org/projects/tor/ticket/25152).

### DIFF
--- a/changes/bug25152
+++ b/changes/bug25152
@@ -1,0 +1,8 @@
+  o Major bugfixes (performance):
+    - Avoid unnecessary calls to circuitmux_find_map_entry() by caching
+      chanid_circid_muxinfo_t entries as much as possible. This is done
+      by passing in chanid_circid_muxinfo_t into functions in circuitmux.c
+      such as circuitmux_make_circuit_active() where a hashmap entry stays
+      the same across multiple subsequent function calls. Fixes bug 25152;
+      bugfix on 0.3.5.1-alpha. Patch by Neel Chauhan.
+
--- a/src/core/or/circuitmux.c
+++ b/src/core/or/circuitmux.c
@@ -561,19 +561,8 @@ circuitmux_set_policy(circuitmux_t *cmux,
  */
 
 cell_direction_t
-circuitmux_attached_circuit_direction(circuitmux_t *cmux, circuit_t *circ)
+circuitmux_attached_circuit_direction(chanid_circid_muxinfo_t *hashent)
 {
-  chanid_circid_muxinfo_t *hashent = NULL;
-
-  /* Try to find a map entry */
-  hashent = circuitmux_find_map_entry(cmux, circ);
-
-  /*
-   * This function should only be called on attached circuits; assert that
-   * we had a map entry.
-   */
-  tor_assert(hashent);
-
   /* Return the direction from the map entry */
   return hashent->muxinfo.direction;
 }
--- a/src/core/or/circuitmux.c
+++ b/src/core/or/circuitmux.c
@@ -90,19 +90,6 @@
 typedef struct chanid_circid_muxinfo_map chanid_circid_muxinfo_map_t;
 
 /*
- * Hash table entry (yeah, calling it chanid_circid_muxinfo_s seems to
- * break the hash table code).
- */
-typedef struct chanid_circid_muxinfo_t chanid_circid_muxinfo_t;
-
-/*
- * Anything the mux wants to store per-circuit in the map; right now just
- * a count of queued cells.
- */
-
-typedef struct circuit_muxinfo_s circuit_muxinfo_t;
-
-/*
  * Structures for circuitmux.c
  */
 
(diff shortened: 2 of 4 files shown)


src/core/or/circuitmux.h
```diff
@@ -16,6 +16,18 @@ typedef struct circuitmux_policy_s circuitmux_policy_t;
 typedef struct circuitmux_policy_data_s circuitmux_policy_data_t;
 typedef struct circuitmux_policy_circ_data_s circuitmux_policy_circ_data_t;
 
+/*
```
nmathewson: Moving the type declarations is okay with me...

src/core/or/circuitmux.h
```diff
@@ -80,6 +92,34 @@ struct circuitmux_policy_circ_data_s {
   uint32_t magic;
 };
 
+/*
+ * This struct holds whatever we want to store per attached circuit on a
+ * circuitmux_t; right now, just the count of queued cells and the direction.
+ */
+
+struct circuit_muxinfo_s {
```
nmathewson: ...but moving the struct definitions seems like a step backwards.  Generally we try to make structure definitions more hidden, and not expose them more.

(Then again, since I am reading these commits in order, I might wind up seeing that you've done this for a very good reason later on.  I'll wait and see...)

src/core/or/relay.c
```diff
@@ -2598,8 +2598,14 @@ update_circuit_on_cmux_(circuit_t *circ, cell_direction_t direction,
...
-  /* Cmux sanity check */
-  if (! circuitmux_is_circuit_attached(cmux, circ)) {
+  /* Now load the circuit map */
+  chanid_circid_muxinfo_t *hashent = NULL;
+  hashent = circuitmux_find_map_entry(cmux, circ);
+  tor_assert(hashent);
+
+  /* Cmux sanity check. We check for hashent == NULL so we call
+   * circuitmux_find_map_entry() less. */
+  if (hashent == NULL) {
```
nmathewson: If we are calling tor_assert(hashent) above, it doesn't make sense to test it for non-NULL here.  I think that the tor_assert() above must be wrong, though ... but it makes me worried that this code is not tested enough.

src/core/or/circuitmux.c
```diff
@@ -561,19 +561,8 @@ circuitmux_set_policy(circuitmux_t *cmux,
  */
 
 cell_direction_t
-circuitmux_attached_circuit_direction(circuitmux_t *cmux, circuit_t *circ)
+circuitmux_attached_circuit_direction(chanid_circid_muxinfo_t *hashent)
```
nmathewson: If we're changing the interface of this function we should change its documentation too.

src/core/or/relay.c
```diff
@@ -2610,7 +2610,7 @@ update_circuit_on_cmux_(circuit_t *circ, cell_direction_t direction,
              file, lineno);
     return;
   }
-  tor_assert(circuitmux_attached_circuit_direction(cmux, circ) == direction);
+  tor_assert(circuitmux_attached_circuit_direction(hashent) == direction);
```
nmathewson: One alternative might be just to remove this assertion entirely -- I don't know if that's the right choice, though.

src/core/or/circuitmux.c
```diff
@@ -1011,23 +1011,17 @@ circuitmux_clear_num_cells(circuitmux_t *cmux, circuit_t *circ)
...
 {
-  chanid_circid_muxinfo_t *hashent = NULL;
-
   tor_assert(cmux);
   tor_assert(circ);
 
-  /* Search for this circuit's entry */
-  hashent = circuitmux_find_map_entry(cmux, circ);
-  /* Assert that we found one */
-  tor_assert(hashent);
```
nmathewson: There's no reason to delete this assertion; and in fact we might want to have an assertion that the hashent matches cmux and circ.

src/core/or/circuitmux.c
```diff
@@ -957,18 +959,15 @@ circuitmux_detach_circuit,(circuitmux_t *cmux, circuit_t *circ))
...
   tor_assert(cmux);
   tor_assert(cmux->policy);
   tor_assert(circ);
 
   /* Policy-specific notification */
   if (cmux->policy->notify_circ_active) {
-    /* Okay, we need to check the circuit for policy data now */
-    chanid_circid_muxinfo_t *hashent = circuitmux_find_map_entry(cmux, circ);
-    /* We should have found something */
-    tor_assert(hashent);
```
nmathewson: This assertion could stay -- and as before, it would be good to check this for consistency with cmux and circ.

src/core/or/circuitmux.c
```diff
@@ -981,18 +980,15 @@ circuitmux_make_circuit_active(circuitmux_t *cmux, circuit_t *circ)
...
   tor_assert(cmux);
   tor_assert(cmux->policy);
   tor_assert(circ);
 
   /* Policy-specific notification */
   if (cmux->policy->notify_circ_inactive) {
-    /* Okay, we need to check the circuit for policy data now */
-    chanid_circid_muxinfo_t *hashent = circuitmux_find_map_entry(cmux, circ);
-    /* We should have found something */
-    tor_assert(hashent);
```
nmathewson: This assertion could stay -- and as before, it would be good to check this for consistency with cmux and circ.

changes/bug25152
```diff
@@ -0,0 +1,8 @@
+  o Major bugfixes (performance):
+    - Avoid unnecessary calls to circuitmux_find_map_entry() by caching
+      chanid_circid_muxinfo_t entries as much as possible. This is done
+      by passing in chanid_circid_muxinfo_t into functions in circuitmux.c
+      such as circuitmux_make_circuit_active() where a hashmap entry stays
+      the same across multiple subsequent function calls. Fixes bug 25152;
+      bugfix on 0.3.5.1-alpha. Patch by Neel Chauhan.
```
nmathewson: When we say "Bugfix on X", we mean that we are fixing a bug, and the bug first appeared in version X.  So if this performance issue first appeared in version 0.a.b.c, and we're fixing it in 0.x.y.z, then we should say "Bugfix on 0.a.b.c", not "Bugfix on 0.x.y.z". 

I think we could just as easily call this particular one a feature.

PR #253, opened 2018-07-31
Bug26979: Improve Appveyor CI IRC logging

teor2345:


teor2345:
This pull request will fail because of https://bugs.torproject.org/26986

PR #256, opened 2018-07-31
[Tor Trac #23588]: Write fascist_firewall_choose_address_ls() and use it in hs_get_extend_info_from_lspecs() (Revision 2)

neelchauhan:
For the bug [here](https://trac.torproject.org/projects/tor/ticket/23588).


src/core/or/policies.c
```diff
@@ -992,7 +992,12 @@ fascist_firewall_choose_address_rs(const routerstatus_t *rs,
 }
 
 /** Like fascist_firewall_choose_address_base(), but takes in a smartlist
- * <b>lspecs</b> consisting of one or more link specifiers.
+ * <b>lspecs</b> consisting of one or more link specifiers. We assume
+ * fw_connection is FIREWALL_OR_CONNECTION as link specifiers cannot
+ * contain DirPorts.
+ * We take in direct_conn to determine whether we are connecting directly
+ * or not. This helps us determine whether to allow IPv6 as we can connect
+ * directly with IPv6, but can extend only with IPv4.
```
teor2345: To avoid confusion, please say "connect directly with IPv4 or IPv6".

src/feature/hs/hs_common.c
```diff
@@ -1723,45 +1714,38 @@ hs_get_extend_info_from_lspecs(const smartlist_t *lspecs,
...
-  } else if (direct_conn &&
-             fascist_firewall_allows_address_addr(&addr_v4, port_v4,
-                                                  FIREWALL_OR_CONNECTION,
-                                                  0, 0)) {
-    /* Direct connection and we can reach it in IPv4 so go for it. */
-    goto validate;
-
-    /* We will add support for falling back to a 3-hop path in a later
-     * release. */
   } else {
```
teor2345: This "else" block is redundant and should be deleted.

src/feature/hs/hs_common.c
```diff
@@ -1723,45 +1714,38 @@ hs_get_extend_info_from_lspecs(const smartlist_t *lspecs,
...
+  if (!tor_addr_port_is_valid_ap(&ap, 0) || !have_legacy_id) {
     goto done;
   }
 
-  /* We know we have IPv4, because we just checked. */
+  /* We know we have an IP address, because we just checked. */
   if (!direct_conn) {
     /* All clients can extend to any IPv4 via a 3-hop path. */
     goto validate;
-  } else if (direct_conn &&
```
teor2345: Don't delete this "else if" block, because its conditions are always true at this point.
Instead, turn it into an "else" block.

src/feature/hs/hs_common.c
```diff
@@ -1723,45 +1714,38 @@ hs_get_extend_info_from_lspecs(const smartlist_t *lspecs,
...
-             fascist_firewall_allows_address_addr(&addr_v4, port_v4,
-                                                  FIREWALL_OR_CONNECTION,
-                                                  0, 0)) {
-    /* Direct connection and we can reach it in IPv4 so go for it. */
```
teor2345: This comment needs to say "we can reach it", without saying "IPv4".

src/feature/hs/hs_common.c
```diff
@@ -1723,45 +1714,38 @@ hs_get_extend_info_from_lspecs(const smartlist_t *lspecs,
     }
   } SMARTLIST_FOREACH_END(ls);
 
+  /* Choose a preferred address first, but fall back to an allowed address.
+   * Skip IPv6 if we're not direct as relays only extend via IPv4. */
+  fascist_firewall_choose_address_ls(lspecs, 0, &ap, direct_conn);
+
   /* Legacy ID is mandatory, and we require IPv4. */
-  if (!have_v4 || !have_legacy_id) {
+  if (!tor_addr_port_is_valid_ap(&ap, 0) || !have_legacy_id) {
```
teor2345: The comment in the deleted else block below now belongs here, but it should say:
 /* If we can't reach any IP address, or we are missing a legacy ID, return NULL. */

src/core/or/policies.c
```diff
@@ -990,6 +991,66 @@ fascist_firewall_choose_address_rs(const routerstatus_t *rs,
...
+ * directly with IPv4 or IPv6, but can extend only with IPv4.
+ */
+void
+fascist_firewall_choose_address_ls(const smartlist_t *lspecs,
+                                   int pref_only, tor_addr_port_t* ap,
+                                   int direct_conn)
+{
+  int have_v4 = 0, have_v6 = 0;
+  uint16_t port_v4 = 0, port_v6 = 0;
+  tor_addr_t addr_v4, addr_v6;
```
teor2345: Please initialise addr_v4 and addr_v6 with tor_addr_make_null().

src/feature/hs/hs_common.c
```diff
@@ -1669,39 +1669,32 @@ hs_pick_hsdir(smartlist_t *responsible_dirs, const char *req_key_str)
...
                                int direct_conn)
 {
-  int have_v4 = 0, have_legacy_id = 0, have_ed25519_id = 0;
+  int have_legacy_id = 0, have_ed25519_id = 0;
   char legacy_id[DIGEST_LEN] = {0};
-  uint16_t port_v4 = 0;
-  tor_addr_t addr_v4;
   ed25519_public_key_t ed25519_pk;
   extend_info_t *info = NULL;
+  tor_addr_port_t ap;
```
teor2345: Please initialise ap with tor_addr_make_null() and ap->port = 0.

teor2345:
Please revert the entire commit 01ffb21. You can revert this commit on the command line using:
```
git revert 01ffb21
```

src/core/or/policies.c
```diff
@@ -990,6 +991,60 @@ fascist_firewall_choose_address_rs(const routerstatus_t *rs,
   }
 }
 
+/** Like fascist_firewall_choose_address_base(), but takes in a smartlist
+ * <b>lspecs</b> consisting of one or more link specifiers.
+ */
+void
+fascist_firewall_choose_address_ls(const smartlist_t *lspecs,
+                                   int pref_only, tor_addr_port_t* ap,
+                                   int direct_conn)
```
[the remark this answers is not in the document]
teor2345: Please remove direct_conn, and remove the function comment about direct_conn.

src/core/or/policies.c
```diff
@@ -990,6 +1001,69 @@ fascist_firewall_choose_address_rs(const routerstatus_t *rs,
...
+      if (have_v4) continue;
+      tor_addr_from_ipv4h(&addr_v4,
+                          link_specifier_get_un_ipv4_addr(ls));
+      port_v4 = link_specifier_get_un_ipv4_port(ls);
+      have_v4 = 1;
+      break;
+    case LS_IPV6:
+      /* Skip if we already seen a v6, or deliberately skip it if we're not a
+       * direct connection. */
+      if (have_v6 || !direct_conn) continue;
```
teor2345: Please remove direct_conn.

src/core/or/policies.h
```diff
@@ -91,6 +91,9 @@ int fascist_firewall_allows_dir_server(const dir_server_t *ds,
 void fascist_firewall_choose_address_rs(const routerstatus_t *rs,
                                         firewall_connection_t fw_connection,
                                         int pref_only, tor_addr_port_t* ap);
+void fascist_firewall_choose_address_ls(const smartlist_t *lspecs,
+                                        int pref_only, tor_addr_port_t* ap,
+                                        int direct_conn);
```
teor2345: Please remove direct_conn.

src/feature/hs/hs_common.c
```diff
@@ -1723,45 +1716,41 @@ hs_get_extend_info_from_lspecs(const smartlist_t *lspecs,
     }
   } SMARTLIST_FOREACH_END(ls);
 
-  /* Legacy ID is mandatory, and we require IPv4. */
-  if (!have_v4 || !have_legacy_id) {
+  /* Choose a preferred address first, but fall back to an allowed address.
+   * Skip IPv6 if we're not direct as relays only extend via IPv4. */
+  fascist_firewall_choose_address_ls(lspecs, 0, &ap, direct_conn);
```
teor2345: Don't call fascist_firewall_choose_address_ls() when direct_conn is false.
Instead, use the IPv4 address, or return NULL if there is no IPv4 address.

teor2345: The function will return NULL if you say:
```
goto done;
```

src/feature/hs/hs_common.c
```diff
@@ -1723,45 +1716,41 @@ hs_get_extend_info_from_lspecs(const smartlist_t *lspecs,
...
+
+  /* Legacy ID is mandatory, and we require an IP address. */
+  if (!tor_addr_port_is_valid_ap(&ap, 0) || !have_legacy_id) {
+    /* If we're missing the legacy ID or the IP address, return NULL. */
     goto done;
   }
 
-  /* We know we have IPv4, because we just checked. */
+  /* We know we have an IP address, because we just checked. */
   if (!direct_conn) {
```
teor2345: Please remove this entire if...else block, and the "validate:" label. It is redundant.

But keep the comment about falling back to a 3-hop path.

src/feature/hs/hs_common.c
```diff
@@ -1698,6 +1698,14 @@ hs_get_extend_info_from_lspecs(const smartlist_t *lspecs,
 
   SMARTLIST_FOREACH_BEGIN(lspecs, const link_specifier_t *, ls) {
     switch (link_specifier_get_ls_type(ls)) {
+    case LS_IPV4:
+      /* Skip if we already seen a v4. */
```
teor2345: Please update the comment to say why we skip when direct_conn is true.

teor2345: Please also update the comment to say what we're doing here:

"If direct_conn is false, initialise ap with the first IPv4 address and port in the link specifiers."

src/test/test_policy.c
```diff
@@ -2422,6 +2435,66 @@ test_policies_fascist_firewall_choose_address(void *arg)
...
+  link_specifier_set_ls_type(fake_ls, LS_IPV4);
+  link_specifier_set_un_ipv4_addr(fake_ls,
+                                  tor_addr_to_ipv4h(&ipv4_or_ap.addr));
+  link_specifier_set_un_ipv4_port(fake_ls, ipv4_or_ap.port);
+  smartlist_add(lspecs, fake_ls);
+
+  /* IPv6 link specifier */
+  fake_ls = link_specifier_new();
+  link_specifier_set_ls_type(fake_ls, LS_IPV6);
+  for (size_t i = 0; i < 16; i++) {
```
teor2345: There's some length setting code missing here, and you can do a memcpy() here if you want.

See, for example:
https://github.com/torproject/tor/blob/master/src/feature/hs/hs_circuit.c#L629

src/test/test_policy.c
```diff
@@ -2422,6 +2435,66 @@ test_policies_fascist_firewall_choose_address(void *arg)
...
+  smartlist_t *lspecs = smartlist_new();
+  link_specifier_t *fake_ls;
+
+  /* IPv4 link specifier */
+  fake_ls = link_specifier_new();
+  link_specifier_set_ls_type(fake_ls, LS_IPV4);
+  link_specifier_set_un_ipv4_addr(fake_ls,
+                                  tor_addr_to_ipv4h(&ipv4_or_ap.addr));
+  link_specifier_set_un_ipv4_port(fake_ls, ipv4_or_ap.port);
+  smartlist_add(lspecs, fake_ls);
```
teor2345: There's some length setting code missing here.

See:
https://github.com/torproject/tor/blob/master/src/feature/hs/hs_circuit.c#L601

src/test/test_policy.c
```diff
@@ -2024,6 +2028,30 @@ test_policies_fascist_firewall_allows_address(void *arg)
...
   STMT_END
 
+/* Check that fascist_firewall_choose_address_ls() returns the expected
+ * results. */
+#define CHECK_CHOSEN_ADDR_LS(fake_ls, pref_only, expect_rv, expect_ap) \
+  STMT_BEGIN \
+    tor_addr_port_t chosen_ls_ap; \
+    tor_addr_make_null(&chosen_ls_ap.addr, AF_UNSPEC); \
+    chosen_ls_ap.port = 0; \
+    fascist_firewall_choose_address_ls(fake_ls, pref_only, &chosen_ls_ap); \
```
teor2345: Please test these cases for fascist_firewall_choose_address_ls():
* fake_ls is NULL
  * the function will crash. Fix it so it logs a bug warning, then returns.
* fake_ls does not have any lspecs
  * the function should not modify ap
* if there is no usable IP address
  * the function should not modify ap
* changing pref_only changes the result
  * when pref_only is 1, chosen_ls_ap is an address, but when pref_only is 0, chosen_ls_ap is unchanged

You can test them in this macro, or in the function.

teor2345:
Hi @neelchauhan, I'm not sure what happened with the commit you just pushed.
I'm seeing the same set of commits on GitHub.

neelchauhan:
The reason why you see the same commits with a new push is because I have two computers and I accidentally pushed an older branch on one so I had to re-push the latest copy on another to avoid losing work.

Sorry for this.

src/core/or/policies.c
```diff
@@ -1017,6 +1017,11 @@ fascist_firewall_choose_address_ls(const smartlist_t *lspecs,
 
   tor_assert(ap);
 
+  if (lspecs == NULL || smartlist_len(lspecs) == 0) {
+    log_warn(LD_BUG, "Unknown or missing link specifiers");
```
teor2345: An empty list of link specifiers isn't a bug in this tor: they're a problem with the data that was sent to this tor.
Let's make an empty list a protocol warning.

src/core/or/policies.c
```diff
@@ -1048,6 +1053,12 @@ fascist_firewall_choose_address_ls(const smartlist_t *lspecs,
     }
   } SMARTLIST_FOREACH_END(ls);
 
+  /* If we don't have IPv4 or IPv6 in link specifiers, log a bug and return. */
+  if (!have_v4 && !have_v6) {
+    log_warn(LD_BUG, "None of our link specifiers have IPv4 or IPv6");
```
teor2345: IPv4 is required. So a missing IPv4 in the link specifiers isn't a bug in this tor: it's a problem with the data that was sent to this tor.
Let's make it a protocol warning.

If IPv6 is also missing, we should change the protocol warning message.

src/core/or/policies.c
```diff
@@ -1018,7 +1018,7 @@ fascist_firewall_choose_address_ls(const smartlist_t *lspecs,
   tor_assert(ap);
 
   if (lspecs == NULL || smartlist_len(lspecs) == 0) {
-    log_warn(LD_BUG, "Unknown or missing link specifiers");
+    log_warn(LD_PROTOCOL, "Unknown or missing link specifiers");
```
teor2345: lspecs == NULL is a bug, but smartlist_len(lspecs) == 0 is a protocol warning

src/feature/hs/hs_common.c
```diff
@@ -1728,45 +1743,38 @@ hs_get_extend_info_from_lspecs(const smartlist_t *lspecs,
...
-  /* Legacy ID is mandatory, and we require IPv4. */
-  if (!have_v4 || !have_legacy_id) {
+  /* Choose a preferred address first, but fall back to an allowed address. */
+  if (direct_conn)
+    fascist_firewall_choose_address_ls(lspecs, 0, &ap);
+
+  /* Legacy ID is mandatory, and we require an IP address. */
+  if (!tor_addr_port_is_valid_ap(&ap, 0)) {
+    /* If we're missing the IP address, log a warning and return NULL. */
+    log_info(LD_NET, "Unknown or invalid IP address in link state");
```
teor2345: The most common reason for this log message will be an unreachable IP address:
```suggestion
    log_info(LD_NET, "Unreachable or invalid IP address in link state");
```

teor2345:
OK, I think we're good here, I'll do the rebase and squash.

teor2345:
Squashed in #583.

PR #257, opened 2018-07-31
Add a new function, tor_api_get_provider_version()

nmathewson:
Closes ticket 26947.

### DIFF
--- a/src/feature/api/tor_api.h
+++ b/src/feature/api/tor_api.h
@@ -58,6 +58,17 @@ int tor_main_configuration_set_command_line(tor_main_configuration_t *cfg,
 void tor_main_configuration_free(tor_main_configuration_t *cfg);
 
 /**
+ * Return the name and version of the software implementing the tor_api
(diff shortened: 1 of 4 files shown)


src/feature/api/tor_api.h
```diff
@@ -57,6 +57,17 @@ int tor_main_configuration_set_command_line(tor_main_configuration_t *cfg,
  */
 void tor_main_configuration_free(tor_main_configuration_t *cfg);
 
+/**
+ * Return the name and version of the software implementing the tor_api
+ * functionality.  Current implementors are "tor" and "libtorrunner".
+ *
```
hellais: Would it be possible to give in the docstring an example of how the version string will be formatted and if there is some sort of assurance on the format it will have?

hellais: FWIW when I run this I get back this version string:
```
- running test_tor_api_get_provider_version()
b'tor 0.3.5.0-alpha-dev'
```

I assume the Added comment is taking into account the fact that you will bump the version when this is merged.

PR #259, opened 2018-08-01
Nss rsa

nmathewson:


src/lib/crypt_ops/crypto_rsa_nss.c
```diff
@@ -0,0 +1,699 @@
+/* Copyright (c) 2001, Matej Pfajfar.
+ * Copyright (c) 2001-2004, Roger Dingledine.
+ * Copyright (c) 2004-2006, Roger Dingledine, Nick Mathewson.
+ * Copyright (c) 2007-2018, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
```
asn-d6: Generic comment: I'd like a log message (perhaps in `crypto_nss_late_init()`) or something saying that it's compiled with libNSS and maybe even the version? Or perhaps along with this log msg: `Tor 0.3.5.0-alpha-dev (git-537faed8034ac7f2) running on Linux with Libevent 2.1.8-stable, OpenSSL 1.1.0h, Zlib 1.2.11, Liblzma N/A, and Libz
std N/A.`

src/lib/crypt_ops/crypto_rsa.c
```diff
@@ -388,3 +392,252 @@ crypto_pk_get_common_digests(crypto_pk_t *pk, common_digests_t *digests_out)
...
+  return rv;
+}
+
+/** PEM-encode the private key portion of <b>env</b> and write it to a
+ * newly allocated string.  On success, set *<b>dest</b> to the new
+ * string, *<b>len</b> to the string's length, and return 0.  On
+ * failure, return -1.
+ */
+int
+crypto_pk_write_private_key_to_string(crypto_pk_t *env,
```
asn-d6: This function is very very similar to the above one. Can we deduplicate?

src/lib/crypt_ops/crypto_rsa.c
```diff
@@ -388,3 +392,252 @@ crypto_pk_get_common_digests(crypto_pk_t *pk, common_digests_t *digests_out)
...
+ done:
+  memwipe(buf, 0, buflen);
+  tor_free(buf);
+  return rv;
+}
+
+/** Read a PEM-encoded private key from the <b>len</b>-byte string <b>s</b>
+ * into <b>env</b>.  Return 0 on success, -1 on failure.  If len is -1,
+ * the string is nul-terminated.
+ */
```
asn-d6: This function is very similar to the above. Should we try to dedup?

src/lib/crypt_ops/crypto_rsa.c
```diff
@@ -388,3 +392,252 @@ crypto_pk_get_common_digests(crypto_pk_t *pk, common_digests_t *digests_out)
...
+
+/** Given a string containing the Base64 encoded DER representation of the
+ * private key <b>str</b>, decode and return the result on success, or NULL
+ * on failure.
+ */
+crypto_pk_t *
+crypto_pk_base64_decode_private(const char *str, size_t len)
+{
+  crypto_pk_t *pk = NULL;
+
```
asn-d6: It's not documented in this function what `len` is here. It's the siez of `str` and also the size of `der`? Shouldn't `base64_decode()` expand the data such that it doesn't fit?

src/lib/crypt_ops/crypto_rsa_nss.c
```diff
@@ -0,0 +1,699 @@
...
+
+/** Return true iff <b>key</b> contains the private-key portion of the RSA
+ * key. */
+int
+crypto_pk_key_is_private(const crypto_pk_t *key)
+{
+  return key && key->seckey;
+}
+
+#ifdef ENABLE_OPENSSL
```
asn-d6: Why this stuff is here and not in `crypto_rsa_openssl.c` ?

src/lib/crypt_ops/crypto_rsa_nss.c
```diff
@@ -0,0 +1,699 @@
...
+  if (seckey)
+    SECKEY_DestroyPrivateKey(seckey);
+
+  return result;
+}
+
+/** Return true iff <b>env</b> has a valid key.
+ */
+int
+crypto_pk_check_key(crypto_pk_t *key)
```
asn-d6: This does not really check that env has a _valid_ key, it just checks that some key exists. Also it's weaker than the OpenSSL equivalent of this function. Is that OK? Should we document it being so, since it's used by external functions?

src/lib/crypt_ops/crypto_rsa_nss.c
```diff
@@ -0,0 +1,699 @@
...
+  return key &&
+    key->pubkey &&
+    key->pubkey->keyType == rsaKey &&
+    DER_GetUInteger(&key->pubkey->u.rsa.publicExponent) == TOR_RSA_EXPONENT;
+}
+
+/** Compare two big-endian integers stored in a and b; return a tristate.
+ */
+STATIC int
+secitem_uint_cmp(const SECItem *a, const SECItem *b)
```
asn-d6: Ouch, do we really have to do this ourselves? Could we use something like `SECITEM_CompareItem()` here?

src/lib/crypt_ops/crypto_rsa_nss.c
```diff
@@ -0,0 +1,699 @@
...
+{
+  int result;
+  char a_is_non_null = (a != NULL) && (a->pubkey != NULL);
+  char b_is_non_null = (b != NULL) && (b->pubkey != NULL);
+  char an_argument_is_null = !a_is_non_null | !b_is_non_null;
+
+  result = tor_memcmp(&a_is_non_null, &b_is_non_null, sizeof(a_is_non_null));
+  if (an_argument_is_null)
+    return result;
+
```
asn-d6: The code in this func up to here is identical to `crypto_pk_cmp_keys()` of openssl. Can we unite?

src/lib/crypt_ops/crypto_rsa_nss.c
```diff
@@ -0,0 +1,699 @@
...
+/** Decrypt <b>fromlen</b> bytes from <b>from</b> with the private key
+ * in <b>env</b>, using the padding method <b>padding</b>.  On success,
+ * write the result to <b>to</b>, and return the number of bytes
+ * written.  On failure, return -1.
+ *
+ * <b>tolen</b> is the number of writable bytes in <b>to</b>, and must be
+ * at least the length of the modulus of <b>key</b>.
+ */
+int
+crypto_pk_private_decrypt(crypto_pk_t *key, char *to,
```
asn-d6: Almost same code as above. Can we dedup?

src/lib/crypt_ops/crypto_rsa_nss.c
```diff
@@ -0,0 +1,699 @@
...
+crypto_pk_public_checksig,(const crypto_pk_t *key, char *to,
+                           size_t tolen,
+                           const char *from, size_t fromlen))
+{
+  tor_assert(key);
+  tor_assert(to);
+  tor_assert(from);
+  tor_assert(tolen < INT_MAX);
+  tor_assert(fromlen < INT_MAX);
+  tor_assert(key->pubkey);
```
asn-d6: Should we use check_key() instead of assert?

src/lib/crypt_ops/crypto_rsa_nss.c
```diff
@@ -0,0 +1,699 @@
...
+};
+
+/** ASN.1-encode the public portion of <b>pk</b> into <b>dest</b>.
+ * Return -1 on error, or the number of characters used on success.
+ */
+int
+crypto_pk_asn1_encode(const crypto_pk_t *pk, char *dest, size_t dest_len)
+{
+  tor_assert(pk);
+  if (pk->pubkey == NULL)
```
asn-d6: Should we use check_key?

src/lib/crypt_ops/crypto_rsa_nss.c
nmathewson: It's in this module because it connects NSS to OpenSSL: it uses both kinds of key.

src/lib/crypt_ops/crypto_rsa_nss.c
nmathewson: I think that SECITEM_CompareItem doesn't actually compare the big-endian integers by integer value -- it's closer to a memcmp, IIUC.

src/lib/crypt_ops/crypto_rsa_nss.c
nmathewson: It currently does that:

Aug 29 16:11:53.148 [notice] Tor 0.3.5.0-alpha-dev (git-e4067aeceae767da) running on Linux with Libevent 2.1.8-stable, NSS 3.38, Zlib 1.2.11, Liblzma 5.2.4, and Libzstd 1.3.5.

(That's in 6a88d8f6b413efdac4b0176cfb78431be46ca9e0 in my TLS branch)

src/lib/crypt_ops/crypto_rsa.c
nmathewson: should be fixed in dbdcf7caba409a

src/lib/crypt_ops/crypto_rsa.c
```diff
@@ -388,3 +392,252 @@ crypto_pk_get_common_digests(crypto_pk_t *pk, common_digests_t *digests_out)
...
+/** Read a PEM-encoded public key from the first <b>len</b> characters of
+ * <b>src</b>, and store the result in <b>env</b>.  Return 0 on success, -1 on
+ * failure.
+ */
+int
+crypto_pk_read_public_key_from_string(crypto_pk_t *env,
+                                      const char *src, size_t len)
+{
+  if (len == (size_t)-1)
```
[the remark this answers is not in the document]
nmathewson: should be fixed in dbdcf7caba409a

src/lib/crypt_ops/crypto_rsa.c
```diff
@@ -388,3 +392,252 @@ crypto_pk_get_common_digests(crypto_pk_t *pk, common_digests_t *digests_out)
...
+/** PEM-encode the private key portion of <b>env</b> and write it to a
+ * newly allocated string.  On success, set *<b>dest</b> to the new
+ * string, *<b>len</b> to the string's length, and return 0.  On
+ * failure, return -1.
+ */
+int
+crypto_pk_write_private_key_to_string(crypto_pk_t *env,
+                                          char **dest, size_t *len)
+{
+  size_t buflen = crypto_pk_keysize(env) * 16;
```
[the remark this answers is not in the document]
nmathewson: should be fixed in dbdcf7caba409a

src/lib/crypt_ops/crypto_rsa.c
nmathewson: should be fixed in dbdcf7caba409a

src/lib/crypt_ops/crypto_rsa_nss.c
nmathewson: actually, it looks like openssl's RSA_check() only works on private keys -- and we only use it in the rest of the code on a key we have just generated, which doesn't make a lot of sense.

Maybe we should just remove it?

src/lib/crypt_ops/crypto_rsa_nss.c
nmathewson: I tried removing this duplication, but it actually made stuff longer.

src/lib/crypt_ops/crypto_rsa_nss.c
nmathewson: So, the duplicated code here is just boilerplate assertions; I think we might be better off leaving this one alone.

src/lib/crypt_ops/crypto_rsa_nss.c
nmathewson: see note on check_key above :)

src/lib/crypt_ops/crypto_rsa_nss.c
asn-d6: ✓

src/lib/crypt_ops/crypto_rsa_nss.c
asn-d6: ✓

src/lib/crypt_ops/crypto_rsa_nss.c
asn-d6: ✓

src/lib/crypt_ops/crypto_rsa_nss.c
asn-d6: ✓

src/lib/crypt_ops/crypto_rsa_nss.c
asn-d6: ✓

src/lib/crypt_ops/crypto_rsa_nss.c
asn-d6: ✓

src/lib/crypt_ops/crypto_rsa.c
asn-d6: ✓

src/lib/crypt_ops/crypto_rsa.c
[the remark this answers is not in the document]
asn-d6: ✓

src/lib/crypt_ops/crypto_rsa.c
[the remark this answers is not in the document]
asn-d6: ✓

src/lib/crypt_ops/crypto_rsa.c
asn-d6: ✓

src/lib/crypt_ops/crypto_rsa_nss.c
asn-d6: WRT openssl RSA_check() and check_key() in general, we also seem to be using it in `crypto_pk_public_encrypt()` which I assume checks non-self-generated public (?) keys too. Not sure how that works. Are you suggesting we remove the RSA_check func?

asn-d6: WRT the NSS `crypto_pk_check_key()`, perhaps we should change the func doc to say that instead of checking for 'valid key',  it just checks that it has a key, because it doesn't check for validity.

Also, it might be worth constifying the args of `crypto_pk_check_key` and `crypto_pk_public_exponent_ok()`

asn-d6: ✓

PR #263, opened 2018-08-06
Bug #27049: Consider hs circs before deciding no circuits are open.

mikeperry-tor:


src/or/circuitlist.c
```diff
@@ -625,7 +625,7 @@ circuit_any_opened_circuits(void)
         TO_CIRCUIT(next_circ)->state == CIRCUIT_STATE_OPEN &&
         TO_CIRCUIT(next_circ)->purpose != CIRCUIT_PURPOSE_C_MEASURE_TIMEOUT &&
         next_circ->build_state &&
-        next_circ->build_state->desired_path_len == DEFAULT_ROUTE_LEN) {
+        next_circ->build_state->desired_path_len >= DEFAULT_ROUTE_LEN) {
```
dgoulet-tor: How does that work with single onion service circuits? I'm not entirely sure, because of preemptive circuit building, if a tor client can NOT have at least 3-hop circuits but I'm guessing their could be a window that it can happen?

changes/bug27409
```diff
@@ -0,0 +1,6 @@
+  o Minor bugfixes (onion services):
+    - When circuits timeout, count other other onion service circuits when
```
dgoulet-tor: Double "other" ?

src/or/circuitlist.c
mikeperry-tor: Hrmm yeah. Ok. We should count SoS circuits here, too. Tried to do that in this commit: 8c950e73d302a6f0b5ada126a5aeac55990f0cc8

changes/bug27409
mikeperry-tor: e86b586211f859c1a5f31301b1fa233910925c31

src/or/circuitlist.c
```diff
@@ -608,6 +608,30 @@ circuit_get_global_origin_circuit_list(void)
...
+ * This means its desired_route_len is either above DEFAULT_ROUTE_LEN or
+ * it's a Single Onion Service circuit.
+ */
+static int
+circuit_is_full_path(const origin_circuit_t *circ)
+{
+  if (!circ->build_state)
+    return 0;
+
+  if ((circ->rend_data || circ->hs_ident) &&
```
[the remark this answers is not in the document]
dgoulet-tor: Every HS circuits have the identifier attached to them. I think this is enough to cover the intro/rend circuits because of the path length of 1.

The HSDir circuit with 3 hops will be caught by the next condition below it seems.

PR #271, opened 2018-08-12
[Tor Trac #25885]: count_acceptable_nodes() would be more accurate using node_has_preferred_descriptor() 

neelchauhan:
For the bug [here](https://trac.torproject.org/projects/tor/ticket/25885).

### DIFF
--- a/src/core/or/circuitbuild.c
+++ b/src/core/or/circuitbuild.c
@@ -1658,22 +1658,25 @@ route_len_for_purpose(uint8_t purpose, extend_info_t *exit_ei)
 STATIC int
 new_route_len(uint8_t purpose, extend_info_t *exit_ei, smartlist_t *nodes)
 {
-  int num_acceptable_routers;
   int routelen;
 
   tor_assert(nodes);
 
   routelen = route_len_for_purpose(purpose, exit_ei);
 
-  num_acceptable_routers = count_acceptable_nodes(nodes);
+  int num_acceptable_direct = count_acceptable_nodes(nodes, 1);
+  int num_acceptable_indirect = count_acceptable_nodes(nodes, 0);
 
-  log_debug(LD_CIRC,"Chosen route length %d (%d/%d routers suitable).",
-            routelen, num_acceptable_routers, smartlist_len(nodes));
+  log_debug(LD_CIRC,"Chosen route length %d (%d direct and %d indirect "
+             "routers suitable).", routelen, num_acceptable_direct,
+             num_acceptable_indirect);
 
-  if (num_acceptable_routers < routelen) {
+  if (num_acceptable_direct < 1 || num_acceptable_indirect < routelen - 1) {
     log_info(LD_CIRC,
-             "Not enough acceptable routers (%d/%d). Discarding this circuit.",
-             num_acceptable_routers, routelen);
+             "Not enough acceptable routers (%d/%d direct and %d/%d "
+             "indirect routers suitable). Discarding this circuit.",
+             num_acceptable_direct, routelen,
+             num_acceptable_indirect, routelen);
     return -1;
   }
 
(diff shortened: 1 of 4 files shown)


src/core/or/circuitbuild.c
```diff
@@ -2418,17 +2414,29 @@ count_acceptable_nodes, (smartlist_t *nodes))
...
 
 //    log_debug(LD_CIRC,"I like %d. num_acceptable_routers now %d.",i, num);
 
-  return num;
+  /* If the number of guard nodes (direct) is greater than one, and the
+   * total number of guard and non-guard nodes (indirect) is greater than
+   * 3, return the route length. Else, return 0 to signify that we do not
+   * have a complete circuit.
+   *
+   * Keep in mind that this function can succeed if direct is greater than
```
nmathewson: This seems to be an unwanted property....

src/core/or/circuitbuild.c
```diff
@@ -2418,17 +2414,29 @@ count_acceptable_nodes, (smartlist_t *nodes))
...
+   * 1 but indirect isn't because guard nodes can also be middle or exit
+   * nodes if configured so on the relay side. */
+  return (direct >= 1) && (direct + indirect >= 3) ? direct + indirect : 0;
```
nmathewson: I think this calculation isn't what we actually want.

After all, we only use this function in one place: to tell if we have enough routers to build a path.  Maybe the right solution here is to have this function return the direct and indirect counts separately, and let new_route_len() decide whether it likes it?

src/core/or/circuitbuild.c
```diff
@@ -2418,17 +2414,29 @@ count_acceptable_nodes, (smartlist_t *nodes))
     if (! node->is_valid)
 //      log_debug(LD_CIRC,"Nope, the directory says %d is not valid.",i);
       continue;
-    if (! node_has_any_descriptor(node))
+    /* If we can use the node, check if we can connect directly. */
+    if (node_has_preferred_descriptor(node, 1))
```
nmathewson: Here you treat "indirect" and "direct" as having a logical relation to one another, as if it's okay to count every "direct" node as also being "indirect" later on (when you do direct+indirect).  That isn't exactly the case, though: for the nodes in the "direct" count, we might not be willing to use them indirectly if we don't have the right descriptor for them.

Probably direct and indirect should be computed separately; maybe "direct" should be an argument to this function?

teor2345: It's also worth noting that bridges are used as entry nodes (direct) but not middle or end nodes (indirect).
The one exception is when a public relay is set as a bridge.

teor2345: To make a path of length N, tor needs at least 1 direct entry node (guard, or bridge, or any node if not using entry guards) and at least N-1 indirect non-entry nodes.

We need to replace the simple test in new_route_len() with separate tests for direct and indirect, and update the log message as well.

src/core/or/circuitbuild.c
```diff
@@ -1658,22 +1658,24 @@ route_len_for_purpose(uint8_t purpose, extend_info_t *exit_ei)
...
+  int num_acceptable_indirect = count_acceptable_nodes(nodes, 0);
 
-  log_debug(LD_CIRC,"Chosen route length %d (%d/%d routers suitable).",
-            routelen, num_acceptable_routers, smartlist_len(nodes));
+  log_debug(LD_CIRC,"Chosen route length %d (%d direct and %d indirect "
+             "routers suitable).", routelen, num_acceptable_direct,
+             num_acceptable_indirect);
 
-  if (num_acceptable_routers < routelen) {
+  if (num_acceptable_direct < 1 && num_acceptable_indirect < routelen - 1) {
```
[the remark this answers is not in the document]
teor2345: I agree, it should be ||.
We need to fail if there are no entry nodes, or if there are not enough non-entry nodes to fill the route.

src/core/or/circuitbuild.c
```diff
@@ -1658,22 +1658,24 @@ route_len_for_purpose(uint8_t purpose, extend_info_t *exit_ei)
...
     log_info(LD_CIRC,
-             "Not enough acceptable routers (%d/%d). Discarding this circuit.",
-             num_acceptable_routers, routelen);
+             "Not enough acceptable routers (%d direct and %d indirect "
```
teor2345: Can you make this log entry say %d/%d direct and %d/%d indirect?

PR #278, opened 2018-08-18
#17873: Listener socket rebinding to/from wildcard address when first bind() fails

rl1987:
https://trac.torproject.org/projects/tor/ticket/17873

### DIFF
--- a/src/core/mainloop/connection.c
+++ b/src/core/mainloop/connection.c
@@ -1213,7 +1213,8 @@ connection_listener_new(const struct sockaddr *listensockaddr,
   tor_addr_t addr;
   int exhaustion = 0;
 
-  if (addr_in_use) *addr_in_use = 0;
+  if (addr_in_use)
+    *addr_in_use = 0;
 
   if (listensockaddr->sa_family == AF_INET ||
       listensockaddr->sa_family == AF_INET6) {
@@ -1295,7 +1296,8 @@ connection_listener_new(const struct sockaddr *listensockaddr,
       int e = tor_socket_errno(s);
       if (ERRNO_IS_EADDRINUSE(e)) {
         helpfulhint = ". Is Tor already running?";
-        if (addr_in_use) *addr_in_use = 1;
+        if (addr_in_use)
+          *addr_in_use = 1;
       }
       log_warn(LD_NET, "Could not bind to %s:%u: %s%s", address, usePort,
                tor_socket_strerror(e), helpfulhint);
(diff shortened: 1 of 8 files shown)


src/core/mainloop/connection.c
```diff
@@ -2759,20 +2844,54 @@ retry_all_listeners(smartlist_t *replaced_conns,
...
+
+    connection_t *new_conn =
+     connection_listener_new_for_port(r->new_port, &skip, &addr_in_use);
+    connection_t *old_conn = r->old_conn;
+
+    if (skip)
+      continue;
+
+    connection_close_immediate(old_conn);
+    connection_mark_for_close(old_conn);
```
ahf: Are both needed here?

src/core/mainloop/connection.c
```diff
@@ -2759,20 +2844,54 @@ retry_all_listeners(smartlist_t *replaced_conns,
...
+#ifdef ENABLE_LISTENER_REBIND
+  SMARTLIST_FOREACH_BEGIN(replacements, listener_replacement_t *, r) {
+    int addr_in_use = 0;
+    int skip = 0;
+
+    tor_assert(r->new_port);
+    tor_assert(r->old_conn);
+
+    connection_t *new_conn =
+     connection_listener_new_for_port(r->new_port, &skip, &addr_in_use);
```
ahf: Nitpick: Add an additional space here.

src/core/mainloop/connection.c
```diff
@@ -1473,6 +1506,8 @@ connection_listener_new(const struct sockaddr *listensockaddr,
    */
   connection_check_oos(get_n_open_sockets(), 0);
 
+  log_notice(LD_NET, "Opened %s on %s",
+             conn_type_to_string(type), fmt_addrport(&addr, usePort));
```
ahf: Should this maybe be at info log level instead?

src/core/mainloop/connection.c
rl1987: 1d92370f40475546f7b672a444cfc777b3cb577c

src/core/mainloop/connection.c
rl1987: I believe so - we want to close the old connection right away, without trying to send any data from its write buffer. It is generally done like this in connection.c when closing no longer configured or failed listeners.

See:
* https://gitweb.torproject.org/tor.git/tree/src/core/mainloop/connection.c#n2773
* https://gitweb.torproject.org/tor.git/tree/src/core/mainloop/connection.c#n3426

src/core/mainloop/connection.c
nmathewson: I think we have it at notice now, right?  We can leave it there if so, and move it to info later if we decide.

(If we don't have this at notice now, please ignore me)

PR #284, opened 2018-08-21
Check for duplicate-close in connection_dir_finished_flusing()

nmathewson:
Fix for 26896.


nmathewson:
This was squashed and merged in master

PR #288, opened 2018-08-22
Ticket27215 035 01

dgoulet-tor:


src/feature/hs/hs_service.c
```diff
@@ -3375,6 +3398,36 @@ hs_service_circuit_has_opened(origin_circuit_t *circ)
...
+/* Return the service version by looking at the key in the service directory.
+ * If the key is not found or unrecognized, -1 is returned. Else, the service
+ * version is returned. */
+int
+hs_service_get_version_from_key(const hs_service_t *service)
+{
+  int version = -1; /* Unknown version. */
+  const char *directory_path;
+
+  tor_assert(service);
```
teor2345: We try to avoid fatal asserts.

We could say:
```
if (BUG(!service)) {
  return -1;
}
```

src/feature/hs/hs_service.c
```diff
@@ -3375,6 +3398,36 @@ hs_service_circuit_has_opened(origin_circuit_t *circ)
...
+
+  /* We'll try to load the key for version 3. If not found, we'll try version
+   * 2 and if not found, we'll send back an unknown version (0). */
```
teor2345: unknown version is -1

doc/tor.1.txt
```diff
@@ -2826,7 +2826,7 @@ The following options are used to configure a hidden service.
 
 [[HiddenServiceVersion]] **HiddenServiceVersion** **2**|**3**::
     A list of rendezvous service descriptor versions to publish for the hidden
-    service. Currently, versions 2 and 3 are supported. (Default: 2)
+    service. Currently, versions 2 and 3 are supported. (Default: 3)
```
teor2345: 😎

src/feature/hs/hs_service.c
dgoulet-tor: I'm not entirely keen on using non fatal `BUG()`. This is a function parameter assert() that is I want it to crash if the function is used without a service that is incorrectly. It can't and shouldn't be used with a NULL service and if it is, I want to find it quickly and us to fix the code flow.

Accepting a NULL service there will lead to tor basically segfaulting. So we can't recover from the BUG() in this case. 

Not only that, but returning -1 for this function means "unknown version for given service object" implying that the object exists. It is a "context dependent function" that can't return that same reason for a non existent object.

We do use those parameter asserts() a *lot* in the HS subsystem because there is a point where we need to consider certain function input for granted so we can only error on specific use cases related to what has been given and not returning error code on "if you failed to use the function properly".

PR #292, opened 2018-08-23
router: Keep RSA onion public key in ASN.1 format

dgoulet-tor:
The OpenSSL "RSA" object is currently 408 bytes compares to the ASN.1 encoding
which is 140 for a 1024 RSA key.

We save 268 bytes per descriptor (routerinfo_t) *and* microdescriptor
(microdesc_t). Scaling this to 6000 relays, and considering client usually
only have microdescriptors, we save 1.608 MB of RAM which is considerable for
mobile client.

This commit makes it that we keep the RSA onion public key (used for TAP
handshake) in ASN.1 format instead of an OpenSSL RSA object.

Changes is done in both routerinfo_t and microdesc_t.

Closes #27246

### DIFF
--- a/src/feature/nodelist/microdesc.c
+++ b/src/feature/nodelist/microdesc.c
@@ -874,7 +874,7 @@ microdesc_free_(microdesc_t *md, const char *fname, int lineno)
   //tor_assert(md->held_by_nodes == 0);
 
   if (md->onion_pkey)
-    crypto_pk_free(md->onion_pkey);
+    tor_free(md->onion_pkey);
   tor_free(md->onion_curve25519_pkey);
   tor_free(md->ed25519_identity_pkey);
   if (md->body && md->saved_location != SAVED_IN_CACHE)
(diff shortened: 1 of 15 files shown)


src/feature/nodelist/microdesc.c
```diff
@@ -1057,3 +1057,40 @@ usable_consensus_flavor,(void))
...
   }
 }
+
+/* Return a newly allocated RSA key object from the ASN-1 encoded RSA onion
+ * public key in the microdescriptor. It is the caller responsability to free
+ * the returned object.
+ *
+ * Return NULL if md or the md's onion pkey is NULL or malformed. */
+crypto_pk_t *
+microdesc_get_rsa_onion_pkey(const microdesc_t *md)
```
nmathewson: These two functions seem redundant with their routerinfo_t equivalent.  Do you think it's reasonable to combine them -- especially the "set" one, since it's longer?

src/feature/nodelist/routerlist.c
```diff
@@ -5498,7 +5498,7 @@ router_differences_are_cosmetic(const routerinfo_t *r1, const routerinfo_t *r2)
       r1->ipv6_orport != r2->ipv6_orport ||
       r1->dir_port != r2->dir_port ||
       r1->purpose != r2->purpose ||
-      !crypto_pk_eq_keys(r1->onion_pkey, r2->onion_pkey) ||
+      !tor_memeq(r1->onion_pkey, r2->onion_pkey, r1->onion_pkey_len) ||
```
nmathewson: Bug here: you need to check that the lengths are equal first!

src/feature/nodelist/microdesc.c
```diff
@@ -1057,3 +1057,40 @@ usable_consensus_flavor,(void))
...
+
+  tor_assert(md);
+  tor_assert(pk);
+
+  len = crypto_pk_asn1_encode(pk, buf, sizeof(buf));
+  if (BUG(len < 0)) {
+    goto done;
+  }
+
+  md->onion_pkey = tor_malloc_zero(len);
```
nmathewson: You could use tor_memdup() here.

src/feature/nodelist/microdesc.c
dgoulet-tor: Agrree. I merge them together and put the helper functions in `router.c`. Hope this is fine. See fixup commit: 61bdbe40c7ea5b7d

src/feature/nodelist/routerlist.c
dgoulet-tor: Fixup eb692dcb0d

src/feature/nodelist/microdesc.c
dgoulet-tor: Fixed in the previous commit.

PR #293, opened 2018-08-26
#27224: Conserve memory allocations

rl1987:
https://trac.torproject.org/projects/tor/ticket/27224


nmathewson:
(squashed and merged)

PR #297, opened 2018-08-29
Ticket25573 v2 squashed3

mikeperry-tor:
New squashed branch.


mikeperry-tor:
I want a better branch name for the actual merge.

PR #301, opened 2018-08-30
Ticket20700 035 02

dgoulet-tor:


src/feature/hs/hs_service.c
```diff
@@ -1070,13 +1077,192 @@ load_service_keys(hs_service_t *service)
...
+  pubkey_b32 = smartlist_get(fields, 2);
+
+  /* Currently, the only supported auth type is "descriptor" and the only
+   * supported key type is "x25519". */
+  if (strcmp(auth_type, "descriptor") || strcmp(key_type, "x25519")) {
+    goto err;
+  }
+
+  client = tor_malloc_zero(sizeof(hs_service_authorized_client_t));
+  if (base32_decode((char *) client->client_pk.public_key,
```
nmathewson: Will this detect truncated public keys?

src/feature/hs/hs_service.c
```diff
@@ -1070,13 +1077,192 @@ load_service_keys(hs_service_t *service)
...
+  tor_assert(service);
+
+  config = &service->config;
+
+  /* Before calling this function, we already call load_service_keys to make
+   * sure that the directory exists with the right permission. So, if we
+   * cannot create a client pubkey key directory, we consider it as a bug. */
+  client_keys_dir_path = hs_path_from_filename(config->directory_path,
+                                               dname_client_pubkeys);
+  if (BUG(hs_check_service_private_dir(get_options()->User,
```
nmathewson: Are we calling this from inside the options code, or from somewhere else?  If it's inside the options code, we can't call get_options() -- we should be taking the or_options_t as an argument instead.  But if it's happening from somewhere else, this is okay.

src/feature/hs/hs_service.c
```diff
@@ -1070,13 +1077,192 @@ load_service_keys(hs_service_t *service)
...
+    SMARTLIST_FOREACH(config->clients, hs_service_authorized_client_t *, p,
+                      service_authorized_client_free(p));
+    smartlist_free(config->clients);
+  }
+
+  config->clients = smartlist_new();
+
+  file_list = tor_listdir(client_keys_dir_path);
+
+  SMARTLIST_FOREACH_BEGIN(file_list, char *, filename) {
```
nmathewson: Could this be  const char *?

src/feature/hs/hs_service.c
```diff
@@ -1070,13 +1077,192 @@ load_service_keys(hs_service_t *service)
...
+    hs_service_authorized_client_t *client = NULL;
+
```
nmathewson: I'm a little concerned that this loop provides no user feedback.  If a file is invalid or ignored, how will the user find out?

src/feature/hs/hs_service.c
```diff
@@ -1070,13 +1077,192 @@ load_service_keys(hs_service_t *service)
...
+      tor_free(client_key_file_path);
+
+      /* If we cannot read the file, continue with the next file. */
+      if (!client_key_str)  {
+        continue;
+      }
+
+      client = parse_authorized_client(client_key_str);
+      /* Free immediately after using it. */
+      tor_free(client_key_str);
```
nmathewson: If it's important to wipe client_key_str below, then it's important to wipe it here too.

src/feature/hs/hs_service.c
```diff
@@ -1070,13 +1077,192 @@ load_service_keys(hs_service_t *service)
...
+  /* If the list of clients already exists, we must clear it first. */
+  if (config->clients) {
+    SMARTLIST_FOREACH(config->clients, hs_service_authorized_client_t *, p,
+                      service_authorized_client_free(p));
+    smartlist_free(config->clients);
+  }
+
+  config->clients = smartlist_new();
+
+  file_list = tor_listdir(client_keys_dir_path);
```
nmathewson: We need to check for whether this value is NULL, I think.

src/test/test_hs_service.c
```diff
@@ -322,12 +332,183 @@ test_load_keys(void *arg)
...
+  tt_assert(!parse_authorized_client("a:b:c:d:e"));
+  /* Wrong auth type. */
+  tt_assert(!parse_authorized_client(
+    "x:x25519:dz4q5xqlb4ldnbs72iarrml4ephk3du4i7o2cgiva5lwr6wkquja"));
+  /* Wrong key type. */
+  tt_assert(!parse_authorized_client(
+    "descriptor:x:dz4q5xqlb4ldnbs72iarrml4ephk3du4i7o2cgiva5lwr6wkquja"));
+  /* Some malformed string. */
+  tt_assert(!parse_authorized_client("descriptor:x25519:aa=="));
+
```
nmathewson: I'd suggest adding checks for 1-field, 2-field, and 0-field strings as well: they're a corner case that we're likely to mess up in the future if we don't have tests.

src/test/test_hs_service.c
```diff
@@ -322,12 +332,183 @@ test_load_keys(void *arg)
...
+  tor_asprintf(&conf, conf_fmt, hsdir_v3, HS_VERSION_THREE);
+  ret = helper_config_service(conf);
+  tor_free(conf);
+  tt_int_op(ret, OP_EQ, 0);
+  /* It's in staging? */
+  tt_int_op(get_hs_service_staging_list_size(), OP_EQ, 1);
+
+#undef conf_fmt
+
+  MOCK(read_file_to_str, mock_read_file_to_str);
```
nmathewson: It's okay to do this kind of mocking -- but in the future, it might be just as good (or better) to just put the files into a temporary directory on disk.  See uses of get_fname() for other places in the tests where we generate temporary files and directories.

src/test/test_hs_service.c
```diff
@@ -322,12 +332,183 @@ test_load_keys(void *arg)
...
+
+    curve25519_public_key_t pubkey;
+    /* This flag will be set if the key is found in clients. */
+    int is_found = 0;
+    base32_decode((char *) pubkey.public_key, sizeof(pubkey.public_key),
+                  pubkey_b32, strlen(pubkey_b32));
+
+    SMARTLIST_FOREACH_BEGIN(service->config.clients,
+                            hs_service_authorized_client_t *, client) {
+      if (!tor_memcmp(&pubkey, &client->client_pk, sizeof(pubkey))) {
```
nmathewson: See documentation for tor_memcmp: it's better to use tor_memeq() in this case.

nmathewson: err, ug. This is not actually documented enough.  Will document now.

src/feature/hs/hs_descriptor.c
```diff
@@ -2475,6 +2504,84 @@ hs_desc_intro_point_free_(hs_desc_intro_point_t *ip)
...
+                              HS_DESC_DESCRIPTOR_COOKIE_LEN));
+
+  /* Calculate x25519(hs_y, client_X) */
+  curve25519_handshake(secret_seed,
+                       auth_ephemeral_sk,
+                       client_pk);
+
+  /* Calculate KEYS = KDF(SECRET_SEED, 40) */
+  xof = crypto_xof_new();
+  crypto_xof_add_bytes(xof, secret_seed, sizeof(secret_seed));
```
nmathewson: Now that I'm looking at this implementation, I'm a little concerned that we're only putting the curve25519 output into the KDF.  I think we should also put something about the hidden service itself (its identity? its blinded identity?) into the KDF, to prevent tricky attacks where the ephemeral public key is copied from one descriptor to another.  What do you think?

src/feature/hs/hs_descriptor.c
```diff
@@ -240,53 +240,73 @@ build_mac(const uint8_t *mac_key, size_t mac_key_len,
   crypto_digest_free(digest);
 }
 
-/* Using a given decriptor object, build the secret input needed for the
- * KDF and put it in the dst pointer which is an already allocated buffer
- * of size dstlen. */
-static void
-build_secret_input(const hs_descriptor_t *desc, uint8_t *dst, size_t dstlen)
+/* Using a secret data and a given decriptor object, build the secret
```
nmathewson: FYI, this comes up a lot: When documenting functions in the future, please use "/**", not "/*".  The first format tells doxygen to parse the comment.

But also please don't make this change on this branch -- I bet it would cause a lot of conflicts.

src/feature/hs/hs_descriptor.c
```diff
@@ -240,53 +240,73 @@ build_mac(const uint8_t *mac_key, size_t mac_key_len,
...
+  tor_assert(secret_input_out);
+
+  secret_input = tor_malloc_zero(secret_input_len);
+
+  /* Copy the secret data. */
+  memcpy(secret_input, secret_data, secret_data_len);
+  offset += secret_data_len;
   /* Copy subcredential. */
-  memcpy(dst + offset, desc->subcredential, sizeof(desc->subcredential));
+  memcpy(secret_input + offset, desc->subcredential,
```
nmathewson: I don't like how this is using "sizeof(desc->subcredential)"  but the calculation for secret_input_len is using DIGEST256_LEN.  We should really be using the same expression in both cases, to keep them in sync.

src/feature/hs/hs_client.c
```diff
@@ -1393,6 +1397,212 @@ hs_client_receive_rendezvous_acked(origin_circuit_t *circ,
...
+  }
+  digest256map_free(client_auths, client_service_authorization_free_void);
+}
+
+/* Check if the auth key file name is valid or not. Return 1 if valid,
+ * otherwise return 0. */
+static int
+auth_key_filename_is_valid(const char *filename)
+{
+  int ret = 1;
```
nmathewson: This should use strcmpend()

src/feature/hs/hs_client.c
```diff
@@ -1393,6 +1397,212 @@ hs_client_receive_rendezvous_acked(origin_circuit_t *circ,
...
+    ret = 0;
+  }
+
+  return ret;
+}
+
+static hs_client_service_authorization_t *
+parse_auth_file_content(const char *client_key_str)
+{
+  char *onion_address = NULL;
```
nmathewson: This seems to be mostly duplicated code from the other parse function.  Could they be combined?

src/feature/hs/hs_client.c
```diff
@@ -1393,6 +1397,212 @@ hs_client_receive_rendezvous_acked(origin_circuit_t *circ,
...
+  }
+
+  key_dir = tor_strdup(options->ClientOnionAuthDir);
+
+  /* Make sure the directory exists and is private enough. */
+  if (check_private_dir(key_dir, 0, options->User) < 0) {
+    goto end;
+  }
+
+  file_list = tor_listdir(key_dir);
```
nmathewson: This function can return NULL

src/test/test_hs_descriptor.c
```diff
@@ -850,103 +851,6 @@ test_build_authorized_client(void *arg)
   UNMOCK(crypto_strongest_rand);
 }
 
-/* bad desc auth type */
```
nmathewson: I'm worried about losing these tests. Do we get them back later in the branch?

src/feature/hs/hs_descriptor.c
```diff
@@ -1182,6 +1182,38 @@ desc_encode_v3(const hs_descriptor_t *desc,
...
+static int
+decode_auth_client(const directory_token_t *tok,
+                   hs_desc_authorized_client_t *client)
+{
+  int ret = -1;
+  tor_assert(tok);
+  tor_assert(tok->n_args >= 3);
+  tor_assert(client);
+
+  if (base64_decode((char *) client->client_id, sizeof(client->client_id),
```
nmathewson: It's not enough to check that this didn't return a negative number -- you need to make sure that it returned the right positive number, or that its input was the right length.

**Please audit all of the base*_decode() functions in this branch to make sure that they don't have this bug; I may have missed some.**

src/feature/hs/hs_descriptor.c
```diff
@@ -1421,6 +1453,73 @@ encrypted_data_length_is_valid(size_t len)
...
+  tor_assert(!tor_mem_is_zero((char *) client_sk,
+                              sizeof(*client_sk)));
+
+  /* Calculate x25519(client_x, hs_Y) */
+  curve25519_handshake(secret_seed, client_sk,
+                       &desc->superencrypted_data.auth_ephemeral_pubkey);
+
+  /* Calculate KEYS = KDF(SECRET_SEED, 40) */
+  xof = crypto_xof_new();
+  crypto_xof_add_bytes(xof, secret_seed, sizeof(secret_seed));
```
nmathewson: Is this also duplicate code? I feel like I've read it before in the first comment on the branch.

src/feature/hs/hs_descriptor.c
```diff
@@ -1421,6 +1453,73 @@ encrypted_data_length_is_valid(size_t len)
...
+  crypto_xof_squeeze_bytes(xof, keystream, sizeof(keystream));
+  crypto_xof_free(xof);
+
+  /* If the client id of auth client is not the same as the calculcated
+   * client id, it means that this auth client is invaild according to the
+   * client secret key client_sk. */
+  if (!tor_memeq(client->client_id, keystream, HS_DESC_CLIENT_ID_LEN)) {
```
nmathewson: Instead of !tor_memeq, you can use tor_memneq

src/feature/hs/hs_descriptor.c
```diff
@@ -2126,10 +2241,22 @@ desc_decode_superencrypted_v3(const hs_descriptor_t *desc,
...
   SMARTLIST_FOREACH_BEGIN(tokens, const directory_token_t *, token) {
     if (token->tp == R3_DESC_AUTH_CLIENT) {
       tor_assert(token->n_args >= 3);
-      /* XXX: Extract each auth client. */
+
+      hs_desc_authorized_client_t *client =
+        tor_malloc_zero(sizeof(hs_desc_authorized_client_t));
+
+      if (decode_auth_client(token, client) < 0) {
+        log_warn(LD_REND, "Auth client is not valid");
```
nmathewson: Is this message actually going to tell the user what they need to know? 

 What will happen to the connection?  Will it be retried?

src/feature/hs/hs_service.c
```diff
@@ -1301,8 +1303,87 @@ service_descriptor_new(void)
...
+  }
+
+  /* An array that has the same length as the list of clients in config2.
+   * If the client in config2 is matched with some client in config1, the
+   * value of this array at the corresponding index will be 1. Otherwise,
+   * it will be 0. */
+  idx_matched =
+    tor_malloc_zero(sizeof(*idx_matched) * smartlist_len(config2->clients));
+
+  SMARTLIST_FOREACH_BEGIN(config1->clients,
```
nmathewson: I don't like doing this quadratic loop here -- do we really need to do this?

nmathewson: One option would be to sort the lists, then compare them in lockstep.

Come to think of it, we should probably also  scramble the lists before we actually use them, so a client doesn't learn anything from its position in the list.

src/feature/hs/hs_service.c
```diff
@@ -1136,23 +1136,31 @@ parse_authorized_client(const char *client_key_str)
                          SPLIT_SKIP_SPACE, 0);
   /* Wrong number of fields. */
   if (smartlist_len(fields) != 3) {
+    log_warn(LD_REND, "The file is in a wrong format.");
```
nmathewson: Let's say which file -- it's not going to help the user otherwise.

src/feature/hs/hs_service.c
```diff
@@ -1136,23 +1136,31 @@ parse_authorized_client(const char *client_key_str)
...
+  if (strcmp(key_type, "x25519")) {
+    log_warn(LD_REND, "The key type '%s' is not supported.", key_type);
     goto err;
   }
 
   client = tor_malloc_zero(sizeof(hs_service_authorized_client_t));
   if (base32_decode((char *) client->client_pk.public_key,
                     sizeof(client->client_pk.public_key),
                     pubkey_b32, strlen(pubkey_b32)) < 0) {
+    log_warn(LD_REND, "The public key cannot be decoded.");
```
nmathewson: Again, we should really be saying _what file_ is busted, or the user won't know what to do about it.

src/feature/hs/hs_service.c
```diff
@@ -1070,13 +1077,192 @@ load_service_keys(hs_service_t *service)
...
   tor_free(fname);
   return ret;
 }
 
+/* Check if the client file name is valid or not. Return 1 if valid,
+ * otherwise return 0. */
+static int
+client_filename_is_valid(const char *filename)
+{
+  int ret = 1;
```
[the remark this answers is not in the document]
dgoulet-tor: Fixup commit: 67fa2261f0

src/feature/hs/hs_service.c
```diff
@@ -1070,13 +1077,192 @@ load_service_keys(hs_service_t *service)
...
+
+  return ret;
+}
+
+/* Parse an authorized client from a string. Return the key on success, return
+ * NULL, otherwise. */
+static hs_service_authorized_client_t *
+parse_authorized_client(const char *client_key_str)
+{
+  char *auth_type = NULL;
```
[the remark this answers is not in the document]
dgoulet-tor: Fixup commit: de70396bcd

src/feature/hs/hs_service.c
dgoulet-tor: Fixup commit: 7396175858

src/feature/hs/hs_service.c
dgoulet-tor: Soooo we might have a bigger problem here. It is ultimately called from `options_act()` and for v3 it is easy to fix but we'll need to do it for v2 as well.

`get_options()->User` is what is accessed during key loading since we double check if the directory permissions are valid.

This will touch many parts of HS and Rend subsystem so I propose we do this after with its own set of commits?

src/feature/hs/hs_service.c
dgoulet-tor: Fixup commit: 7eaa650d6d

src/feature/hs/hs_service.c
dgoulet-tor: Fixup commit: 55054869a3

src/test/test_hs_service.c
dgoulet-tor: Fixup commit: 629c1880d54

src/test/test_hs_service.c
dgoulet-tor: Fixup commit: f8e98f82a6

src/feature/hs/hs_descriptor.c
```diff
@@ -2475,6 +2504,84 @@ hs_desc_intro_point_free_(hs_desc_intro_point_t *ip)
   tor_free(ip);
 }
 
+/* Build a fake client info for the descriptor */
+void
```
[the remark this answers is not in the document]
dgoulet-tor: Commit c6403e6aa7

src/feature/hs/hs_descriptor.c
dgoulet-tor: Hmmm the entire HSv3 subsystem uses `/*` unfortunately... So we could do one commit to fix all of those in the `hs/` directory :).

src/feature/hs/hs_descriptor.c
dgoulet-tor: Fixup commit fe50c14be8

src/feature/hs/hs_client.c
dgoulet-tor: Fixup commit: 9d1fe4817b

src/feature/hs/hs_client.c
dgoulet-tor: Since the format is different, we choose to keep it separated. I don't see much how we could combine without adding complexity by passing to the function some sort of format description to look for?...

src/feature/hs/hs_client.c
dgoulet-tor: Fixup commit: 4ba612ad71

src/feature/hs/hs_descriptor.c
dgoulet-tor: Fixup commit: 97404e3a5c

src/feature/hs/hs_descriptor.c
dgoulet-tor: Yes it is the same! Good find... I think we should fix that one in a future commit. It might move things around a bit too much for this big branch. Noting down to do this!

src/feature/hs/hs_descriptor.c
dgoulet-tor: Fixup commit 54d9da19cd

src/feature/hs/hs_descriptor.c
dgoulet-tor: I've improved a bit the message in fixup commit 1f98b67b81.

We could make it that the log is much more verbose but at this stage this is decoding of a descriptor so it will ultimately result in the SOCKS connection being closed client side because we can't properly decode the descriptor.

src/feature/hs/hs_service.c
dgoulet-tor: This is unfortunately needed... It will happen rarely that is when the service is HUP. 

The reason we need this is to detect if there are new client authorization configured for the service and if so, we need to build a new descriptor for the service and republish.

So for that, we have to go through all clients from both services object (old and new) and figure out if any of them changed (removed or added).

> Come to think of it, we should probably also scramble the lists before we actually use them, so a client doesn't learn anything from its position in the list.

Oh hmmmm, yeah I think that is not a stupid idea. We should scramble that list after loading the key file. I propose we do this after, noting down.

src/feature/hs/hs_service.c
dgoulet-tor: This has been fixed with a future commit in this branch.

src/feature/hs/hs_service.c
dgoulet-tor: Same as above.

src/feature/hs/hs_service.c
haxxpop: I agree to scramble the list. It's safer to not let the client know the position.
Yes it's better to improve the running time. Do we need to do it in this branch?

src/feature/hs/hs_descriptor.c
haxxpop: Yeah, I also have seen `/*` in many places :)

src/feature/hs/hs_descriptor.c
asn-d6: Hm. I'm not sure exactly what kind of attack that could enable, but it does seem like a sane KDF use choice so we can do it.

I made a torspec PR with a design that adds the HS public identity key to the KDF computation: 
https://github.com/torproject/torspec/pull/34

asn-d6: Please also see my branch `ticket20700_035_02` for two commits that implement the above.
https://github.com/asn-d6/tor/tree/ticket20700_035_02

Some test vectors broke but I don't have the script that generated them so I just updated them... Can you please rerun the script if you have it?

src/feature/hs/hs_service.c
nmathewson: hm, ok.  Actually, it might be safe to leave this alone, since the User option is not allowed to change while Tor is running.

src/feature/hs/hs_service.c
nmathewson: A separate branch is fine, but IMO it should happen in 0.3.5 (especially since it is easy to do)

src/feature/hs/hs_descriptor.c
```diff
@@ -2701,14 +2701,16 @@ hs_desc_encode_descriptor,(const hs_descriptor_t *desc,
...
-   * I will enable it again when I finish writing the decoding */
-  /* ret = hs_desc_decode_descriptor(*encoded_out, */
-  /*                                 desc->subcredential, NULL); */
-  /* if (BUG(ret < 0)) { */
-  /*   goto err; */
-  /* } */
+  /* Try to decode what we just encoded. Symmetry is nice!, but it is
+   * symmetric only if the client auth is disabled. That is, the descriptor
+   * cookie will be NULL. */
+  if (!descriptor_cookie) {
```
asn-d6: Can't we also test-decode even if we have a descriptor cookie?

haxxpop: The reason is that `hs_desc_decode_descriptor` parameter requires the client private key not the descriptor cookie and we can't find the private key using descriptor cookie.

And I think using a private key as a parameter of `hs_desc_decode_descriptor` is more reasonable because we use a private key to decode not a descriptor cookie

asn-d6: Hm, I see what you mean. IIf I'm not mistaken, I think with some medium-level refactoring we could make that smooth and also have it test-decode in the client auth case as well (so that this code path gets used).

Perhaps we can open a ticket about this.

nmathewson:
A later version of this was merged

PR #311, opened 2018-09-09
[Tor Trac #26368]: Consider circuit isolation when closing redundant intro points 

neelchauhan:
For the bug [here](https://trac.torproject.org/projects/tor/ticket/26368).

### DIFF
--- a/src/feature/rend/rendclient.c
+++ b/src/feature/rend/rendclient.c
@@ -360,10 +360,23 @@ rend_client_close_other_intros(const uint8_t *rend_pk_digest)
       origin_circuit_t *oc = TO_ORIGIN_CIRCUIT(c);
       if (oc->rend_data &&
           rend_circuit_pk_digest_eq(oc, rend_pk_digest)) {
-        log_info(LD_REND|LD_CIRC, "Closing introduction circuit %d that we "
-                 "built in parallel (Purpose %d).", oc->global_identifier,
-                 c->purpose);
-        circuit_mark_for_close(c, END_CIRC_REASON_IP_NOW_REDUNDANT);
+        int stream_is_not_ok = 0;
+
+        for (edge_connection_t *oc_stream = oc->p_streams; oc_stream != NULL;
+             oc_stream = oc_stream->next_stream) {
+          if (!connection_edge_compatible_with_circuit(
+                EDGE_TO_ENTRY_CONN(oc_stream), TO_ORIGIN_CIRCUIT(c))) {
+            stream_is_not_ok = 1;
+            break;
+          }
+        }
+
+        if (stream_is_not_ok) {
+          log_info(LD_REND|LD_CIRC, "Closing introduction circuit %d that we "
+                   "built in parallel (Purpose %d).", oc->global_identifier,
+                   c->purpose);
+          circuit_mark_for_close(c, END_CIRC_REASON_IP_NOW_REDUNDANT);
+        }
       }
     }
   }


src/feature/rend/rendclient.c
```diff
@@ -350,15 +350,33 @@ rend_client_rendcirc_has_opened(origin_circuit_t *circ)
...
+           ec_oc = ec_oc->next_stream) {
+        for (edge_connection_t *ec_circ = circ->p_streams; ec_circ != NULL;
+             ec_circ = ec_circ->next_stream) {
+          if (EDGE_TO_ENTRY_CONN(ec_oc)->entry_cfg.isolation_flags ==
+              EDGE_TO_ENTRY_CONN(ec_circ)->entry_cfg.isolation_flags) {
+            isolation_bits_match = 1;
+            goto validate;
+          }
+        }
+      }
```
dgoulet-tor: Ok so I need to fully understand this part.

The intro circuit `circ` received an ACKED so we want to close all other intro circuits matching the .onion *unless* a circuit is flagged to be isolated.

The problem I see with the above is that _all_ isolation flags need to match. But I think what we need is to NOT close any circuits that have *an* isolation flag for the .onion (I'm not entirely sure if all isolation flags apply to onion circuits, would be good to validate that. First glance seems `isolation_flags_mixed` is used to match circuits in `circuit_is_better()`).

In other words, if I have two matching circuits for the same .onion and the circuit (here `oc`) has at least one stream with an isolation flag (again, relevant to onion service), we shouldn't close it.

This is actually more complicated than I thought. A unit test would really be desirable because I'm still unsure of the possible edge cases and even if we are doing it right :S...

neelchauhan: When I check if a `oc` has at least one isolation flag, should I check for *any* isolation flag, or a specific one?

If it is the former, I am thinking about something like this:

    diff --git a/src/feature/rend/rendclient.c b/src/feature/rend/rendclient.c
    index 10b67ceda..dc2d33281 100644
    --- a/src/feature/rend/rendclient.c
    +++ b/src/feature/rend/rendclient.c
    @@ -361,10 +361,21 @@ rend_client_close_other_intros(const uint8_t *rend_pk_digest)
           origin_circuit_t *oc = TO_ORIGIN_CIRCUIT(c);
           if (oc->rend_data &&
               rend_circuit_pk_digest_eq(oc, rend_pk_digest)) {
    -        log_info(LD_REND|LD_CIRC, "Closing introduction circuit %d that we "
    -                 "built in parallel (Purpose %d).", oc->global_identifier,
    -                 c->purpose);
    -        circuit_mark_for_close(c, END_CIRC_REASON_IP_NOW_REDUNDANT);
    +        int has_isolation_flag = 0;
    +        for (edge_connection_t *oc_stream = oc->p_streams; oc_stream != NULL;
    +             oc_stream = oc_stream->next_stream) {
    +          if (EDGE_TO_ENTRY_CONN(oc_stream)->entry_cfg.isolation_flags) {
    +            has_isolation_flag = 1;
    +            break;
    +          }
    +        }
    +
    +        if (!has_isolation_flag) {
    +          log_info(LD_REND|LD_CIRC, "Closing introduction circuit %d that we "
    +                   "built in parallel (Purpose %d).", oc->global_identifier,
    +                   c->purpose);
    +          circuit_mark_for_close(c, END_CIRC_REASON_IP_NOW_REDUNDANT);
    +        }
           }
         }
       }

The code in this comment seems much simpler than the code at this branch. Would that be okay? If so, the above code will replace the code in this comment.

Also if the above code is okay, will we still need a unit test?

If it is the latter, which flag should I check for?

PR #315, opened 2018-09-10
Defer reporting bootstrap progress (ticket 27169)

tlyu:


src/feature/control/control.c
```diff
@@ -7121,41 +7122,36 @@ control_event_bootstrap(bootstrap_status_t status, int progress)
     }
   }
 
-  if (status > bootstrap_percent ||
-      (progress && progress > bootstrap_percent)) {
-    int loglevel = LOG_NOTICE;
-    bootstrap_status_to_string(status, &tag, &summary);
```
dgoulet-tor: By removing this, I believe `tag` and `summary` will be undefined if they hit the snprintf below.

src/feature/control/control.c
```diff
@@ -7206,9 +7212,7 @@ control_event_bootstrap_problem(const char *warn, const char *reason,
   if (we_are_hibernating())
     dowarn = 0;
 
-  while (status>=0 && bootstrap_status_to_string(status, &tag, &summary) < 0)
-    status--; /* find a recognized status string based on current progress */
-  status = bootstrap_percent; /* set status back to the actual number */
+  tor_assert(bootstrap_status_to_string(bootstrap_phase, &tag, &summary) == 0);
```
dgoulet-tor: After this removal, is `status` correct? It points to `bootstrap_percent` from the initial assignment and it is now never changed. Is this accurate since we seems to rely on the `bootstrap_phase` but still the status value is used in the log after ("Problem bootstrapping....") ?

src/core/or/connection_or.c
```diff
@@ -705,6 +705,7 @@ connection_or_finished_connecting(or_connection_t *or_conn)
   log_debug(LD_HANDSHAKE,"OR connect() to router at %s:%u finished.",
             conn->address,conn->port);
   control_event_bootstrap(BOOTSTRAP_STATUS_HANDSHAKE, 0);
+  control_event_boot_first_orconn();
```
dgoulet-tor: So the bootstrap process starts with **connecting** to a directory server (5%). Then, right here at this point, the boostrap will change to `Finishing handshake with directory server` (10%). With this addition, we'll flag that we have a "first connect() to a OR".

Which means that we can't go pass 10% in the first place if we haven't at least connected to our directory server right? Which makes me wonder if this will really delay any directory information bootstrap progress if our first connect to an OR conn is always at the 10% mark? Won't we be stuck at 5% if we are unable to connect to an  OR? (in this case at 5%, the directory server).

src/feature/control/control.c
tlyu: This is a minor mistake I made when rewriting history. 15c24d669f restores the call to `bootstrap_status_to_string()`, so the right thing happens if all the patches are applied. I can fix this one commit by editing the branch if you like.

src/feature/control/control.c
tlyu: Yes, it should be correct. `status` now never gets modified by the loop, so it remains set to `bootstrap_percent` by its initializer. We use `bootstrap_phase` to get the milestone tag and summary, but we report `bootstrap_percent` (or equivalently, `status`) as the "percent" progress in the log message.

If it helps, I can eliminate the `status` variable and use `bootstrap_phase` and `bootstrap_percent` for clarity.

src/core/or/connection_or.c
tlyu: I think being stuck at 5% or 10% gives the user a clearer indication of what's wrong than being stuck at 80%.

>Which means that we can't go pass 10% in the first place if we haven't at least connected to our directory server right?

Empirically, if there is enough cached directory info, bootstrap will jump from 0% to 80% without necessarily being able to connect to _any_ relay or bridge. I can paste example logs of "warm cache" cases both before and after these patches, if that helps. (probably to the Trac ticket though)

dgoulet-tor: > Empirically, if there is enough cached directory info, bootstrap will jump from 0% to 80% without necessarily being able to connect to _any_ relay or bridge. I can paste example logs of "warm cache" cases both before and after these patches, if that helps. (probably to the Trac ticket though)

Oh ok I see the jump from 0 to 80 but stuck at the connecting to OR step which is not fun for the user.

PR #317, opened 2018-09-11
[Tor Trac #27490]: When ClientPreferIPv6ORPort is set to auto, and a relay is being chosen for a directory or orport connection, prefer IPv4 or IPv6 at random

neelchauhan:
For the ticket [here](https://trac.torproject.org/projects/tor/ticket/27490).

### DIFF
--- a/changes/ticket27490
+++ b/changes/ticket27490
@@ -1,0 +1,6 @@
+  o Minor features (ipv6):
+    - If ClientPreferIPv6ORPort is set to auto and a relay is being chosen
+      for a Directory or ORPort connection, we choose IPv4 or IPv6 at random.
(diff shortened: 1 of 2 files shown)


teor2345:
I think that this patch will break existing configs when:
* there are a small number of Bridges (or EntryNodes),
* they are all marked as IPv4-preferred when they are loaded, and
* the device running tor only has IPv6.

We can fix this bug in a few different ways:
* make sure we always fall back to non-preferred addresses if all preferred addresses fail:
  * we will find these bugs when we test
* reset the preferred flags on all nodes if all preferred addresses failL
  * I think we should remove the flag instead
* stop setting the IPv6 preferred flag on nodes:
  * see https://trac.torproject.org/projects/tor/ticket/27648

PR #321, opened 2018-09-12
Add an --enable-pic option for people who hack Tor into libraries

nmathewson:
Closes ticket 23846

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -55,6 +55,8 @@ AC_ARG_ENABLE(zstd-advanced-apis,
    AS_HELP_STRING(--disable-zstd-advanced-apis, [Build without support for zstd's "static-only" APIs.]))
 AC_ARG_ENABLE(nss,
    AS_HELP_STRING(--enable-nss, [Use Mozilla's NSS TLS library. (EXPERIMENTAL)]))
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -1157,7 +1163,9 @@ m4_ifdef([AS_VAR_IF],[
     TOR_CHECK_CFLAGS(-Wstack-protector)
     TOR_CHECK_CFLAGS(--param ssp-buffer-size=1)
     if test "$bwin32" = "false" && test "$enable_libfuzzer" != "yes" && test "$enable_oss_fuzz" != "yes"; then
-       TOR_CHECK_CFLAGS(-fPIE)
+       if test "$enable_pic" != "yes"; then
+           TOR_CHECK_CFLAGS(-fPIE)
+       fi
        TOR_CHECK_LDFLAGS(-pie, "$all_ldflags_for_check", "$all_libs_for_check")
```
dgoulet-tor: Shouldn't that LDFLAGS check be in the if/fi ?

Also, no LDFLAGS for -fPIC ?

nmathewson: No -- I'll document why. It's a little subtle.

PR #323, opened 2018-09-13
Ticket27678

mikeperry-tor:
https://trac.torproject.org/projects/tor/ticket/27678

### DIFF
--- a/src/core/or/command.c
+++ b/src/core/or/command.c
@@ -480,6 +480,8 @@ command_process_relay_cell(cell_t *cell, channel_t *chan)
   const or_options_t *options = get_options();
   circuit_t *circ;
   int reason, direction;
+  uint32_t orig_delivered_bw = 0;
+  uint32_t orig_overhead_bw = 0;
 
   circ = circuit_get_by_circid_channel(cell->circ_id, chan);
 
@@ -512,6 +514,11 @@ command_process_relay_cell(cell_t *cell, channel_t *chan)
     /* Count the payload bytes only. We don't care about cell headers */
     ocirc->n_read_circ_bw = tor_add_u32_nowrap(ocirc->n_read_circ_bw,
                                                CELL_PAYLOAD_SIZE);
+
+    /* Stash the original delivered and overhead values, so we can inform the
+     * control port about dropped cells immediately after processing below. */
+    orig_delivered_bw = ocirc->n_delivered_read_circ_bw;
+    orig_overhead_bw = ocirc->n_overhead_read_circ_bw;
   }
 
   if (!CIRCUIT_IS_ORIGIN(circ) &&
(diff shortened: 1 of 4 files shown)


src/core/or/command.c
```diff
@@ -512,6 +514,11 @@ command_process_relay_cell(cell_t *cell, channel_t *chan)
     /* Count the payload bytes only. We don't care about cell headers */
     ocirc->n_read_circ_bw = tor_add_u32_nowrap(ocirc->n_read_circ_bw,
                                                CELL_PAYLOAD_SIZE);
+
+    /* Stash the original delivered and overhead values, so we can inform the
+     * control port about dropped cells immediately after processing below. */
+    orig_delivered_bw = ocirc->n_delivered_read_circ_bw;
```
asn-d6: What parts of the code can modify these values and we are keeping the originals? Perhaps we can put it in the comment.

src/core/or/command.c
```diff
@@ -535,6 +542,7 @@ command_process_relay_cell(cell_t *cell, channel_t *chan)
                (unsigned)cell->circ_id);
       if (CIRCUIT_IS_ORIGIN(circ)) {
         circuit_log_path(LOG_WARN, LD_OR, TO_ORIGIN_CIRCUIT(circ));
+        control_event_circ_bandwidth_used_for_circ(TO_ORIGIN_CIRCUIT(circ));
```
asn-d6: We seem to be calling this function from three places. How can we know that these are the only places we need to call it from?

src/core/or/command.c
```diff
@@ -560,9 +568,24 @@ command_process_relay_cell(cell_t *cell, channel_t *chan)
...
 
+  if (CIRCUIT_IS_ORIGIN(circ)) {
+    origin_circuit_t *ocirc = TO_ORIGIN_CIRCUIT(circ);
+
+    /* If neither the delivered nor overhead values changed, this cell
+     * was dropped due to being invalid. Emit a control port event for CIRC_BW,
+     * so the controller can react quickly. */
+    if (orig_delivered_bw == ocirc->n_delivered_read_circ_bw &&
+        orig_overhead_bw == ocirc->n_overhead_read_circ_bw) {
+      control_event_circ_bandwidth_used_for_circ(ocirc);
```
asn-d6: Does it matter if we double-call this new function and emit two control events? Maybe we can add a guard in the function to make sure we don't double or triple call it?

src/core/or/command.c
```diff
@@ -480,6 +480,8 @@ command_process_relay_cell(cell_t *cell, channel_t *chan)
   const or_options_t *options = get_options();
   circuit_t *circ;
   int reason, direction;
+  uint32_t orig_delivered_bw = 0;
```
asn-d6: Nitpick: Commit msg is wrong, references 27679 instead of 27678.

src/core/or/command.c
mikeperry-tor: Clarified comment in 02b5a6c67f2d3287ffd2d5ca773ff3534d1c774c

src/core/or/command.c
mikeperry-tor: We call it when we get an error to close the circuit, or when the codepaths called from circuit_receive_relay_cell() fails to update bw values (because the cell was not valid).

Comment 02b5a6c67f2d3287ffd2d5ca773ff3534d1c774c also clarifies this.

src/core/or/command.c
mikeperry-tor: It does not matter if we call this function multiple times. It already has a guard to prevent redundant events. See updated comments in 66b05e34cc58472085f3ab1ba92343558ba2a757.

src/core/or/command.c
mikeperry-tor: Ok I will fix this in a new branch. I don't like that commit summary anyway, since it doesn't change code.

PR #325, opened 2018-09-13
#27620:  Use trunnel to parse and generate SOCKS wire format in tor-resolve 

rl1987:
https://trac.torproject.org/projects/tor/ticket/27620


nmathewson:
Squashed and merged as #556

PR #327, opened 2018-09-13
Bug4700 rebase

asn-d6:


changes/bug4700
```diff
@@ -0,0 +1,5 @@
+  o Minor features (onion services):
+    - Next generation onion services can now use the per-service
```
dgoulet-tor: I would say we should use "verson 3" instead of "next gen".

src/app/config/config.c
```diff
@@ -457,6 +457,7 @@ static config_var_t option_vars_[] = {
   VAR("HiddenServiceMaxStreams",LINELIST_S, RendConfigLines, NULL),
   VAR("HiddenServiceMaxStreamsCloseCircuit",LINELIST_S, RendConfigLines, NULL),
   VAR("HiddenServiceNumIntroductionPoints", LINELIST_S, RendConfigLines, NULL),
+  VAR("HiddenServiceExportCircuitID", LINELIST_S,  RendConfigLines, NULL),
```
dgoulet-tor: This option seems to be about controlling if we export the circuit ID in the `HAProxy` format... What about we make it a bit more specific to HAProxy?

src/core/or/connection_edge.c
```diff
@@ -597,31 +597,34 @@ connected_cell_format_payload(uint8_t *payload_out,
...
 
-/* DOCDOCDOC */
-static void
-send_ha_proxy_header(const edge_connection_t *edge_conn,
-                     connection_t *conn)
+/* This is an onion service client connection: Export the client circuit ID
+ * according to the HAProxy proxy protocol. */
+STATIC void
+export_hs_client_circuit_id(const edge_connection_t *edge_conn,
+                            connection_t *conn)
```
dgoulet-tor: Since this is HAProxy format specific, I liked the mention of "haproxy" in the name.

src/core/or/connection_edge.c
```diff
@@ -618,6 +617,11 @@ export_hs_client_circuit_id(const edge_connection_t *edge_conn,
     src_port = gid & 0x0000ffff;
   }
 
+  /* Grab the original dest port from the hs ident */
+  if (edge_conn->hs_ident) {
+    dst_port = edge_conn->hs_ident->orig_virtual_port;
+  }
```
dgoulet-tor: Can we really have a `dst_port = 0` that makes sense to send the HAProxy header?

changes/bug4700
ahf: Agreed

src/app/config/config.c
ahf: I think the feature should be independent of the HAProxy implementation. For now we *only* support HAProxy here (and we might only do that for a long time?), but maybe there are other systems like that we could do (for Freenode's v3 IRCd for example?). Maybe it could make sense to have a flag here so the config line in `torrc` becomes something like:

`HiddenServiceExportCircuitID HAProxy` ?

src/core/or/connection_edge.c
```diff
@@ -597,6 +597,40 @@ connected_cell_format_payload(uint8_t *payload_out,
...
+/* This is an onion service client connection: Export the client circuit ID
+ * according to the HAProxy proxy protocol. */
+STATIC void
+export_hs_client_circuit_id(const edge_connection_t *edge_conn,
+                            connection_t *conn)
+{
+  char *buf;
+  const char dst_ipv6[] = "::1";
+  /* See RFC4193 regarding fc00::/7 */
+  const char src_ipv6_prefix[] = "fc00:dead:beef:4dad:";
```
ahf: Nice.

src/test/test_helpers.c
```diff
@@ -125,6 +125,22 @@ connection_write_to_buf_mock(const char *string, size_t len,
   buf_add(conn->outbuf, string, len);
 }
 
+char *
+buf_get_contents(buf_t *buf, size_t *sz_out)
+{
```
ahf: Add `tor_assert(buf)` and `tor_assert(sz_out)` ?

src/app/config/config.c
asn-d6: @ahf here is a comment by roger that de-suggests that: https://trac.torproject.org/projects/tor/ticket/4700#comment:17

Not that it matters, but some extra feedback.

src/core/or/connection_edge.c
ahf: Actually, I'm unsure about this? Is it fine this is a static string or should it be configurable?

src/test/test_helpers.c
asn-d6: ✓

changes/bug4700
asn-d6: ✓

src/core/or/connection_edge.c
asn-d6: ✓ (0ca04c004d)

src/core/or/connection_edge.c
mahrud: I would prefer this to be configurable too, but I'm not sure how to do so without adding more torrc options. Maybe an environmental variable?

ahf:
This have already been merged. Closing.

PR #328, opened 2018-09-13
Add support for lower-level byte counting with NSS

nmathewson:
This is harder than with OpenSSL, since OpenSSL counts the bytes on
its own and NSS doesn't.  To fix this, we need to define a new
PRFileDesc layer that has its own byte-counting support.

Closes ticket 27289.

### DIFF
--- a/src/lib/tls/nss_countbytes.c
+++ b/src/lib/tls/nss_countbytes.c
@@ -53,8 +53,8 @@ typedef struct tor_nss_bytecounts_t {
 /**
  * Initialize this module, if it is not already initialized.
  **/
-static void
-countbytes_init(void)
+void
+tor_nss_countbytes_init(void)
 {
   if (countbytes_initialized)
     return;
@@ -192,8 +192,8 @@ countbytes_recv_fn(PRFileDesc *fd, void *buf, PRInt32 amount,
 PRFileDesc *
 tor_wrap_prfiledesc_with_byte_counter(PRFileDesc *stack)
 {
-  if (PREDICT_UNLIKELY(! countbytes_initialized)) {
-    countbytes_init();
+  if (BUG(! countbytes_initialized)) {
+    tor_nss_countbytes_init();
   }
 
   tor_nss_bytecounts_t *bytecounts = tor_malloc_zero(sizeof(*bytecounts));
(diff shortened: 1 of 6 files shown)


src/lib/tls/nss_countbytes.c
```diff
@@ -0,0 +1,244 @@
...
+}
+
+/**
+ * Return the tor_nss_bytecounts_t object for a given IO layer. Asserts that
+ * the IO layer is in fact a layer created by this module.
+ */
+static tor_nss_bytecounts_t *
+get_counts(PRFileDesc *fd)
+{
+  tor_assert(fd->identity = countbytes_layer_id);
```
ahf: Should probably be `==`?

src/lib/tls/nss_countbytes.c
```diff
@@ -0,0 +1,244 @@
...
+  return (tor_nss_bytecounts_t*) fd->secret;
+}
+
+/** Helper: increment the read-count of an fd by n. */
+#define INC_READ(fd, n) do {                            \
```
ahf: Should we use `STMT_BEGIN` and `STMT_END` here?

src/lib/tls/nss_countbytes.c
```diff
@@ -0,0 +1,244 @@
...
+    get_counts(fd)->n_read += (n);                      \
+  } while (0)
+
+/** Helper: increment the write-count of an fd by n. */
+#define INC_WRITTEN(fd, n) do {                            \
```
ahf: Same as above.

src/lib/tls/nss_countbytes.c
```diff
@@ -0,0 +1,244 @@
...
+/**
+ * Wrap a PRFileDesc from NSPR with a new PRFileDesc that will count the
+ * total number of bytes read and written.  Return the new PRFileDesc.
+ *
+ * This function takes ownership of its input.
+ */
+PRFileDesc *
+tor_wrap_prfiledesc_with_byte_counter(PRFileDesc *stack)
+{
+  if (PREDICT_UNLIKELY(! countbytes_initialized)) {
```
ahf: It wouldn't be cleaner to explicitly initialize this when Tor starts instead of doing it lazily right before the subsystem is needed?

nmathewson:
This was squashed and merged

PR #330, opened 2018-09-13
Bug23512 v2 032

mikeperry-tor:


src/or/circuitlist.c
```diff
@@ -1844,6 +1845,58 @@ circuit_mark_all_dirty_circs_as_unusable(void)
...
 
+/**
+ * Report any queued cells on or_circuits as written in our bandwidth
+ * totals, for the specified channel direction.
+ *
+ * When we close a circuit or clear its cell queues, we've read
+ * data and recorded those bytes in our read statistics, but we're
+ * not going to write it. This discrepancy can be used by an adversary
+ * to infer information from our public relay statistics and perform
+ * attacks such as guard discovery.
```
dgoulet-tor: Can I ask you to add a line here saying that this function is part of the "fast path" and MUST remain as fast as possible? I'm on a sekrit goal of, over time, identifying clearly all those functions :P.

src/or/circuitlist.c
```diff
@@ -1844,6 +1845,58 @@ circuit_mark_all_dirty_circs_as_unusable(void)
...
+
+  if (dir == CIRCUIT_N_CHAN)
+    chan = c->n_chan;
+  else
+    chan = or_circ->p_chan;
+
+  if (dir == CIRCUIT_N_CHAN)
+    cells = c->n_chan_cells.n;
+  else
+    cells = or_circ->p_chan_cells.n;
```
dgoulet-tor: Instead of adding the `circuit_channel_direction_t` and doing these if/else, can't we simply pass the channel to the function?

src/or/circuitlist.c
```diff
@@ -1844,6 +1845,58 @@ circuit_mark_all_dirty_circs_as_unusable(void)
...
+  /* If we still know the chan, determine real cell size. Otherwise,
+   * assume it's a wide circid channel */
+  if (chan)
+    cell_size = get_cell_network_size(chan->wide_circ_ids);
+  else
+    cell_size = CELL_MAX_NETWORK_SIZE;
+
+  /* The missing written bytes are the cell counts times their cell
+   * size plus TLS per cell overhead */
+  written_sync = cells*(cell_size+TLS_PER_CELL_OVERHEAD);
```
dgoulet-tor: Both `cells` and `cell_size` are `size_t`, I would be extra careful with the overflow here even though unlikely. Just to be safe. The 32 bits platform always scares me there :P.

src/or/channeltls.h
```diff
@@ -12,6 +12,8 @@
 #include "or.h"
 #include "channel.h"
 
+#define TLS_PER_CELL_OVERHEAD 29
+
```
dgoulet-tor: Once we get a hang on pastly, would be good to have a comment explaining that value. Lets keep that in mind.

src/or/circuitlist.c
dgoulet-tor: Nvm here... The queue is on the circuit, not the channel so we need the direction to know which queue to get.

PR #338, opened 2018-09-14
#27675: Tweak test_rebind.py

rl1987:
https://trac.torproject.org/projects/tor/ticket/27675

### DIFF
--- a/src/test/test_rebind.py
+++ b/src/test/test_rebind.py
@@ -1,5 +1,3 @@
-#!/usr/bin/python3
-
 from __future__ import print_function
 
 import sys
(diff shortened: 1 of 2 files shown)


src/test/test_rebind.py
```diff
@@ -35,6 +35,9 @@ def pick_random_port():
 
     return port
 
+if sys.hexversion < 0x02070000:
```
tlyu: Should probably also check for >=3.1 if 3.x because the empty format string field is a >=3.1 feature.

rl1987: Fixed in 22518390b0e66a4a6f9b0be12533722d62422759.

PR #345, opened 2018-09-17
Make CacheDirectoryGroupReadable an autobool.

nmathewson:
Since the default cache directory is the same as the default data
directory, we don't want the default CacheDirectoryGroupReadable
value (0) to override an explicitly set "DataDirectoryGroupReadable
1".

To fix this, I'm making CacheDirectoryGroupReadable into an
autobool, and having the default (auto) value mean "Use the value of
DataDirectoryGroupReadable if the directories are the same, and 0
otherwise."

Fixes bug 26913; bugfix on 0.3.3.1-alpha when the CacheDirectory
option was introduced.


nmathewson:
This was merged and marked for backport.

PR #349, opened 2018-09-17
Bug23512 v4 032

mikeperry-tor:


nmathewson:
A version of this was squashed and merged; see tickets for subsequent developments/

PR #355, opened 2018-09-18
hs-v3: Consolidate descriptor cookie computation code

dgoulet-tor:
Both client and service had their own code for this. Consolidate into one
place so we avoid duplication.

Closes #27549


nmathewson:
Squashed and merged

PR #357, opened 2018-09-19
Fix arm compilation with openssl <1.1

nmathewson:
Bug 27781; bugfix on 0.3.5.1-alpha.


nmathewson:
partially backported and merged forward.

PR #358, opened 2018-09-19
hs-v3: Don't assert if we don't have the intro point descriptor

dgoulet-tor:
When sending the INTRODUCE1 cell, we lookup the intro point onion key but
there is a possibility that between opening the circuit and sending the cell
the relay descriptor of the node could have been removed leading to not
finding the onion key and a non fatal assert client side which also closes the
SOCKS connection without trying anything else.

We now look for that missing onion key and recover gracefully by closing the
intro point circuit, noting down the error in the failure cache and tor then
tries a new intro point.

Note that the send introduce1 function now makes sure to close the intro
circuit in case of an error because the caller is expecting it by validating
it with a non fatal assert. It is especially important so we can recover and
select a new intro point.

Fixes #27774

### DIFF
--- a/changes/ticket27774
+++ b/changes/ticket27774
@@ -1,0 +1,4 @@
+  o Minor bugfixes (hidden service v3):
+    - Client side would dump a stack trace if tor doesn't have the descriptor
+      for the intro point it is trying to connect to. Fixes bug 27774; bugfix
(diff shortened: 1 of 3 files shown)


nmathewson:
This has been squashed and merged to master.

PR #309, opened 2018-09-05
Connection refactor1

nmathewson:
Created PR so that #27414 gets CI testing

PR #359, opened 2018-09-20
#27417: Connection refactor1

teor2345:
Obsoletes #309

### DIFF
--- a/src/core/mainloop/main.c
+++ b/src/core/mainloop/main.c
@@ -1002,77 +1002,20 @@ conn_close_if_marked(int i)
 
   if ((SOCKET_OK(conn->s) || conn->linked_conn) &&
       connection_wants_to_flush(conn)) {
-    ssize_t sz = connection_bucket_write_limit(conn, now);
-    if (!conn->hold_open_until_flushed)
-      log_info(LD_NET,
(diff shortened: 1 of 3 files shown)


src/core/mainloop/main.c
```diff
@@ -1002,77 +1002,20 @@ conn_close_if_marked(int i)
...
-    if (conn->linked_conn) {
-      retval = buf_move_to_buf(conn->linked_conn->inbuf, conn->outbuf,
-                               &conn->outbuf_flushlen);
-      if (retval >= 0) {
+    retval = connection_flush_before_close(conn);
+
+    if (conn->linked_conn && retval >= 0) {
         /* The linked conn will notice that it has data when it notices that
          * we're gone. */
         connection_start_reading_from_linked_conn(conn->linked_conn);
```
dgoulet-tor: So here is a potential problem. With this refactor, this function can be called _after_ the code below here that is moved into `connection_flush_before_close()`.

I'm not entirely sure of the implications here but this is a behavior change and thus this refactoring just became much more complicated.

For this entire commit, I would strongly recommend to break it down in small pieces that is "only code moving" commit and then behavior change commit. Else, this makes the review of this code very complicated because every bit and pieces needs to be looked if git can't recognize this as "only code move".

PR #362, opened 2018-09-20
hs-v3: Don't assert if we don't have the intro point descriptor

dgoulet-tor:
When sending the INTRODUCE1 cell, we build the needed data for the cell and
there might be issues with the rendezvous point node_t that makes us not able
to do so. In that case, close the rendezvous circuit and return a transient
error meaning that Tor can recover by selecting a new rendezvous point.

If we are unable to build the INTRODUCE1 cell, we'll also close the rendezvous
circuit betting on the fact that some data wasn't encodable. It is practically
impossible in theory.

Fixes #27774

### DIFF
--- a/src/feature/hs/hs_circuit.c
+++ b/src/feature/hs/hs_circuit.c
@@ -646,16 +646,16 @@ get_lspecs_from_node(const node_t *node, smartlist_t *lspecs)
  * already allocated intro1_data object with the needed key material and link
  * specifiers.
  *
- * If rp_node has an invalid primary address, intro1_data->link_specifiers
- * will be an empty list. Otherwise, this function can't fail. The ip
- * MUST be a valid object containing the needed keys and authentication
- * method. */
-static void
+ * Return 0 on success or a negative value if we couldn't properly filled the
+ * introduce1 data from the RP node. In other word, it means the RP node is
+ * unusable to use in the introduction. */
+static int
 setup_introduce1_data(const hs_desc_intro_point_t *ip,
                       const node_t *rp_node,
                       const uint8_t *subcredential,
                       hs_cell_introduce1_data_t *intro1_data)
 {
+  int ret = -1;
   smartlist_t *rp_lspecs;
 
   tor_assert(ip);
@@ -667,6 +667,11 @@ setup_introduce1_data(const hs_desc_intro_point_t *ip,
    * circuit that we've picked previously. */
   rp_lspecs = smartlist_new();
   get_lspecs_from_node(rp_node, rp_lspecs);
+  if (smartlist_len(rp_lspecs) == 0) {
+    /* We can't rendezvous without link specifiers. */
+    smartlist_free(rp_lspecs);
+    goto end;
+  }
 
   /* Populate the introduce1 data object. */
   memset(intro1_data, 0, sizeof(hs_cell_introduce1_data_t));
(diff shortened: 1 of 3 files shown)


src/feature/hs/hs_client.c
```diff
@@ -576,10 +576,21 @@ send_introduce1(origin_circuit_t *intro_circ,
   /* Send the INTRODUCE1 cell. */
   if (hs_circ_send_introduce1(intro_circ, rend_circ, ip,
                               desc->subcredential) < 0) {
-    /* Unable to send the cell, the intro circuit has been marked for close so
-     * this is a permanent error. */
-    tor_assert_nonfatal(TO_CIRCUIT(intro_circ)->marked_for_close);
-    goto perm_err;
+    if (TO_CIRCUIT(intro_circ)->marked_for_close) {
```
asn-d6: Why not do this in `hs_circ_send_introduce1()` instead of clobbering this function with this info, and trying to infer what `hs_circ_send_introduce1()` has done?

src/feature/hs/hs_client.c
```diff
@@ -576,10 +576,21 @@ send_introduce1(origin_circuit_t *intro_circ,
...
+      /* If the introduction circuit was closed, we were unable to send the
+       * cell for some reasons. In any case, the intro circuit has to be
+       * closed by the above function. We'll return a transient error so tor
+       * can recover and pick a new intro point. To avoid picking that same
+       * intro point, we'll note down the intro point failure so it doesn't
+       * get reused. */
+      hs_cache_client_intro_state_note(service_identity_pk,
+                                       &intro_circ->hs_ident->intro_auth_pk,
+                                       INTRO_POINT_FAILURE_UNREACHABLE);
```
asn-d6: UNREACHABLE or GENERIC?

src/feature/hs/hs_client.c
dgoulet-tor: Good question. The reason is that I wanted that to be a "client action" and not part of the "HS circuit" layer. The client there decides to note down the intro point client cache failure, not the circuit layer.

In other word, I wanted to not have `hs_circuit` to callback into `hs_client`.

src/feature/hs/hs_client.c
dgoulet-tor: Good catch. Fixup commit to make it GENERIC: b5d938b617

src/feature/hs/hs_circuit.c
```diff
@@ -1130,14 +1144,14 @@ hs_circ_send_introduce1(origin_circuit_t *intro_circ,
...
-   * bad. */
-  if (BUG(!intro1_data.link_specifiers) ||
-      !smartlist_len(intro1_data.link_specifiers)) {
-    log_warn(LD_REND, "Unable to get link specifiers for INTRODUCE1 cell on "
-             "circuit %u.", TO_CIRCUIT(intro_circ)->n_circ_id);
-    goto done;
+
+  /* We should never select an invalid rendezvous point in theory but if we
+   * do, this function will fail to populate the introduce data. */
+  if (BUG(setup_introduce1_data(ip, exit_node, subcredential,
```
asn-d6: Hm, why are we wrapping this into a BUG()? If it's a BUG, why do we introduce whole new behavior just for a buggy possibility? Instead we should fix the bug.

Or is this an edge-case that can happen normally? In this case we should not call BUG.

nmathewson:
This has been squashed and merged to master.

PR #363, opened 2018-09-20
Split routerlist dirserv

nmathewson:


nmathewson:
This was revised, squashed, and merged.

PR #370, opened 2018-09-22
#27325: Rework NETINFO wire format handling to rely on trunnel

rl1987:
https://trac.torproject.org/projects/tor/ticket/27325


src/trunnel/netinfo.trunnel
```diff
@@ -0,0 +1,21 @@
+const NETINFO_ADDR_TYPE_IPV4 = 4;
+const NETINFO_ADDR_TYPE_IPV6 = 6;
+
+struct netinfo_addr {
+  u8 addr_type IN [NETINFO_ADDR_TYPE_IPV4, NETINFO_ADDR_TYPE_IPV6];
```
dgoulet-tor: Hmmm is it possible this is missing 3 values by any chance (looking at tor-spec.txt):

```
   "Type" is one of:
      0x00 -- Hostname
      0x04 -- IPv4 address
      0x06 -- IPv6 address
      0xF0 -- Error, transient
      0xF1 -- Error, nontransient
```

src/core/or/channeltls.c
```diff
@@ -1719,52 +1747,69 @@ channel_tls_process_netinfo_cell(cell_t *cell, channel_tls_t *chan)
...
   }
 
-  my_addr_type = (uint8_t) cell->payload[4];
-  my_addr_len = (uint8_t) cell->payload[5];
-  my_addr_ptr = (uint8_t*) cell->payload + 6;
-  end = cell->payload + CELL_PAYLOAD_SIZE;
-  cp = cell->payload + 6 + my_addr_len;
+  timestamp = netinfo_cell_get_timestamp(netinfo_cell);
 
+  netinfo_addr_t *my_addr = netinfo_cell_get_other_addr(netinfo_cell);
```
dgoulet-tor: Can we use the `const` getter here you think?

src/core/or/channeltls.c
```diff
@@ -1719,52 +1747,69 @@ channel_tls_process_netinfo_cell(cell_t *cell, channel_tls_t *chan)
...
 
+  if (tor_addr_from_netinfo_addr(&my_apparent_addr, my_addr) == -1) {
+    connection_or_close_for_error(chan->conn, 0);
+    netinfo_cell_free(netinfo_cell);
+    return;
+  }
+
   if (my_addr_type == RESOLVED_TYPE_IPV4 && my_addr_len == 4) {
-    tor_addr_from_ipv4n(&my_apparent_addr, get_uint32(my_addr_ptr));
+    uint32_t ipv4 = netinfo_addr_get_addr_ipv4(my_addr);
```
dgoulet-tor: I believe the `ipv4` will be network order here so I would try to reflect that in the name of the variable because we'll get confused really fast :S

dgoulet-tor: Actually, why do we need to do that if we have the address already in `my_apparent_addr` ? Can't we simply compare that with `me->addr` ?

src/trunnel/netinfo.trunnel
dgoulet-tor: Seems also that the are mapped from `RESOLVED_TYPE_*` values. We should put a strong comment to keep them insync.

src/core/or/channeltls.c
```diff
@@ -1719,52 +1747,69 @@ channel_tls_process_netinfo_cell(cell_t *cell, channel_tls_t *chan)
...
         !tor_addr_is_null(&me->ipv6_addr) &&
         tor_addr_eq(&my_apparent_addr, &me->ipv6_addr)) {
       TLS_CHAN_TO_BASE(chan)->is_canonical_to_peer = 1;
     }
   }
 
-  n_other_addrs = (uint8_t) *cp++;
-  while (n_other_addrs && cp < end-2) {
+  n_other_addrs = netinfo_cell_get_n_my_addrs(netinfo_cell);
+  for (size_t i = 0; i < n_other_addrs; i++) {
```
dgoulet-tor: `size_t i` --> `uint8_t i`. I say we match the type with our top limit `n_other_addrs`.

src/core/or/connection_or.c
```diff
@@ -2426,6 +2427,32 @@ connection_or_send_versions(or_connection_t *conn, int v3_plus)
...
+{
+  sa_family_t addr_family = tor_addr_family(tor_addr);
+
+  if (BUG(addr_family != AF_INET && addr_family != AF_INET6))
+    return NULL;
+
+  netinfo_addr_t *netinfo_addr = netinfo_addr_new();
+
+  if (addr_family == AF_INET) {
+    netinfo_addr_set_addr_type(netinfo_addr, NETINFO_ADDR_TYPE_IPV4);
```
dgoulet-tor: When decoding, we used `RESOLVED_TYPE_...`, here the `NETINFO_...` are used. Whichever we use, we should simply be consistent and stick to use one.

src/core/or/connection_or.c
```diff
@@ -2426,6 +2427,32 @@ connection_or_send_versions(or_connection_t *conn, int v3_plus)
...
+    netinfo_addr_set_len(netinfo_addr, 4);
+    netinfo_addr_set_addr_ipv4(netinfo_addr, tor_addr_to_ipv4h(tor_addr));
+  } else if (addr_family == AF_INET6) {
+    netinfo_addr_set_addr_type(netinfo_addr, NETINFO_ADDR_TYPE_IPV6);
+    netinfo_addr_set_len(netinfo_addr, 16);
+    uint8_t *ipv6_buf = netinfo_addr_getarray_addr_ipv6(netinfo_addr);
+    // XXX: this looks sketchy
+    memcpy(ipv6_buf, tor_addr->addr.in6_addr.s6_addr, 16);
+    // XXX: can we avoid memcpy here?
```
dgoulet-tor: The HS subsystem does it like this:

```
    const uint8_t *in6_addr = tor_addr_to_in6_addr8(&ap.addr);
    uint8_t *ipv6_array = link_specifier_getarray_un_ipv6_addr(ls);
    memcpy(ipv6_array, in6_addr, addr_len);
```

You can probably ignore `addr_len` and instead use 16.

src/core/or/connection_or.c
```diff
@@ -2448,49 +2474,65 @@ connection_or_send_netinfo,(or_connection_t *conn))
   memset(&cell, 0, sizeof(cell_t));
   cell.command = CELL_NETINFO;
 
+  netinfo_cell_t *netinfo_cell = netinfo_cell_new();
+
   /* Timestamp, if we're a relay. */
   if (public_server_mode(get_options()) || ! conn->is_outgoing)
-    set_uint32(cell.payload, htonl((uint32_t)now));
+    netinfo_cell_set_timestamp(netinfo_cell, (uint32_t)now);
```
dgoulet-tor: You sure you can remove the `htonl()` ? Trunnel won't magically switch it :S... Spec specifies:

```
The timestamp is a big-endian unsigned integer number of seconds since the Unix epoch.
```

src/core/or/connection_or.c
```diff
@@ -2448,49 +2474,65 @@ connection_or_send_netinfo,(or_connection_t *conn))
...
-    *out = 0;
+  }
+
+  const char *errmsg = NULL;
+  if ((errmsg = netinfo_cell_check(netinfo_cell))) {
+    log_warn(LD_OR, "Failed to validate NETINFO cell with error: %s",
+                    errmsg);
+    r = -1;
+    goto cleanup;
+  }
```
dgoulet-tor: I think it is OK to remove this check because the encoding function does call `netinfo_cell_check()` and `-1` is returned if the check fails. `-2` is the allocation failed.

src/core/or/connection_or.c
```diff
@@ -2434,8 +2461,7 @@ connection_or_send_netinfo,(or_connection_t *conn))
   cell_t cell;
   time_t now = time(NULL);
   const routerinfo_t *me;
-  int len;
-  uint8_t *out;
+  int r = 0;
```
dgoulet-tor: A trick that I like is set this to `-1` so then every `goto` will have the right error value set. Then if you reach the `cleanup:` label, just before you set `r = 0`. This way, any code addition later, we avoid the error of someone forgetting to set `r` within the error path.

src/trunnel/netinfo.trunnel
rl1987: The old code did not use any address type values other 4 and 6. I suppose they cannot realistically appear in a valid NETINFO cell during handshake?

Added a comment in 9cd116b21323c432e528de8eb02ece617290ee79.

src/core/or/channeltls.c
rl1987: Yes. Did so in b0a8cf0a90a00615bbaf12da4e29d2828c1005b7.

src/core/or/channeltls.c
rl1987: Trunnel does change between endianness automatically - see `netinfo_addr_encode()` and `netinfo_addr_parse_into()`.

Simplified the code to rely on `my_apparent_addr` in 92ada4d8951c4cb4d4d0242ae987c1660f54690d.

src/core/or/channeltls.c
rl1987: Fixed in 1ae0ab87f66c3e322d8a6a8ca9059b36f1ccc4de.

src/core/or/connection_or.c
rl1987: Removed the comments in 216880fe367760120f155974730a6366e8548392 and used the the helper function in 6867bea916a4c71454e3ae5fc6804c50f04db2e5.

src/core/or/connection_or.c
rl1987: Did so in e9b6b8a15aa483de0c4308bef27602f2aeba2f66.

src/core/or/connection_or.c
rl1987: Yes, as trunnel changes the endianness for us:

```
597   /* Encode u32 timestamp */
598   trunnel_assert(written <= avail);
599   if (avail - written < 4)
600     goto truncated;
601   trunnel_set_uint32(ptr, trunnel_htonl(obj->timestamp));
602   written += 4; ptr += 4;
```

src/core/or/connection_or.c
rl1987: But we do lose the error message if we rely on `_encode()` calling `_check()`. 

While calling `_check()` in here is not strictly necessary, it may prove useful in case struct gets populated incorrectly. I would say let's leave it here just to be safe and make debugging easier.

src/trunnel/netinfo.trunnel
dgoulet-tor: Code doesn't use it but doesn't mean the protocol doesn't allow it. 

This specific "netinfo addr" structure actually should be generalized (and probably put outside the netinfo context) because it is also a structure that is used in the `RELAY_RESOLVED` cell so all type need to be there.

It just happens that the NETINFO cell will only care about two of those values.

src/core/or/connection_or.c
dgoulet-tor: Great!

src/core/or/connection_or.c
dgoulet-tor: Ok.

PR #376, opened 2018-09-27
Split stats

nmathewson:


src/feature/stats/geoip.c
```diff
@@ -660,10 +691,7 @@ geoip_note_client_seen(geoip_client_action_t action,
     int country_idx = geoip_get_country_by_addr(addr);
     if (country_idx < 0)
       country_idx = 0; /** unresolved requests are stored at index 0. */
-    if (country_idx >= 0 && country_idx < smartlist_len(geoip_countries)) {
-      geoip_country_t *country = smartlist_get(geoip_countries, country_idx);
-      ++country->n_v3_ns_requests;
-    }
+    increment_v3_ns_request(country_idx);
```
dgoulet-tor: So hmmm `country_idx` is a integer but this function expects a `country_t` which is a typedef to `uint16_t`. I do see a overflow issue but of course if 65535 countries would exists so unlikely but still, I think we should fix that.

PR #383, opened 2018-10-02
#27928: Refrain from listener rebinding when address families differ

rl1987:
https://trac.torproject.org/projects/tor/ticket/27928

### DIFF
--- a/src/core/mainloop/connection.c
+++ b/src/core/mainloop/connection.c
@@ -2792,6 +2792,15 @@ retry_listener_ports(smartlist_t *old_conns,
           break;
         }
 #ifdef ENABLE_LISTENER_REBIND
+        /* Rebinding may be needed if all of the following are true:
+         * 1) Address family is the same in old and new listeners.
+         * 2) Port number matches exactly (numeric value is the same).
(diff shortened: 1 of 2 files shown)


src/core/mainloop/connection.c
```diff
@@ -2793,6 +2793,7 @@ retry_listener_ports(smartlist_t *old_conns,
         }
 #ifdef ENABLE_LISTENER_REBIND
         const int may_need_rebind =
+          tor_addr_family(&wanted->addr) == tor_addr_family(&conn->addr) &&
```
asn-d6: On the trac ticket you said: ` we check if port matches exactly and *one* of the addresses is a wildcard address:`

Where is the wildcard check happening in this code?
Also, can you please document the previous `may_need_rebind` check and also add the new check?

rl1987: The part that checks for one wildcard address is `bool_neq(tor_addr_is_null(&wanted->addr), tor_addr_is_null(&conn->addr));`. By wildcard address I mean 0.0.0.0 for IPv4 or :: for IPv6.

Wrote a comment in 92bc882ec9440be504f9ebbb714cf6622e455605.

PR #386, opened 2018-10-06
Bug27968 SIGNAL HALT race condition in test-rebind.py

teor-travel:
See https://trac.torproject.org/projects/tor/ticket/27968

### DIFF
--- a/src/test/test_rebind.py
+++ b/src/test/test_rebind.py
@@ -7,12 +7,15 @@ import os
 import time
 import random
 
+def fail(msg):
+    print('FAIL')
+    sys.exit(msg)
+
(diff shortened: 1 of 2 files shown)


src/test/test_rebind.py
```diff
@@ -88,6 +88,14 @@ def pick_random_port():
...
 
-time.sleep(0.1)
+wait_for_log('exiting cleanly')
 print('OK')
-tor_process.terminate()
+
+try:
+    tor_process.terminate()
+except OSError as e:
+    if e.errno == 3: # No such process
```
tlyu: This should probably use the symbolic name from the `errno` module instead of hard coding a number.

PR #387, opened 2018-10-07
#27913: Try running test-stem on travis

rl1987:
https://trac.torproject.org/projects/tor/ticket/27913

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,8 @@ matrix:
     # - env: HARDENING_OPTIONS=""
     ## We check asciidoc with distcheck, to make sure we remove doc products
     - env: DISTCHECK="yes" ASCIIDOC_OPTIONS=""
+    # We clone our stem repo and run `make test-stem`
+    - env: TEST_STEM="yes"
     ## Check rust online with distcheck, to make sure we remove rust products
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -191,6 +193,7 @@ script:
   - ./configure $CONFIGURE_FLAGS
   ## We run `make check` because that's what https://jenkins.torproject.org does.
   - if [[ "$DISTCHECK" == "" ]]; then make check; fi
+  - if [[ "$TEST_STEM" != "" ]]; then git clone https://git.torproject.org/stem.git/ ; export STEM_SOURCE_DIR=`pwd`/stem; make test-stem; fi
```
tlyu: Could we maybe use the github.com mirror instead? That way we could be more robust against network connectivity problems. (Travis already needs to be able to reach github.com to start the build.)

Also, maybe use `git clone --depth 1` to speed things up? We don't need the full git history of stem for this test, as far as I know.

Needs a changes file.

rl1987: Fixed `git clone` command in 31812e55e8b9b34a01cad449c9d4561054f8db75 and added changes file in 3afe83dac33f788e2bc7f3aee09d64775a8a6593.

PR #399, opened 2018-10-12
#27913: Try running test-stem on Travis (2)

rl1987:
https://trac.torproject.org/projects/tor/ticket/27913

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -195,7 +195,7 @@ install:
   - if [[ "$RUST_OPTIONS" != "" ]]; then rustup --version; fi
   - if [[ "$RUST_OPTIONS" != "" ]]; then rustc --version; fi
   - if [[ "$RUST_OPTIONS" != "" ]]; then cargo --version; fi
-  - if [[ "$TEST_STEM" != "" ]]; then echo "stem.__version__" | ./stem/tor-prompt; fi
+  - if [[ "$TEST_STEM" != "" ]]; then pushd stem; python -c "from stem import stem; print(stem.__version__);"; popd; fi
 
 script:
   - ./autogen.sh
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -200,6 +202,7 @@ script:
   - ./configure $CONFIGURE_FLAGS
   ## We run `make check` because that's what https://jenkins.torproject.org does.
   - if [[ "$DISTCHECK" == "" ]]; then make check; fi
```
tlyu: I just noticed this job takes significantly longer to run, so maybe we could skip the `make check` when we're doing `make test-stem`? We'd probably want to do something like `make all && make test-stem` though to avoid parallelization problems.

onioncount: +1

I wish we could fix the parallelisation problems, but that's a separate issue that deserves a separate ticket. (And they're so hard to debug.)

I suggest make src/app/tor, unless test-stem uses some of the other build products.

rl1987: Did these things and build takes less time now.

.travis.yml
```diff
@@ -199,7 +201,8 @@ script:
   - echo "Configure flags are $CONFIGURE_FLAGS"
   - ./configure $CONFIGURE_FLAGS
   ## We run `make check` because that's what https://jenkins.torproject.org does.
-  - if [[ "$DISTCHECK" == "" ]]; then make check; fi
+  - if [[ "$DISTCHECK" == "" && "$TEST_STEM" == "" ]]; then make check; fi
+  - if [[ "$TEST_STEM" != "" ]]; then git clone --depth 1 https://github.com/torproject/stem.git ; export STEM_SOURCE_DIR=`pwd`/stem; make src/app/tor test-stem; fi
```
teor2345: Technically, git clone stem belongs in the install section.
We also do our environmental variable setup in install.
Please also log the stem version. I think git clone already logs the stem commit.

It seems like distcheck and test-stem are incompatible, because distcheck deletes src/app/tor.
If both options are specified, should we skip stem and log a message?

Nitpick: I'd like to see test stem added as the last item in both sections, but it doesn't really matter.

rl1987: These are addressed in further commits now. Opened a new pull request with rebased branch and cleaned up git history: https://github.com/torproject/tor/pull/423

PR #405, opened 2018-10-15
test: Update approx_time before the test starts

dgoulet-tor:
This way we have the same time source when the IP is created and tested later.

Fixes #27810


nmathewson:
cherry-picked to 0.3.4

PR #410, opened 2018-10-17
Bug26812

asn-d6:


src/feature/rend/rendservice.c
```diff
@@ -546,14 +546,48 @@ copy_service_on_prunning(rend_service_t *dst, rend_service_t *src)
   dst->n_intro_points_wanted = src->n_intro_points_wanted;
 }
 
+/** A SIGHUP occured and service pruning might have happened; some services
+ *  might not be available anymore. Check if any introduction circuits are now
+ *  useless and need to be cleaned up. */
+static void
+clean_intro_circs_after_service_prune(const smartlist_t *surviving_services) {
```
dgoulet-tor: This is a fun `{` location? :P

dgoulet-tor: Also, I think you can remove `service_` from the name, we are in `rendservice.c` so it is imply. I think just `clean_intro_circs_after_pruning()` will be more meaningful. No strong opinion.

src/feature/rend/rendservice.c
```diff
@@ -546,11 +546,26 @@ copy_service_on_prunning(rend_service_t *dst, rend_service_t *src)
   dst->n_intro_points_wanted = src->n_intro_points_wanted;
 }
 
+/** Close the introduction circuits of <b>s</b>. This function does not remove
+ *  the intro points from the service. */
+static void
+close_intro_circs_of_service(const rend_service_t *s)
```
dgoulet-tor: Again, we are in `rendservice.c`, the service is implied so I would go with `close_intro_circuits()`?

src/feature/rend/rendservice.c
asn-d6: Fixed in new branch.

asn-d6: Fixed in new branch.

src/feature/rend/rendservice.c
asn-d6: Fixed in new branch.

PR #417, opened 2018-10-18
Notify systemd of ShutdownWaitLength

Hello71:


src/feature/hibernate/hibernate.c
```diff
@@ -856,6 +856,10 @@ hibernate_begin(hibernate_state_t new_state, time_t now)
                "connections, and will shut down in %d seconds. Interrupt "
                "again to exit now.", options->ShutdownWaitLength);
     shutdown_time = time(NULL) + options->ShutdownWaitLength;
+#ifdef HAVE_SYSTEMD
+    sd_notifyf(0, "EXTEND_TIMEOUT_USEC=%lu",
+            options->ShutdownWaitLength * 1000000UL + 10000000);
```
teor2345: This could be rewritten as:
```suggestion
            options->ShutdownWaitLength * TOR_USEC_PER_SEC + 10 * TOR_USEC_PER_SEC);
```

But I wonder if you actually meant:
```suggestion
            options->ShutdownWaitLength * TOR_USEC_PER_SEC + 1 * TOR_USEC_PER_SEC);
```

Either way, it might be worth adding a comment to explain the extra second(s).

teor2345: Or even:
```suggestion
            (options->ShutdownWaitLength + 1) * TOR_USEC_PER_SEC);
```

teor2345: Yes, I deliberately changed the number from 10 to 1, when I said:

> But I wonder if you actually meant:

src/feature/hibernate/hibernate.c
```diff
@@ -856,6 +856,18 @@ hibernate_begin(hibernate_state_t new_state, time_t now)
...
+    /* tell systemd that we may need more than the default 90 seconds to shut
+     * down so they don't kill us. add some extra time to actually finish
+     * shutting down, otherwise systemd will kill us immediately after the
+     * EXTEND_TIMEOUT_USEC expires.
+     *
+     * 2^31 usec = ~2147 sec = ~35 min. probably nobody will actually set
+     * ShutdownWaitLength to more than that, but use a longer type just in case
+     */
+    sd_notifyf(0, "EXTEND_TIMEOUT_USEC=%" PRIu64,
+            (uint64_t)(options->ShutdownWaitLength + 30) * TOR_USEC_PER_SEC);
```
teor2345: ShutdownWaitLength is an int. To avoid (a very unlikely) overflow on addition, use:
```suggestion
            ((uint64_t)(options->ShutdownWaitLength) + 30) * TOR_USEC_PER_SEC);
```

changes/ticket28113
```diff
@@ -0,0 +1,3 @@
+  o Minor features (relay usability):
+    - Notify systemd of ShutdownWaitLength so it can be set to more than 90
```
teor2345: The Tor default is actually 30s:
https://github.com/torproject/tor/blob/master/contrib/dist/tor.service.in
```suggestion
    - Notify systemd of ShutdownWaitLength so it can be set to more than 30
```

src/feature/hibernate/hibernate.c
teor2345: I am not sure why you are giving tor an extra 30 seconds beyond the shutdown wait length to shut down:
https://www.freedesktop.org/software/systemd/man/systemd.service.html#TimeoutStopSec=

1s or 2s seems like a reasonable time for tor to run its event loop and exit, but 30s might make relay operators think that their device has hung.

Hello71: I made it a very conservative number because I don't know how long tor might actually take, and because the default is 90 seconds. In fact, it might make more sense to use 90, because that's the default, and ShutdownWaitLength is a sort of "added time". Also note that systemd won't wait around forever like an idiot; as long as tor actually exits before this time the shutdown process will continue immediately. The only difference is that if tor actually fails to exit for some reason then systemd will wait for 30 seconds longer. Also, this won't do anything with the default configurations for ShutdownWaitLength and DefaultTimeoutStopSec, because tor will tell systemd to wait for 30+30=60 seconds, but systemd will wait min(60, DefaultTimeoutStopSec=90)=90 seconds.

changes/ticket28113
Hello71: Yes, but the systemd default is 90.

src/feature/hibernate/hibernate.c
Hello71: For future reference, I think you should've opened two separate comments here (I think that's possible...).

Hello71: Oh, I remember why I increased this from my first revision: I wanted to consider the case where tor may in fact take an extra few tens of seconds to clean up, which is not entirely absurd, like if we try to free memory that was actually swapped out because it was used at startup and then never again.

teor2345: > I made it a very conservative number because I don't know how long tor might actually take, and because the default is 90 seconds. In fact, it might make more sense to use 90, because that's the default, and ShutdownWaitLength is a sort of "added time". Also note that systemd won't wait around forever like an idiot; as long as tor actually exits before this time the shutdown process will continue immediately. The only difference is that if tor actually fails to exit for some reason then systemd will wait for 30 seconds longer. Also, this won't do anything with the default configurations for ShutdownWaitLength and DefaultTimeoutStopSec, because tor will tell systemd to wait for 30+30=60 seconds, but systemd will wait min(60, DefaultTimeoutStopSec=90)=90 seconds.

I understand now. Please add a comment that explains the systemd default of 90 seconds.

PR #418, opened 2018-10-19
ISP

marcus-campbell-codes:
Hides your internet traffic  from your ISP


teor2345:
Thanks for this pull request!

Tor doesn't just hide your traffic from your ISP, it also hides it from the remote site, and all the routers in-between.

Do you think there's a better way to say that?

PR #422, opened 2018-10-19
Memoize summarize_protover_flags()

nmathewson:
Our tests showed that this function is responsible for a huge number
of our malloc/free() calls.  It's a prime candidate for being
memoized.

Closes ticket 27225.


nmathewson:
squashed and merged to master

PR #424, opened 2018-10-19
Please

lilmohawk:
Made it so the directions are bold so people can see it better


teor2345:
Hi, thanks for this pull request.

Tor's README isn't in markdown, so adding `**` won't make the text bold.
It might even make it harder to read.

What do you think?

PR #425, opened 2018-10-20
#21900: Fallback to local DNS when no other nameservers are known (2)

rl1987:
https://trac.torproject.org/projects/tor/ticket/21900

### DIFF
--- a/src/feature/relay/dns.c
+++ b/src/feature/relay/dns.c
@@ -1388,16 +1388,23 @@ configure_nameservers(int force)
   evdns_set_log_fn(evdns_log_cb);
   if (conf_fname) {
     log_debug(LD_FS, "stat()ing %s", conf_fname);
-    if (stat(sandbox_intern_string(conf_fname), &st)) {
+    int missing_resolv_conf = 0;
+    int stat_res = stat(sandbox_intern_string(conf_fname), &st);
+
+    if (stat_res) {
       log_warn(LD_EXIT, "Unable to stat resolver configuration in '%s': %s",
                conf_fname, strerror(errno));
-      goto err;
-    }
-    if (!force && resolv_conf_fname && !strcmp(conf_fname,resolv_conf_fname)
+      missing_resolv_conf = 1;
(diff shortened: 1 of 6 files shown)


src/feature/relay/dns.c
```diff
@@ -1357,6 +1357,36 @@ evdns_err_is_transient(int err)
...
+configured_nameserver_address(const size_t idx)
+{
+ struct sockaddr_storage sa;
+ ev_socklen_t sa_len = sizeof(sa);
+
+ if (evdns_base_get_nameserver_addr(the_evdns_base, (int)idx,
+                                    (struct sockaddr *)&sa,
+                                    sa_len) > 0) {
+   tor_addr_t *tor_addr = tor_malloc(sizeof(tor_addr_t));
+   tor_addr_from_sockaddr(tor_addr, (const struct sockaddr *)&sa, NULL);
```
nmathewson: This call is allowed to fail; we need to check its return value.

src/feature/relay/dns.c
```diff
@@ -1357,6 +1357,36 @@ evdns_err_is_transient(int err)
...
+ * Return address of configured nameserver in <b>the_evdns_base</b>
+ * at index <b>idx</b>.
+ */
+tor_addr_t *
+configured_nameserver_address(const size_t idx)
+{
+ struct sockaddr_storage sa;
+ ev_socklen_t sa_len = sizeof(sa);
+
+ if (evdns_base_get_nameserver_addr(the_evdns_base, (int)idx,
```
nmathewson: This function wasn't added to Libevent until 2.1.6-beta; We only require 2.0.x, I think.  If you want to use it, we need to check for its presence in autoconf, and only call it when it's available.

nmathewson: Oh wait, I see that you added a check later.  Great.

src/feature/relay/dns.c
rl1987: Fixed in b7edfcbf6bb3a27d914ad883ae75413a4d25739a.

PR #435, opened 2018-10-23
[Tor Trac #28127]: Hidden service option HiddenServiceAuthorizeClient is incompatible with version 3

neelchauhan:
For the bug [here](https://trac.torproject.org/projects/tor/ticket/28127).

### DIFF
--- a/src/feature/hs/hs_config.c
+++ b/src/feature/hs/hs_config.c
@@ -419,7 +419,7 @@ config_generic_service(const config_line_t *line_,
           dup_opt_seen = line->key;
         goto err;
       }
-      have_version = 1;
+      have_version = service->config.hs_version_explicitly_set = 1;
       continue;
     }
     /* Virtual port. */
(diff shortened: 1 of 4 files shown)


src/feature/hs/hs_config.c
```diff
@@ -556,11 +553,18 @@ config_service(const config_line_t *line, const or_options_t *options,
...
+
   /* We'll try to learn the service version here by loading the key(s) if
-   * present. Depending on the key format, we can figure out the service
-   * version. If we can't find a key, the configuration version will be used
-   * which has been set previously. */
-  service->config.version = config_learn_service_version(service);
+   * present and we did not set HiddenServiceVersion. Depending on the key
+   * format, we can figure out the service version. */
+  if (!service->config.hs_version_explicitly_set)
+    service->config.version = config_learn_service_version(service);
```
dgoulet-tor: We try to always put the `{}` even if this pattern is everywhere in the code base.

src/feature/hs/hs_config.c
```diff
@@ -577,16 +581,20 @@ config_service(const config_line_t *line, const or_options_t *options,
     tor_assert_nonfatal_unreached();
     goto err;
   }
+
```
dgoulet-tor: Lets not remove this whitespace. The reason is that the `if(ret < 0)` being glued to the `switch` condition is to reflect that they are linked together and that adding anything in between is probably a bad idea.

src/feature/hs/hs_service.h
```diff
@@ -220,6 +220,9 @@ typedef struct hs_service_config_t {
   /* Is this service ephemeral? */
   unsigned int is_ephemeral : 1;
 
+  /* Have we explicitly set HiddenServiceVersion? */
+  unsigned int hs_version_explicitly_set : 1;
+
```
dgoulet-tor: Super nitpick here: Can you put that next to the `version` so we have the parameters from the same context next to each other?

PR #439, opened 2018-10-23
Adaptive padding rebased 0.3.6

mikeperry-tor:
Review branch for asn. Strongly recommend reviewing in commit order.

WARNING: Github is being lame. The below order seems to be timestamp order, not git commit order :/

Actual git commit order is the best way to review.


src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2251 @@
...
+       * param1 is 'p' (success probability) */
+      return ceil(log1p(-p)/log1p(-dist.param1));
+    case CIRCPAD_DIST_WEIBULL:
+      p = crypto_rand_double();
+      /* https://en.wikipedia.org/wiki/Weibull_distribution \
+       *    #Cumulative_distribution_function
+       * param1 is k, param2 is Lambda */
+      return dist.param2*pow(-log1p(p), 1.0/dist.param1);
+    case CIRCPAD_DIST_PARETO:
+      p = 1.0-crypto_rand_double(); // Pareto quantile needs (0,1]
```
teor2345: Let's make crypto_rand_double() take flags for including the endpoints. (Or let's make a new function that takes those flags, and implement crypto_rand_double() using that function.)

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2251 @@
...
+       *
+       * for any 0 < eps < 2.  Hence, as long as |xi| < eps/-2log U,
+       * f(xi) = -log U (1 + E(xi)) for |E(xi)| <= eps/2.  |E(xi)| is
+       * the relative error of f(xi) from -log U; from this bound, the
+       * relative error of -log U from f(xi) is at most (eps/2)/(1 -
+       * eps/2) = eps/2 + (eps/2)^2 + (eps/2)^3 + ... < eps for 0 < eps
+       * < 1.  Since -log U < 1000 for all U in (0, 1] in binary64
+       * floating-point, we can safely cut xi off at 1e-20 < eps/4000
+       * and attain <1ulp error.
+       */
```
teor2345: Ideally, all our floating point code should have a detailed error analysis like this.
In practice, we should document why we chose particular code, but I don't think we need to calculate ulps.
(How accurate do these distributions need to be?)

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2251 @@
...
+      if (fabs(dist.param2) <= 1e-20)
+        return -dist.param1*tor_mathlog(p);
```
teor2345: Here are my remaining questions:
* what are typical values for param1 and param2? What are the min and max values?
* is covering the full range of outputs important? What range of outputs do we need?
* is the discontinuity at the changeover point significant? Do we need to give advice about setting param2 to just above or just below 1e-20?
* can we get infinite floating point results from either part of this function? will infinite floating point results cause errors?

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,534 @@
...
+  CIRCPAD_EVENT_LENGTH_COUNT = 1<<6
+} circpad_event_t;
+
+/** Boolean type that says if we decided to transition states or not */
+typedef enum {
+  CIRCPAD_STATE_UNCHANGED = 0,
+  CIRCPAD_STATE_CHANGED = 1
+} circpad_decision_t;
+
+/** The type for histogram bins */
```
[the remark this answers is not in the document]
teor2345: Should there be a type for histogram bin indexes?

teor2345:
GitHub doesn't like the number of comments I added in my review: some of them will be hidden.

src/core/or/circuitbuild.c
```diff
@@ -2580,7 +2584,24 @@ choose_good_middle_server(uint8_t purpose,
     return choice;
   }
```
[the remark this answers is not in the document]
mikeperry-tor: I think the sanest way for them to interact is by having the vanguards addon inspect MiddleNodes, and only choose HSLayer2Nodes from that set. Otherwise, the manpage should say that researchers must manually set both values to the same nodes if they want to test padding on vanguards-like circuits without using the addon.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2251 @@
...
+    return CIRCPAD_DELAY_INFINITE;
+
+  if (bin == 0)
+    return start_usec;
+
+  if (bin == 1)
+    return start_usec+1;
+
+  return start_usec
+      + (state->range_sec*USEC_PER_SEC)/(1<<(state->histogram_len-bin+1));
```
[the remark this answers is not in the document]
mikeperry-tor: Hrmm. A bunch of 0 sized bins.

We should probably have a circpad_machine_validate() function that checks for obvious signs of machine degeneracy like this while initializing machines. We might be able to wait for the serialization code to be written before we do that, since we only have one hardcoded machine right now that is presumed to be valid.

src/core/or/circuitpadding.c
mikeperry-tor: Hrmm.. I am kinda against a bunch of flags if this is the only case we need it for.

I could easily write up a crypto_rand_double64() though, and use that here.

src/lib/smartlist_core/smartlist_foreach.h
```diff
@@ -83,6 +83,14 @@
          ++var ## _sl_idx) {                                    \
       var = (sl)->list[var ## _sl_idx];
 
+#define SMARTLIST_FOREACH_REVERSE_BEGIN(sl, type, var)  \
```
asn-d6: We should document this new macro and maybe unittest it.

Furthermore, we should document why we are using it in the code instead of the normal SMARTLIST_FOREACH.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2221 @@
...
+  circuit_t *circ = TO_CIRCUIT(on_circ);
+
+#ifdef TOR_UNIT_TESTS
+  /* Tests don't have to init our padding machines */
+  if (!origin_padding_machines)
+    return;
+#endif
+
+  for (int i = 0; i < CIRCPAD_MAX_MACHINES; i++) {
+    if (circ->padding_info[i])
```
asn-d6: IIUC this is the check that makes us not negotiate padding with every hop? We should document this important check.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2221 @@
...
+      continue;
+
+    /* We have a free index */
```
asn-d6: What is a free index?

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2221 @@
...
+    SMARTLIST_FOREACH_REVERSE_BEGIN(origin_padding_machines,
```
asn-d6: Why do we go reverse here?

In general, I count more than 10 for loops that loop over machines and another 5 smartlist lo ops that loop over machines. Some of those loops are nested loops...

Can we introduce some functions that abstract those loops into something more convenient and easier to understand? For example, we could have a function that  gets us the circ->padding_machine or the circpad_machine_t from the machine  type.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2221 @@
...
+                                    circpad_machine_t *,
+                                    machine) {
+      if (machine->machine_index == i &&
```
asn-d6: Relevant to the comment above, what's the index check here? Can we replace that with a pointer to the global machine or a function?

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2221 @@
...
+circpad_negotiate_padding(origin_circuit_t *circ,
+                          circpad_machine_num_t machine,
+                          int target_hopnum,
+                          int command)
+{
+  circpad_negotiate_t type;
+  cell_t cell;
+  ssize_t len;
+
+  if (!circpad_circuit_supports_padding(circ, tar
```
asn-d6: This seems to be an error path. Let's document it as such and also explain the scenarios that can make us end up here.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2221 @@
...
+ * This means that technically the last bin (histogram_len-2)
+ * has range [start_usec+range_sec*USEC_PER_SEC,
+ * CIRCPAD_DELAY_INFINITE].
+ */
+STATIC int
+circpad_histogram_usec_to_bin(circpad_machineinfo_t *mi, circpad_delay_t us)
+{
+  const circpad_state_t *state = circpad_machine_current_state(mi);
+  circpad_delay_t start_usec;
+  int bin;
```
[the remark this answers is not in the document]
asn-d6: I patched it with:
```
+  if (!state) {
+    return CIRCPAD_DELAY_INFINITE;
+  }
```
just to be able to proceed.

src/core/or/circuitpadding.c
mikeperry-tor: I don't actually know the typical values here. It is supported in case someone needs it, because it was easy to do. Same for Weibull..

For the changeover point, in the limit as Xi->0, these functions are the same. So it's just a matter of trying to pick which one to use when close to 0. That's the analysis Riastradh did in the comment.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2251 @@
...
+#include "feature/nodelist/nodelist.h"
+
+#include "app/config/config.h"
+
+HANDLE_IMPL(circpad_machineinfo, circpad_machineinfo_t,)
+#define circpad_machineinfo_handle_free(h)    \
+   FREE_AND_NULL(circpad_machineinfo_handle_t, \
+                 circpad_machineinfo_handle_free_, (h))
+
+#define USEC_PER_SEC (1000000)
```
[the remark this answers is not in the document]
mikeperry-tor: TOR_USEC_PER_SEC is only defined in C files and test files. I can name my thing to be the same. Or I can move the tvdiff.c definitions out of their C and test files and into their header, and use that.

src/lib/smartlist_core/smartlist_foreach.h
mikeperry-tor: I did the comment in eb11886fccada2b56d87cf85fda5fb553298bfbf.

I added a note for the unittest in the branch's TODO file, along with notes for a pile of other tests.

src/core/or/circuitpadding.c
mikeperry-tor: Also commented in eb11886fccada2b56d87cf85fda5fb553298bfbf

src/core/or/circuitpadding.c
mikeperry-tor: Commented in eb11886fccada2b56d87cf85fda5fb553298bfbf.

src/core/or/circuitpadding.c
mikeperry-tor: We go in reverse so that when we add additional machine definitions in the consensus or torrc, those take precedence over older ones, and over any hardcoded machines.

Most of the loops are just over two padding machines indexes. I am not sure how to simplify them further. Though your comment did remind me that we need to check each index each time: 9d7e119aabcf1680df8aa9126b5ca9582176e999

This one nested loop is more efficient after we already have machines on the circuit than doing it the other way around.

src/core/or/circuitpadding.c
mikeperry-tor: Commented in eb11886fccada2b56d87cf85fda5fb553298bfbf -- machine indexes need to be specified explicitly so that when there are two machines on the circuit, events arrive at them in a predictable deterministic order that they are designed to expect -- because they specify which index they go into.

src/core/or/circuitpadding.c
mikeperry-tor: Also documented in eb11886fccada2b56d87cf85fda5fb553298bfbf

src/core/or/circuitpadding.c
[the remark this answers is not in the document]
mikeperry-tor: Huh. This is checked explicitly by the caller (circpad_machine_remove_token()). Unless maybe the unit tests don't check it, and that's where the error came from?

Anyway I added a check in d00ff6b78ef98dcda8380251743e4c63c5816ece

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2221 @@
...
+
+  log_fn(LOG_INFO, LD_CIRC, "Empty routerstatus in padding check");
+  return 0;
+}
+
+/**
+ * Get a node_t for the nth hop in our circuit, starting from 1.
+ */
+static const node_t *
+circuit_get_nth_hop(origin_circuit_t *circ, int hop)
```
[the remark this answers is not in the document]
mikeperry-tor: 65a584d66eae8b351587bc2a7a39d802509d678e

src/app/config/config.c
```diff
@@ -413,6 +413,10 @@ static config_var_t option_vars_[] = {
   V(ExcludeExitNodes,            ROUTERSET, NULL),
   OBSOLETE("ExcludeSingleHopRelays"),
   V(ExitNodes,                   ROUTERSET, NULL),
+  /* Researchers need a way to tell their clients to use specific
+   * middles that they also control, to allow safe live-network
+   * experimentation with new padding machines. */
+  V(MiddleNodes,                 ROUTERSET, NULL),
```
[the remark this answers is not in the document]
mikeperry-tor: Ok I put this in the TODO file. I will probably not get to it until next week,

src/feature/nodelist/nodelist.c
```diff
@@ -2393,6 +2393,7 @@ compute_frac_paths_available(const networkstatus_t *consensus,
   smartlist_free(mid);
   smartlist_free(exits);
 
+  // XXX-MP-AP: Do the same for options->MiddleNodes?
```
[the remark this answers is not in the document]
mikeperry-tor: Ok put this in the TODO too.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2283 @@
...
+    for (int b = 0; b < state->histogram_len; b++)
+      histogram_total += histogram[b];
+  } else if (state->iat_dist.type == CIRCPAD_DIST_NONE) {
+    /* We have a histogram, but it's immutable */
+    histogram = state->histogram;
+    histogram_total = state->histogram_total;
+  } else {
+    /* Sample from a fixed IAT distribution and return */
+    double val = circpad_distribution_sample(state->iat_dist);
+    val = MAX(0, val);
```
teor2345: When comparing int and double, which type gets converted to the other type?

teor2345: Oh, it doesn't matter, because the maximum value is 2**46.

teor2345:
```suggestion
    /* These comparisons are safe, because the output is in the range [0, 2**46]
    val = MAX(0, val);
```

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2251 @@
...
+}
+
+/**
+ * Sample a value from the specified probability distribution.
+ *
+ * This performs inverse transform sampling
+ * (https://en.wikipedia.org/wiki/Inverse_transform_sampling).
+ *
+ * XXX: These formulas were taken verbatim. Need a floating wizard
+ * to check them for catastropic cancellation and other issues (teor?).
```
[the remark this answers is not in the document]
teor2345: I think the code for these distributions belongs in src/lib/math, with the existing laplace.c and laplace.h.

The interface is slightly different: these distributions return double, but the Laplace function returns int64_t.
We might get cleaner code if we make this function, and all the functions it calls, return uint64_t. In particular, we can avoid using double for uniform distributions.

src/core/or/circuitpadding.c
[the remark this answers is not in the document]
teor2345: On a related note: this line is very complex, it contains 14-16 elements: 8 values and 6-8 operators (and 3 sets of brackets). Most humans are good with 3-7 things in their head at once.

For maintainability, please split it into at least 3 lines, using temporary variables. If you make your variables const, most compilers will eliminate the extra operations. (Some will do it Ben if they're not const.)

src/core/or/circuitpadding.c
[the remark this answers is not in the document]
teor2345: I'm not sure, I think @nmathewson knows where time constants belong. Somewhere in lib?

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2283 @@
...
+    start_usec = state->start_usec;
+
+  if (us <= start_usec)
+    return 0;
+
+  if (us == start_usec+1)
+    return 1;
+
+  bin = state->histogram_len -
+    tor_log2((state->range_sec*USEC_PER_SEC)/(us-start_usec+1));
```
teor2345: This line contains 6 values and 6 operations, please split for maintainability.

src/core/or/circuitpadding.c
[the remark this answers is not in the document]
teor2345: Many of these distribution calculations contain more than 7 elements (values and operations), please split for maintainability.

src/core/or/circuitpadding.c
teor2345: I am suggesting we add two flags: allow_zero, and allow_one.
I'll also need these flags when I rewrite the Laplace code.

src/core/or/circuitpadding.c
asn-d6: How about 370f33ff41 as an attempt to abstract away these loops? Hide away implementation details and make it more obvious what they do?

We have a similar pattern in `hs_service.c` (see `FOR_EACH_SERVICE_BEGIN` and `FOR_EACH_DESCRIPTOR_BEGIN`) to reduce the loop details around the codebase, and also to make it less likely we forget checks like the ones from 9d7e119.

src/core/or/circuitpadding.c
asn-d6: Thanks for the comments.

Is it also possible to document `origin_padding_machines` and `relay_padding_machines` and also specify where those come from and how they get initialized?

src/core/or/circuit_st.h
```diff
@@ -177,6 +183,18 @@ struct circuit_t {
   /** Hashtable node: used to look up the circuit by its HS token using the HS
       circuitmap. */
   HT_ENTRY(circuit_t) hs_circuitmap_node;
+
```
asn-d6: Hmm, I see we are adding these three new padding-related elements to the circuit.
Is it possible to document where and when these get initialized and how they interact with each other?

IIUC, padding_machine comes from the consensus. But can they be NULL? What if the consensus does not specify any padding machines?

Also padding_info specifies the current state of the machines? When is it NULL? When the machine is not yet active?

What about padding_handles? Does it just point to padding_info?

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2283 @@
...
+ * Returns 1 if limits are set and we've hit them. Otherwise returns 0.
+ */
+static bool
+circpad_machine_reached_padding_limit(circpad_machineinfo_t *mi)
+{
+  const circpad_machine_t *machine = CIRCPAD_GET_MACHINE(mi);
+  uint8_t max_padding_pct;
+  uint16_t allowed_cells;
+
+  /* If consensus parameters have been set, they override the per-machine
```
pylls: Similar as for being able to specify MiddleNodes, it would be useful to be able to override consensus limits for sake of research.

src/app/config/config.c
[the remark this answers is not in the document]
mikeperry-tor: 74404f7f5aa2a73b96d2fa1e7bf89c23c10090ac

src/feature/nodelist/nodelist.c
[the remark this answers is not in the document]
mikeperry-tor: 40da984ff334206cd8c8a3387cad9949e67bdb25

src/core/or/circuitbuild.c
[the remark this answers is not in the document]
mikeperry-tor: Documented in 74404f7f5aa2a73b96d2fa1e7bf89c23c10090ac

src/core/or/circuitpadding.c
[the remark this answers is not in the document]
mikeperry-tor: Switched name to TOR_USEC_PER_SEC in f4d84dba5a70199cde252c363120924688cc1d34.

I put an XXX there for Nick when he looks at the branch later this week.

src/core/or/circuitpadding.c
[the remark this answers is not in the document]
mikeperry-tor: 5fa69f86ae6c4843599e2ad994fffb814e26aae4

src/core/or/circuitpadding.c
mikeperry-tor: The current implementation of crypto_rand_double() tries to use whatever INT is on the platform as the underlying range for the double. I am saying I could make one that always uses uint64_t.

src/core/or/circuitpadding.c
mikeperry-tor: OK I took this and tweaked it a bit and did f7dc98f14be33873116fba08d1b0a70df5f34fdd

src/core/or/circuitpadding.c
mikeperry-tor: 1208109dc4a20b838c11c827d1a24abfdf253e49

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,534 @@
...
+typedef enum {
+  /** Don't remove any tokens */
+  CIRCPAD_TOKEN_REMOVAL_NONE = 0,
+  /**
+   * Remove from the first non-zero higher bin index when current is zero.
+   * This is the recommended strategy from the Adaptive Padding paper. */
+  CIRCPAD_TOKEN_REMOVAL_HIGHER = 1,
+  /** Remove from the first non-zero lower bin index when current is empty. */
+  CIRCPAD_TOKEN_REMOVAL_LOWER = 2,
+  /** Remove from the closest non-zero lower bin index in either direction
```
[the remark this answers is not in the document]
mikeperry-tor: a257db867892094bba13445e9cc9bf85491f1bef

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,534 @@
...
+   * state machine runs in terms of sent padding cells or all
+   * sent cells. Values are sampled from this distribution, clamped
+   * to max_len, and then start_len is added to that value.
+   *
+   * It may be specified instead of or in addition to
+   * the infinity bins and bins empty conditions. */
+  circpad_distribution_t length_dist;
+
+  /** A minimum length value, added to the output of length_dist */
+  uint16_t start_length;
```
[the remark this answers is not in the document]
mikeperry-tor: This is an offset to the output of length_dist, in cells, so it's the minimum possible value that can be returned after samling that dist. I think 65535 cells is an OK upper bound for the minimum value that can be returned. The dist itself can span to much larger values as needed (up to uint64_t max_length)

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,534 @@
...
+  /** Which hop in the circuit should we send padding to/from?
+   *  1-indexed (ie: hop #1 is guard, #2 middle, #3 exit). */
+  uint8_t target_hopnum : 3;
+
+  /** This machine can only kill fascists if the right conditions are met. */
+  circpad_machine_conditions_t conditions;
+
+  /** How many padding cells can be sent before we apply overhead limits?
+   * XXX: Note that we can only allow up to 64k of padding cells on an
+   * otherwise quiet circuit. Is this enough? It's 33MB. */
```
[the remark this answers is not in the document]
mikeperry-tor: 514*65535/1024/1024 ~= 32MB.

And these are the limits to exempt an otherwise quiet circuit from overhead limits.. I could still see an argument for this one being larger, though. However, doing anywhere near 33MB of padding on quiet circuits means quite a lot of overhead on the network..Though I suppose the histograms could make such a thing sufficiently unlikely, if we wanted some clients to rarely randomly download lots of padding data on a small subset of quiet circuits.

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,534 @@
...
+  uint16_t padding_sent;
+  /** A scaled count of non-padding packets sent, used to limit padding
+   *  overhead. When this reaches UINT16_MAX, we cut it and padding_sent in
+   *  half. */
+  uint16_t nonpadding_sent;
+
+  /**
+   * EWMA estimate of the RTT of the circuit from this hop
+   * to the exit end, in microseconds. */
+  circpad_delay_t rtt_estimate_us;
```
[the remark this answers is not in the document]
mikeperry-tor: In the headers the only time field I can find that doesn't end in _us is start_usec. Is that the one you mean?

src/core/or/protover.c
```diff
@@ -396,7 +400,8 @@ protover_get_supported_protocols(void)
     "LinkAuth=3 "
 #endif
     "Microdesc=1-2 "
-    "Relay=1-2";
+    "Relay=1-2 "
+    "Padding=1";
```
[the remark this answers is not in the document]
mikeperry-tor: I think c2886919c24fa1f0545d498947f88110634de28b does this? I'm not 100% sure tho.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2221 @@
...
+  /* Check that the target hop is opened */
+  if (target_hop->state != CPATH_STATE_OPEN) {
+    log_fn(LOG_WARN,LD_CIRC,
+           "Padding circuit %u has %d hops, not %d",
+           circ->global_identifier,
+           circuit_get_cpath_opened_len(circ), hopnum);
+    return -1;
+  }
+
+  log_fn(LOG_INFO,LD_CIRC, "Negotiating padding on circuit %u.",
```
asn-d6: That's not always true because we use the same function to send DROP cells, so it's not always negotiating padding.

On this note, please check commit f937559 which improves and adds some logs, which I found useful during testing. As I test this further, I might add/remove more logs.

src/core/or/circuitpadding.h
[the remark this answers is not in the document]
mikeperry-tor: Hrmm I thought about this, but I decided to go with just int because I need negative values to mean "bin not found", for the token removal functions.

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,534 @@
...
+
+  // These next three fields are origin machine-only...
+  /** Origin side or relay side */
+  uint8_t origin_side : 1;
+
+  /** Which hop in the circuit should we send padding to/from?
+   *  1-indexed (ie: hop #1 is guard, #2 middle, #3 exit). */
+  uint8_t target_hopnum : 3;
+
+  /** This machine can only kill fascists if the right conditions are met. */
```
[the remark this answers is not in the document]
mikeperry-tor: d56aef8bedea99887ad89418604a3dc883fda8a7

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2251 @@
...
+  /* Truncate the high bin in case it's the infinity bin:
+   * Don't actually schedule an "infinite"-1 delay */
+  bin_end = MIN(bin_end, start_usec+state->range_sec*USEC_PER_SEC*2);
+
+  // Sample uniformly between histogram[i] to histogram[i+1]-1,
+  // but no need to sample if they are the same timeval (aka bin 0).
+  if (bin_end <= bin_start+1)
+    return bin_start;
+  else
+    return (circpad_delay_t)crypto_rand_uint64_range(bin_start, bin_end);
```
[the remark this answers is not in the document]
mikeperry-tor: CIRCPAD_DELAY_INFINITE is UINT32_MAX, so crypto_rand_int_range() is not the full range here. we need uint64 or some signed type conversion or something.. leaving as-is for now.

src/core/or/circuitpadding.c
mikeperry-tor: 5fa69f86ae6c4843599e2ad994fffb814e26aae4

src/core/or/circuit_st.h
mikeperry-tor: 821ced37805e9d2ca1110de9ade8d0bfef3aa5e6 maybe clarifies this a bit?

Also I got rid of the handles. They are not needed (because we always cancel timers when freeing padding_info) and just take up space.

src/core/or/circuitpadding.c
mikeperry-tor: Hrmm I simplified this. I don't think we actually need consensus machine limits, since presumably any machines we accept into the consensus should have their own machine limit.

I changed the consensus limit to a global limit in db6e20fd573f0a4111fdba273fece6cd347bd240 and put a TODO in the source for them not to apply to machines loaded from torrc (which we have not implemented yet).

src/core/or/circuitpadding.c
mikeperry-tor: Ok I took this commit but changed the loud LOG_WARNS back to LOG_INFO. 3fbc3749b64516e759ae95c825692c74d2c8e039

One way to do this sort of thing without extra WARNs is make sure all the logs we're interested in have a circpad prefix we can grep for. Another might be to create a different LD instead of LD_CIRC.

PADDING_TODO.txt
```diff
@@ -0,0 +1,62 @@
...
+  - Fix timer flapping in tests (mock monotime for libevent?)
+  - Test SMARTLIST_FOREACH_REVERSE_BEGIN()
+  - Tests for circuit getting marked or freed before callback.
+  - Tests for detecting padding from wrong hop
+  - Tests for machine replacement on all conditions
+  - Tests for STATE_END teardown due to infinity, bins_empty, and/or length
+    with pending timers, new machines, etc.
+  - Tests for consensus and machine rate limiting
+  - Test all types of token removal
+  - Go down dist precision rabbithole and comment
```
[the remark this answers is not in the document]
mikeperry-tor: Yes, this is actually a risk we can defer on. As per our conversation earlier, changes to *functionality* (like adding new distributions) require bumping the versions of the circuit padding protocol and doing a bunch of version-related checks. But changes to accuracy and correctness of our existing features does not require protocol revision.

So while I would like to get this as accurate as possible for 0.3.6, it is not the end of the world if it is found to be insufficient during experimentation.

mikeperry-tor:
Closing this PR in favor of the squashed one (https://github.com/torproject/tor/pull/461).

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,534 @@
...
+typedef uint32_t circpad_delay_t;
+
+/**
+ * An infinite padding cell delay means don't schedule any padding --
+ * simply wait until a different event triggers a transition.
+ *
+ * This means that the maximum delay we can scedule is UINT32_MAX-1
+ * microseconds, or about 4300 seconds (1.25 hours).
+ * XXX: Is this enough if we want to simulate light, intermittent
+ * activity on an onion service?
```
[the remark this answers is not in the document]
asn-d6: Hm, good point. This might be an issue in the future if we want to do simulate some very light activity.

What would happen if we used `uint64_t` for `circpad_delay_t`? How much more overhead for the structures?

src/core/or/circuitpadding.h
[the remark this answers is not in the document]
teor2345: Consistency in the C files is also important for maintainability.
In the C files, I've also found bin_start, bin_end, and rtt_time.

Can you standardise on "usec" everywhere? Including function names?

PR #445, opened 2018-10-26
#27325: Rework NETINFO wire format handling to rely on trunnel (2)

rl1987:
https://trac.torproject.org/projects/tor/ticket/27325


src/core/or/channeltls.c
```diff
@@ -1720,52 +1748,70 @@ channel_tls_process_netinfo_cell(cell_t *cell, channel_tls_t *chan)
...
+
     tor_addr_t addr;
-    const uint8_t *next =
-      decode_address_from_payload(&addr, cp, (int)(end-cp));
-    if (next == NULL) {
+
+    if (tor_addr_from_netinfo_addr(&addr, netinfo_addr) == -1) {
       log_fn(LOG_PROTOCOL_WARN,  LD_OR,
              "Bad address in netinfo cell; closing connection.");
       connection_or_close_for_error(chan->conn, 0);
```
nmathewson: Hang on, does this mean that we previously rejected unrecognized address types?  That's not good...

src/trunnel/netinfo.trunnel
```diff
@@ -0,0 +1,21 @@
+const NETINFO_ADDR_TYPE_IPV4 = 4;
+const NETINFO_ADDR_TYPE_IPV6 = 6;
+
+struct netinfo_addr {
```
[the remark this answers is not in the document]
rl1987: Fixed in 3bec371d04b291ec0351f01eeafdc842ea2a51d6.

src/core/or/channeltls.c
rl1987: Fixed in 5b2acbec0e50a1858c43d0cafe9c4696152cde27.

src/trunnel/netinfo.trunnel
```diff
@@ -5,14 +5,14 @@ const NETINFO_ADDR_TYPE_IPV4 = 4;
...
 
 struct netinfo_addr {
-  u8 addr_type IN [NETINFO_ADDR_TYPE_IPV4, NETINFO_ADDR_TYPE_IPV6];
-  u8 len IN [4, 16];
+  u8 addr_type;
+  u8 len;
   union addr[addr_type] {
     NETINFO_ADDR_TYPE_IPV4: u32 ipv4;
     NETINFO_ADDR_TYPE_IPV6: u8 ipv6[16];
     default: fail;
```
nmathewson: This is still going to mean that if addr_type isn't recognized, parsing fails.

Try replacing the union line with `union addr[addr_type] with length len {`, and the default line with `default: ignore` -- I think that will do it.

Also, we should have unit tests that make sure that this case actually works.

rl1987: Fixed netinfo.trunnel in c92c0cbc9f7a1ad889cfdda6943ca484d110438e. Added unit test in c659603ac5a428d93ea625aa28b6b51ef9f2f9b3.

PR #461, opened 2018-10-31
Adaptive padding rebased 0.3.6 squashed

mikeperry-tor:
Strong recommend to review in git commit order (which is different than github's weirdly sorted order below).


src/core/or/or.h
```diff
@@ -834,6 +837,10 @@ typedef struct protover_summary_flags_t {
    * service rendezvous point supporting version 3 as seen in proposal 224.
    * This requires HSRend=2. */
   unsigned int supports_v3_rendezvous_point: 1;
+
+  /** True iff this router has a protocol list that allows clients to
+   * negotiate link-level padding. Requires Padding=1. */
```
teor2345: Shouldn't this be "Padding >= 1"?

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+  /** Which machine index slot should this machine go into in
+   *  the array on the circuit_t */
+  uint8_t machine_index : 1;
+
+  /** Send a padding negotiate to shut down machine at end state? */
+  uint8_t negotiate_end : 1;
+
+  // These next three fields are origin machine-only...
+  /** Origin side or relay side */
+  uint8_t origin_side : 1;
```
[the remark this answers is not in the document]
teor2345: > Also, we can turn this into a `bool` instead of a `uint8_t :1`. Same for all the booleans above. I think it might be clearer.

Careful: 3 bools take up 3 bytes, but 3 bitfields take up 3 bits, and those 3 bits will will be packed into the same byte as the 3-bit target_hopnum field.

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+  circpad_time_t last_received_time_us;
+
+  /**
+   * The time at which we scheduled a non-padding packet,
+   * or selected an infinite delay.
+   *
+   * Monotonic time in microseconds since system start.
+   * This is 0 if we haven't chosen a padding delay.
+   */
+  circpad_time_t padding_scheduled_at_us;
```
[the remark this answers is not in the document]
teor2345: I commented on the other pull reequest that we should standardise us to usec.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+ *
+ * TODO: Don't apply limits to machines form torrc.
+ *
+ * Returns 1 if limits are set and we've hit them. Otherwise returns 0.
+ */
+static bool
+circpad_machine_reached_padding_limit(circpad_machineinfo_t *mi)
+{
+  const circpad_machine_t *machine = CIRCPAD_GET_MACHINE(mi);
+
```
asn-d6: This function contains thick code. These two if blocks are quite hard to parse. Let's try to break them down into more steps.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+    return CIRCPAD_STATE_UNCHANGED;
+  }
+
+  if (mi->padding_timer_scheduled) {
+    /* Cancel current timer (if any) */
+    timer_disable(mi->padding_timer);
+    mi->padding_timer_scheduled = 0;
+  }
+
+  /* in_us = in microseconds */
```
asn-d6: Nit: Rename `in_us` to `in_usec`

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+    val = MIN(val, state->range_sec*TOR_USEC_PER_SEC);
+
+    /* This addition is exact: val is at most 2**16 * USEC_PER_SEC ~= 2**46,
+     * start_usec is at most 2**32, and doubles have a precision of 53 bits. */
+    val += start_usec;
+    /* Clamp the distribution at infinite delay val */
+    return (circpad_delay_t)MIN(tor_llround(val), CIRCPAD_DELAY_INFINITE);
+  }
+
+  bin_choice = crypto_rand_int(histogram_total);
```
asn-d6: Suggestion: Let's rename `histogram_total` to `histogram_total_tokens` to make it more explicit.

In this function it's not clear whether `histogram_total` is the number of bins or the number of tokens, and actually the above line strongly implies the former when it's the latter.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+
+  tor_assert(i < state->histogram_len);
+  tor_assert(histogram[i] > 0);
+
+  // Store this index to remove the token upon callback.
+  if (state->token_removal != CIRCPAD_TOKEN_REMOVAL_NONE) {
+    mi->chosen_bin = i;
+  }
+
+  if (i == state->histogram_len-1) {
```
asn-d6: I wonder if we can make it more clear that `histogram_len-1` is the infinity bin, since sometimes it's not clear that this is what we are doing.

There are various places in the code that compute `histogram_len-1` and act on it. Perhaps we can introduce a function called `get_infinity_bin_index()` or something?

src/core/or/circuitpadding.h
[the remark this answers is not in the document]
asn-d6: I don't think there are enough `circpad_machine_t` structs alive to worry about sizes here, but I agree that this might be suboptimal for not much gain. Let's forget turning them to bools.

However, changing the variable names to indicate they are booleans is still worth it IMO.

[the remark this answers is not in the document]
mikeperry-tor: All the variable changing you guys asked for in this and other comments is done across these three commits (so as to squash sanely): d8c4eff8c5d62d172eac0d5096f32f0be99ce6cf 979cc31a3e3e97619760c7aa4f94c8f767cdd29a e18d5a3111375b995a57e809f080011d6ebe5ad5

src/core/or/circuitpadding.h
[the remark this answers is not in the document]
mikeperry-tor: All the variable changing you guys asked for in this and other comments is done across these three commits (so as to squash sanely): d8c4eff8c5d62d172eac0d5096f32f0be99ce6cf 979cc31a3e3e97619760c7aa4f94c8f767cdd29a e18d5a3111375b995a57e809f080011d6ebe5ad5

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2403 @@
...
```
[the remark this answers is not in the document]
mikeperry-tor: All the variable changing you guys asked for in this and other comments is done across these three commits (so as to squash sanely): d8c4eff8c5d62d172eac0d5096f32f0be99ce6cf 979cc31a3e3e97619760c7aa4f94c8f767cdd29a e18d5a3111375b995a57e809f080011d6ebe5ad5

src/core/or/circuitpadding.c
mikeperry-tor: All the variable changing you guys asked for in this and other comments is done across these three commits (so as to squash sanely): d8c4eff8c5d62d172eac0d5096f32f0be99ce6cf 979cc31a3e3e97619760c7aa4f94c8f767cdd29a e18d5a3111375b995a57e809f080011d6ebe5ad5

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2403 @@
...
+      /* If this is not the same state, switch and init tokens,
+       * otherwise just reschedule padding. */
+      if (mi->current_state != s) {
+        mi->current_state = s;
+        circpad_machine_setup_tokens(mi);
+        circpad_choose_state_length(mi);
+
+        /* If we transition to the end state, check to see
+         * if this machine wants to be shut down at end */
+        if (s == CIRCPAD_STATE_END) {
```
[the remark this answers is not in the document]
mikeperry-tor: 0c88a3f4d7b73a4db07a8a37792ecb134c1d7475

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2403 @@
...
+  if (us <= start_usec)
+    return 0;
+
+  if (us == start_usec+1)
+    return 1;
+
+  const circpad_time_t histogram_range_usec =
+                       state->range_sec*TOR_USEC_PER_SEC;
+  bin = state->histogram_len -
+    tor_log2(histogram_range_usec/(us-start_usec+1));
```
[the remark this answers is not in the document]
mikeperry-tor: 0c88a3f4d7b73a4db07a8a37792ecb134c1d7475 and baa5c421cdef7c2f991fc57b81cf0526b705005b (I found some bugs here while trying to explain this in the code and in the spec -- it's still a little wild but the tests are much better and I verified they are doing the right thing). See also ed6284a41b6dca65bd2b19c2f2a066802beb10d2 which makes use of the INFINITY_BIN macro you asked for to make this clearer still.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2403 @@
...
+    /* We have a histogram, but it's immutable */
+    histogram = state->histogram;
+    histogram_total = state->histogram_total;
+  } else {
+    /* Sample from a fixed IAT distribution and return */
+    double val = circpad_distribution_sample(state->iat_dist);
+    /* These comparisons are safe, because the output is in the range
+     * [0, 2**46], and double has a precision of 53 bits. */
+    val = MAX(0, val);
+    val = MIN(val, state->range_sec*TOR_USEC_PER_SEC);
```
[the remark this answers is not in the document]
mikeperry-tor: Space issues in the state machines, basically.. I added this and other variable precision/size stuff to consider to the TODO in 7e25e9e5844141123f59f5a5f161bc0925d93b61 and ff865bb0a23b9c1c06f99e408f493e01400219da

src/core/or/or.h
mikeperry-tor: e7be646f54cbc603ee2a74fe09277cbf8c0bb11e. Tho technically the code as-is checks to see if Padding=1 is present in the protover.. But I think that's right. If we bump padding in a way that is backwards-compatible, we can do Padding=1-2. Otherwise we'll do Padding=2.

(Sadly, if we don't get consensus serialization into 0.3.6, we'll have to do Padding=2 for that, not Padding=1-2, since Padding=1 relays won't read the consensus).

src/core/or/circuitpadding.c
mikeperry-tor: 555e5a2a91ed261f2faee6b8b9827e781649e279 - also cleaned up some potential int overflow issues here.

src/core/or/circuitpadding.c
mikeperry-tor: 5b39c9d3d8ffe75cb5a5fbcc5f9f23c8a2ef1362 and ed6284a41b6dca65bd2b19c2f2a066802beb10d2. Macro is better because it will work with circpad_state_t and circpad_machineinfo_t (the infinity bin checks apply to both of those, and the fields have the same name).

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+  circpad_delay_t start_usec;
+
+  tor_assert(state);
+
+  if (state->use_rtt_estimate)
+    start_usec = mi->rtt_estimate_us+state->start_usec;
+  else
+    start_usec = state->start_usec;
+
+  if (state->token_removal != CIRCPAD_TOKEN_REMOVAL_NONE) {
```
asn-d6: This is a weird if block because its branches are disconnected.
IIUC, it checks whether the histogram is immutable or mutable, and it also checks if we are using a histogram or a distribution in the same if block, and if it's a distribution it samples it on the spot.

Might be nicer to do this cleanly and first check if we are using a distribution, and handle that (ideally in its own function). And then if we are using a histogram, check if it's mutable or immutable. instead of doing everything at the same time.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+  }
+
+  if (mi->state_length != CIRCPAD_STATE_LENGTH_INFINITE) {
+    tor_assert(mi->state_length > 0);
+    mi->state_length--;
+  }
+
+  /* Update padding counts */
+  mi->padding_sent++;
+  if (mi->padding_sent == UINT16_MAX) {
```
asn-d6: Hmm, what's the story with this being just a uint16_t? Why not make it bigger?

Also can you please explain the rationale behind the clamping? You had explained it at another point but I've missed it.  Is this something that should be explained in the spec? Or perhaps a comment in the code?

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+    CIRCPAD_EVENT_PADDING_RECV |
+    CIRCPAD_EVENT_NONPADDING_RECV;
+
+  /* If we are in burst state, and we send a non-padding cell, then we cancel
+     the timer for the next padding cell:
+     We dont want to send fake extends when actual extends are going on */
+  circ_client_machine->burst.transition_cancel_events =
+    CIRCPAD_EVENT_NONPADDING_SENT;
+
+  circ_client_machine->burst.transition_events[CIR
```
asn-d6: I ran this branch on the normal network as an HSv3 and I noticed that I will send padding cells (and my machines will transition states) even tho no middle relays actually support padding. Is this something normal and expected?

Maybe this behavior should be documented in the spec? Right now section `3.2. State Machine Selection` implies that it should not send padding.

mikeperry-tor: Yikes yeah it should not be sending padding at all on the live network. This means two failures: 1) the protover check returned an incorrect result, 2) and the machine did not wait for the padding negotiated response to start padding (though not waiting for the response is maybe sometimes desirable, but I didn't try to code it that way).

asn-d6: WRT 1) IIRC protover returned that padding is not supported (as it was supposed to). But perhaps that doesnt matter in terms of setting up the machine and starting it up on one side.

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+ * These constants form a bitfield to specify the types of events
+ * that can cause transitions between state machine states.
+ *
+ * Note that SENT and RECV are relative to this endpoint. For
+ * relays, SENT means packets destined towards the client and
+ * RECV means packets destined towards the relay. On the client,
+ * SENT means packets destined towards the relay, where as RECV
+ * means packets destined towards the client.
+ */
+typedef enum {
```
nmathewson: I'm a little confused why these are bits; it seems to me that a short array where you do something like `nextstate[eventno]` would be easier... but I haven't read the rest yet

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+ * Each circuit has up to two state machines, and each state
+ * machine consists of these states. Machines transition between
+ * these states using the event transition specifiers below.
+ */
+typedef enum {
+  CIRCPAD_STATE_START = 0,
+  CIRCPAD_STATE_BURST = 1,
+  CIRCPAD_STATE_GAP = 2,
+  CIRCPAD_STATE_END = 3
+} circpad_statenum_t;
```
nmathewson: I don't understand why we believe that 4 states is the right number.  Like, how sure are we there won't be more later?

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+ * RECV means packets destined towards the relay. On the client,
+ * SENT means packets destined towards the relay, where as RECV
+ * means packets destined towards the client.
+ */
+typedef enum {
+  CIRCPAD_EVENT_NONPADDING_RECV = 1<<0,
+  CIRCPAD_EVENT_NONPADDING_SENT = 1<<1,
+  CIRCPAD_EVENT_PADDING_SENT = 1<<2,
+  CIRCPAD_EVENT_PADDING_RECV = 1<<3,
+  CIRCPAD_EVENT_INFINITY = 1<<4,
```
nmathewson: Infinity of what?  I think we should have documentation for each of these enum values.

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+  CIRCPAD_EVENT_BINS_EMPTY = 1<<5,
+  CIRCPAD_EVENT_LENGTH_COUNT = 1<<6
+} circpad_event_t;
+
+/** Boolean type that says if we decided to transition states or not */
```
nmathewson: Would "bool" be good enough here?

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+#include "lib/evloop/timers.h"
+
+typedef struct circuit_t circuit_t;
+typedef struct origin_circuit_t origin_circuit_t;
+typedef struct cell_t cell_t;
+
+/**
+ * Circpad state specifier.
+ *
+ * Each circuit has up to two state machines, and each state
```
nmathewson: I had thought that it had up to one state machine for each hop?

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+ * flags into this more compact representation.
+ */
+typedef enum {
+  CIRCPAD_CIRC_BUILDING = 1<<0,
+  CIRCPAD_CIRC_OPENED = 1<<1,
+  CIRCPAD_CIRC_NO_STREAMS = 1<<2,
+  CIRCPAD_CIRC_STREAMS = 1<<3,
+  CIRCPAD_CIRC_HAS_RELAY_EARLY = 1<<4,
+  CIRCPAD_CIRC_HAS_NO_RELAY_EARLY = 1<<5
+} circpad_circuit_state_t;
```
nmathewson: What does it mean if neither "STREAMS" nor "NO_STREAMS" is set? Does it mean that a state machine can never be applied? Or is this not allowed? Or something else?

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+
+/**
+ * This type specifies all of the conditions that must be met before
+ * a client decides to initiate padding on a circuit.
+ *
+ * A circuit must satisfy every sub-field in this type in order
+ * to be considered to match the conditions.
+ */
+typedef struct circpad_machine_conditions_t {
+  /** Only apply the machine *if* the circuit has at least this many hops */
```
nmathewson: Can we use a full uint8_t here? If not, let's make it "unsigned".

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+  uint8_t min_hops : 3;
+
+  /** Only apply the machine *if* vanguards are enabled */
```
nmathewson: bool?

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+  /** Remove from the closest bin by time value (since bins are
+   *  exponentially spaced). */
+  CIRCPAD_TOKEN_REMOVAL_CLOSEST_USEC = 4,
+  /** Only remove from the exact bin corresponding to this delay. If
+   *  the bin is 0, simply do nothing. Don't pick another bin. */
+  CIRCPAD_TOKEN_REMOVAL_EXACT = 5
+} circpad_removal_t;
+
+/** The maximum length any histogram can be. */
+/** Each bin is twice as large as the previous, with an extra bin for 0 */
```
nmathewson: s/large/wide/ ? Maybe there should  be an explanation around here of what a histogram is and what it means?

Also, is "0" the minimum value or the infinity value? Because 0 isn't an allowed option in prop254 as I understand it?

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+#define CIRCPAD_MAX_HISTOGRAM_LEN (sizeof(circpad_delay_t)*8 + 1)
+
+/**
+ * Distribution types supported by circpad_distribution_sample().
+ *
+ * These can be used instead of histograms for the inter-packet
+ * timing distribution, or to specify a distribution on the number
+ * of cells that can be sent while in a specific state of the state
+ * machine. */
```
nmathewson: These aren't in prop254; where are they specified?

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+ * A circuit padding state machine state.
+ *
+ * This struct describes the histograms and parameters of a single
+ * state in the adaptive padding machine. Instances of this struct
+ * exist in global circpad machine definitions that come from torrc
+ * or the consensus, and are immutable.
+ */
+typedef struct circpad_state_t {
+  /** how long the histogram is (in bins). Histograms must have at least 2
+   *  bins (or 0, if iat distributions are in use). */
```
nmathewson: "IAT"?

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+  uint8_t histogram_len;
+  /** histogram itself: an array of uint16s of tokens, whose
```
nmathewson: Is this constant or are these bins that get emptied (and refilled?)

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+   * widths are exponentially spaced, in microseconds */
+  circpad_hist_bin_t histogram[CIRCPAD_MAX_HISTOGRAM_LEN];
+  /** total number of tokens */
```
nmathewson: Is this the sum of all the bins in "histogram"? If so let's document that

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+  uint32_t histogram_total;
+
+  /** Microseconds of the first bin of histogram, or base of iat dist */
```
nmathewson: This comment doesn't make sense to me.

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+  circpad_delay_t start_usec;
+  /** The span of the histogram in seconds, used to calculate bin with.
+   *  For iat dist use, this is used as a max delay cap on the distribution. */
+  uint16_t range_sec;
+
+  /**
+   * The iat_dist is a parametrized way of encoding inter-packet delay
```
nmathewson: This comment never explains what iat_dist actually means

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+   * the infinity bins and bins empty conditions. */
+  circpad_distribution_t length_dist;
+
+  /** A minimum length value, added to the output of length_dist */
+  uint16_t start_length;
+
+  /** A cap on the length value that can be sampled from the length_dist */
+  uint64_t max_length;
+
+  /** Should we decrement length when we see a nonpadding packet?
```
nmathewson: Let's use "unsigned" for bitfields that aren't uint8/16/32/64_t

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+   */
+  uint8_t use_rtt_estimate : 1;
+
+  /** This specifies the token removal strategy to use upon padding and
+   *  non-padding activity. */
+  circpad_removal_t token_removal;
+} circpad_state_t;
+
+/**
+ * Mutable padding machine info.
```
nmathewson: "info" implies that we haven't thought this through completely.  Do we mean "state"?  Should the state type vary from machine to machine?

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+/**
+ * This specifies a particular padding machine to use after negotiation.
+ *
+ * The constants for machine_num_t are in trunnel.
+ * We want to be able to define extra numbers in the consensus/torrc, though.
+ */
+typedef uint8_t circpad_machine_num_t;
+
+/** Global state machine structure from the consensus */
+typedef struct circpad_machine_t {
```
nmathewson: So this is, like, the specification for a machine?

src/core/or/circuit_st.h
```diff
@@ -177,6 +182,19 @@ struct circuit_t {
   /** Hashtable node: used to look up the circuit by its HS token using the HS
       circuitmap. */
   HT_ENTRY(circuit_t) hs_circuitmap_node;
+
+  /** Adaptive Padding state machines: these are immutable. The state machines
+   *  that come from the consensus are saved to a global structure, to avoid
+   *  per-circuit allocations. This merely points to the global copy. */
+  const circpad_machine_t *padding_machine[CIRCPAD_MAX_MACHINES];
```
nmathewson: What do the indices of this array mean?

src/core/or/circuit_st.h
```diff
@@ -177,6 +182,19 @@ struct circuit_t {
...
+
+  /** Adaptive Padding machine info for above machines. This is the
+   *  per-circuit mutable information, such as the current state and
+   *  histogram token counts. Some of it is optional (aka NULL).
+   *  If a machine is being shut down, these indexes can be NULL
+   *  without the corresponding padding_machine being NULL, while we
+   *  wait for the other end to respond to our shutdown request. */
+  circpad_machineinfo_t *padding_info[CIRCPAD_MAX_MACHINES];
```
nmathewson: As above; what do the indices mean?

src/core/or/protover.c
```diff
@@ -49,6 +52,7 @@ static const struct {
   { PRT_HSREND, "HSRend" },
   { PRT_DESC, "Desc" },
   { PRT_MICRODESC, "Microdesc"},
+  { PRT_PADDING, "Padding"},
```
nmathewson: Is this in the spec?  This needs to be in the spec. These are crucial.

src/core/or/circuitpadding.h
mikeperry-tor: We do the inverse, actually. These are bits because there can be many more ways to get from one state to the other than there are states. So bitfield is a more compact way to represent these many transition arcs.

So instead of nextstate[event1] = next_state; nextstate[event2] = next_state; nextstate[event3] = next_state; we have transiton_events[next_state] |= event1|event2|event3.

src/core/or/circuitpadding.h
mikeperry-tor: Since 2006 nobody has needed more than 2 states and 2 machines, for at least 4 different applications of this research concept and several independent implementations. However, it may be the case that people will invent more creative ways to use this system than have been studied previously, since it is now a real thing that can be used relatively easily.

Do you think we should make the number of states fully arbitrary? I could see a remote possibility of someone building up a far more complex HMM-style machine using way more states. I am just not sure if it is worth the complexity and potential clarity loss.

src/core/or/circuitpadding.h
mikeperry-tor: I wanted the enum because the set of functions that need to pass this information back via return is kinda large, and it is way easier to understand why and how they are returning what they are returning with an enum rather than True/False or 1/0.

src/core/or/circuitpadding.h
mikeperry-tor: No, each circuit has up to two machines. They can be both pad to the same hop (which is what WTF-PAD and APE do -- they use one machine for handling received packets and a second machine for handling sent packets), or they can be at different hops.

src/core/or/circuitpadding.h
mikeperry-tor: The mutable piece is the important modifier here. This is all the information that we need to be able to update as part of running the state machine. Calling it the mutable "state" would be misleading -- some of the information here is associated with the current state, but some of it is associated with an entire padding state machine.

This whole struct is an optimization to reduce the need to re-allocate things that don't need to be re-allocated for each circuit (ie stuff in circpad_machine_t), and to make a strong barrier to prevent bugs from accidentally modifying global padding machines.

src/core/or/circuitpadding.h
mikeperry-tor: Yes.

src/lib/smartlist_core/smartlist_foreach.h
```diff
@@ -83,6 +83,14 @@
          ++var ## _sl_idx) {                                    \
       var = (sl)->list[var ## _sl_idx];
 
+#define SMARTLIST_FOREACH_REVERSE_BEGIN(sl, type, var)  \
```
nmathewson: Let's have a unit test for this.

Also, let's document that it isn't compatible with SMARTLIST_DEL_CURRENT

src/core/or/circuitpadding.h
nmathewson: Hm.  So, I guess 4 is probably good enough for now, especially if we don't think these need to be serialized except in torrc file?

src/core/or/circuitpadding.h
nmathewson: Let's have this called "machine_spec" and the other one called machine_state then.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+#include "core/or/cell_st.h"
+#include "core/or/extend_info_st.h"
+#include "core/crypto/relay_crypto.h"
+#include "feature/nodelist/nodelist.h"
+
+#include "app/config/config.h"
+
+/* XXX: This is a dup of the constant in ./src/lib/time/tvdiff.c.
+ * Should/Do we have a header for time constants like this? */
+#define TOR_USEC_PER_SEC (1000000)
```
nmathewson: lib/defs would be a good place to put a header like that.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+    break;
+  default:
+    descr ="INVALID";
+  }
+
+  return descr;
+}
+
+/**
+ * Return the circpad_stat_t for the current state based on the
```
nmathewson: stat -> state.

Also let's document that it can return NULL, and when it does so.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+circpad_machine_remove_higher_token(circpad_machineinfo_t *mi,
+                                    circpad_delay_t target_bin_us)
+{
+  /* We need to remove the token from the first bin
+   * whose upper bound is greater than the target, and that
+   * has tokens remaining. */
+  int i = circpad_machine_first_higher_index(mi, target_bin_us);
+
+  if (i >= 0 && i < mi->histogram_len-1) {
+    tor_assert(mi->histogram[i]);
```
nmathewson: This kind of assertion is worrying.  Is this kind of a bug really worth crashing Tor over?  Remember, we've had a lot of DoS bugs in the past caused by logic errors that ran into a tor_assert().

Please go over this file and see how many of the assertions in it could use BUG or tor_assert_nonfatal() instead.

nmathewson: I'd suggest a decrement macro that does something like `if (!BUG(foo == 0)) --foo;`

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+  /* If we have scheduled padding some time in the future, we want to see what
+     bin we are in at the current time */
+  target_bin_us = (circpad_delay_t)
+                  MIN((current_time - mi->padding_scheduled_at_us),
+                      CIRCPAD_DELAY_INFINITE-1);
+
+  /* We are treating this non-padding cell as a padding cell, so we cancel
+     padding timer, if present. */
+  mi->padding_scheduled_at_us = 0;
+  if (mi->padding_timer_scheduled) {
```
nmathewson: Would it be more convenient to have a timer_is_scheduled() function to call here?  This seems like something that the timer itself should know.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+}
+
+/**
+ * Check our state's cell limit count and tokens.
+ *
+ * Returns 1 if either limits are hit and we decide to change states,
+ * otherwise returns 0.
+ */
+static circpad_decision_t
+circpad_check_token_supply(circpad_machineinfo_t *mi)
```
nmathewson: This function is a predicate; it should have a name that describes _what_ it is checking.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+circpad_send_command_to_hop(origin_circuit_t *circ, int hopnum,
+                            uint8_t relay_command, const uint8_t *payload,
+                            ssize_t payload_len)
+{
+  crypt_path_t *target_hop = circuit_get_cpath_hop(circ, hopnum);
+  int ret;
+
+  /* Check that the cpath has the target hop */
+  if (!target_hop) {
+    log_fn(LOG_WARN,LD_CIRC,
```
nmathewson: Should this (and the warning below) be LD_BUG?

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+  }
+
+  /* Update padding counts */
+  mi->padding_sent++;
+  if (mi->padding_sent == UINT16_MAX) {
+    mi->padding_sent /= 2;
+    mi->nonpadding_sent /= 2;
+  }
+  circpad_global_padding_sent++;
+  if (circpad_global_padding_sent == UINT64_MAX) {
```
nmathewson: If you sent a billion padding cells every second, it would take over 500 years to send this much padding.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+ * Timers are associated with circpad_machineinfo_t's. When the machineinfo
+ * is freed on a circuit, the timers are cancelled. Since the lifetime
+ * of machineinfo is always longer than the timers, handles are not
+ * needed.
+ */
+static void
+circpad_send_padding_callback(tor_timer_t *timer, void *args,
+                              const struct monotime_t *time)
+{
+  circpad_machineinfo_t *mi = ((circpad_machineinfo_t*)args);
```
nmathewson: Let's also mention the lifetime rules here in the documentation for the timer fields.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+
+  // TODO-MP-AP: Unify this counter with channelpadding for rephist stats
+  //total_timers_pending--;
+}
+
+/**
+ * Cache our consensus parameters upon consensus update.
+ */
+void
+circpad_new_consensus_params(networkstatus_t *ns)
```
nmathewson: could this argument be const?

(Please look over the file to see what other arguments could be const)

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+     * infinity event later? That would be a strange machine,
+     * but there's no reason to make it impossible. */
+    return CIRCPAD_STATE_UNCHANGED;
+  }
+
+  if (in_us <= 0) {
+    return circpad_send_padding_cell_for_callback(mi);
+  }
+
+  timeout.tv_sec = in_us/TOR_USEC_PER_SEC;
```
nmathewson: Is this 32-bit math or 64-bit math?  We need to be careful of doing frequent 64-bit division on 32-bit architectures; it's really slow.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+ * the particular hopnum.
+ *
+ * Hopnum starts at 1 (1=guard, 2=middle, 3=exit, etc).
+ *
+ * Payload may be null.
+ *
+ * Returns negative on error, 0 on success.
+ */
+static int
+circpad_send_command_to_hop(origin_circuit_t *circ, int hopnum,
```
nmathewson: Can we split the parts of this file that interact with the network into their own module, and maybe abstract them somehow? I expect that would make testing easier.

src/core/or/circuitpadding.c
nmathewson: Can we split the parts of this file that do padding negotiation into their own module? This file is pretty large...

src/core/or/circuitpadding.h
asn-d6: Thanks for suggesting this Nick. I'm also not sure whether this is worth doing, but I always thought that using just 4 states is really arbitrary.

I'm not sure if going HMM-style is the right way to do it.

But here is some pseudocode of generalizing the current 4 state system to an arbitrary number.

However, I don't really know how we could serialize this in the consensus, as Nick mentions.

```suggestion
typedef struct circpad_machine_event_t {
  int from_state;
  int to_state;
  enum circpad_event_t
};

typedef struct circpad_machine_state_t {
  char *name;
  circpad_machine_event_t *transition_events;
  circpad_state_t;
}

static int
initialize_machine_states(machine)
{
  machine_register_new_state(machine, "alpha");
  machine_register_new_state(machine, "beta");
  machine_register_new_state(machine, "gamma");

  /* Register transition events from START -> BURST */
  retval = machine_register_transition_events(machine, "alpha", "beta",
                                              CIRCPAD_EVENT_PADDING_RECV |
                                              CIRCPAD_EVENT_NONPADDING_RECV);
  if (retval < 0) { goto err; }

  /* Register transition events from BETTA -> GAMMA */
  retval = machine_register_transition_events(machine, "beta", "gamma", CIRCPAD_EVENT_NONPADDING_RECV);
  if (retval < 0) { goto err; }

  retval = machine_register_cancel_events(machine, "beta", CIRCPAD_EVENT_PADDING_RECV);
  if (retval < 0) { goto err; }
}
```

mikeperry-tor: Ok this was very simply to do so I just did it. 

36f85b04c2706d13a6b4086d39eacfab6345f0d2 has the header changes to support 255 states. 139ff5c6b0e78ff8edd54b1797088eee9e50b5c4 is the changes to the implementation. c41eefa615afd415857a7a8e8bbf0ff54fb4bf4d is the tests.

We don't need complicated registration APIs unless we want custom-defined string names. I don't think that is necessary. I think for custom states, we can just say "CUSTOM" or print the number.

With this change, the padding machines are now effectively a full hidden markov model. This means that future padding systems can do things like simulate entire browsing sessions across a variety of simulated pages (one or more pages per state). This is a very powerful ability, and worth providing.

Remaining question is if we should invert the state->event transition array to be event->state mapping, as nick suggested in https://github.com/torproject/tor/pull/461#discussion_r231697374. Now that we can have many more states than events, inverting may be a good idea. Also, if we invert, we can probably support up to 65535 states, which we maybe also want to do, because with that inversion, going from 255 to 65535 costs a byte and nothing else.

src/core/or/circuitpadding.c
teor2345: And collect the other TOR_USEC_PER_SEC definitions in the tor codebase.
(We just added another one in 0.3.5.4)

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+     * but there's no reason to make it impossible. */
+    return CIRCPAD_STATE_UNCHANGED;
+  }
+
+  if (in_us <= 0) {
+    return circpad_send_padding_cell_for_callback(mi);
+  }
+
+  timeout.tv_sec = in_us/TOR_USEC_PER_SEC;
+  timeout.tv_usec = (in_us%TOR_USEC_PER_SEC);
```
teor2345: A modulus is also a division.

> Is this 32-bit math or 64-bit math? We need to be careful of doing frequent 64-bit division on 32-bit architectures; it's really slow.

src/core/or/circuitpadding.h
mikeperry-tor: circpad_machine_spec_t seems ok. I don't like calling circpad_machineinfo_t circpad_machine_state_t, because we already have states in the machines... They are state machines. We need a different word..

maaaaayybe circpad_machinespec_t and circpad_machinestate_t are ok, but we still have circpad_state_t... Feels super weird to use the discrete math term 'state' and the generic term 'state' in the same codebase. I feel like mutableinfo is still better.

src/core/or/circuitpadding.c
mikeperry-tor: 22fbd5fa796c483bda6e10b0015c9feedf288300

src/core/or/circuitpadding.c
mikeperry-tor: 43012df348c6e158dc6826fac143d0cf036c3a62

src/core/or/circuitpadding.c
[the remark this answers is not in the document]
mikeperry-tor: Taken as-is in e4d18acbec3ea09e6a16fbd8dc8f9ca70ef3d60a (made it a fixup commit message).

mikeperry-tor: I think I found the issue: we weren't checking (or properly returning) error codes from the negotiation function.  7525900705a4bfb013fff7f895de47ced4cbc1fc

Note that machines do in fact set up right as soon as conditions are met. So that explains 2). I think this property is OK. We tear them down upon receipt of any error, and some machines may want to start padding ASAP.

src/core/or/circuitpadding.h
mikeperry-tor: Ok, inverting this does make sense if we may have more states than events. Since I think we do want to have the option for more states, I did this inversion.

Header changes:  62748b4a3b7413d56ce68009c0003ffe7fe531ec. Implementation changes: bf45f1866702a7d3586a0c1a479aacb4a4ede55c. Test changes: c54dab42ca887445c9102871b30956eee71c9ae0

src/core/or/circuitpadding.h
mikeperry-tor: Ok, I also expanded the max number of states to 65535 and inverted the transition array. See #461 (comment)

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+                      == NULL);
+  tor_assert_nonfatal(on_circ->padding_info[machine->machine_index] == NULL);
+
+  on_circ->padding_info[machine->machine_index] =
+      circpad_circuit_machineinfo_new(on_circ, machine->machine_index);
+  on_circ->padding_machine[machine->machine_index] = machine;
+}
+
+static void
+circpad_circ_client_machine_init(void)
```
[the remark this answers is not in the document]
mikeperry-tor: So we can do this a few different ways.. We can set time limits that mark a circuit as not to be closed until the timer is up, or we can set a flag on machines that say don't close the circuit if this machine is active. Or we could set that flag, and have an additional timer that says the machine stays alive until its up.

Timers are hard because we'd need a distribution that specifies lifetime. Setting a flag that says "don't close circuit if this machine is active" is easy. And then those machines can shut themselves down using state length limits, infinity bin probability, and or bins empty.

What do you think about these options?

src/test/test_circuitpadding.c
```diff
@@ -0,0 +1,975 @@
...
+    CIRCPAD_EVENT_PADDING_RECV |
+    CIRCPAD_EVENT_NONPADDING_RECV;
+
+  circ_client_machine.burst.transition_cancel_events =
+    CIRCPAD_EVENT_NONPADDING_SENT;
+
+  // FIXME: Is this what we want?
+  circ_client_machine.burst.token_removal = CIRCPAD_TOKEN_REMOVAL_HIGHER;
+
+  // FIXME: Tune this histogram
```
asn-d6: So I've been writing unittests. I wanted to run a unittest that tests more token removal functions. I started off by copying the setup logic of `test_circuitpadding_tokens()` but then I wanted to verify the width of the various bins of the histogram below. I know that the first bin should be of width 0, and after that the bins should double in size until `range_usec`.

I tried to verify this assumption by writing code like this:
```
  /* Test first bin width */
  circpad_delay_t d1 = circpad_histogram_bin_to_usec(mi, 0);
  circpad_delay_t d2 = circpad_histogram_bin_to_usec(mi, 1)-1;
  uint32_t first_bin_width = d2-d1;
  printf("first_bin_width: %u\n", first_bin_width);

  d1 = circpad_histogram_bin_to_usec(mi, 1);
  d2 = circpad_histogram_bin_to_usec(mi, 2)-1;
  uint32_t second_bin_width = d2-d1;
  printf("second_bin_width: %u\n", second_bin_width);

  d1 = circpad_histogram_bin_to_usec(mi, 2);
  d2 = circpad_histogram_bin_to_usec(mi, 3)-1;
  uint32_t third_bin_width = d2-d1;
  printf("third_bin_width: %u\n", third_bin_width);

  d1 = circpad_histogram_bin_to_usec(mi, 3);
  d2 = circpad_histogram_bin_to_usec(mi, 4)-1;
  uint32_t fourth_bin_width = d2-d1;
  printf("fourth_bin_width: %u\n", fourth_bin_width);
```

And here are the outputs:
```
first_bin_width: 0
second_bin_width: 249998
third_bin_width: 249999
fourth_bin_width: 4294466794
```

I think it's normal that the 4th bin comes out to UINT_MAX because the `bin_to_usec()` function returns infinite delay if `bin >= CIRCPAD_INFINITY_BIN(state)`, but I don't understand why the other two bins don't come out to exponential size.

Maybe what I'm doing with subtracting -1 from bin_to_usec is not right, but I got that from the documentation of the function:
```
 * Calculate the lower bound of a histogram bin. The upper bound
 * is obtained by calling this function with bin+1, and subtracting 1.
```

src/test/test_circuitpadding.c
```diff
@@ -0,0 +1,1026 @@
...
+
+extern smartlist_t *connection_array;
+extern networkstatus_t *current_ns_consensus;
+extern networkstatus_t *current_md_consensus;
+
+#define USEC_PER_SEC (1000000)
+#define NSEC_PER_USEC (1000)
+#define NSEC_PER_MSEC (1000*1000)
+
+circid_t get_unique_circ_id_by_chan(channel_t *chan);
```
asn-d6: Why are we re-defining these functions in the test file? Normally we don't need to do that, we just define them in their proper .c file and mark them as `STATIC` if they are internal and only needed in the tests.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+ *
+ * This performs inverse transform sampling
+ * (https://en.wikipedia.org/wiki/Inverse_transform_sampling).
+ *
+ * XXX: These formulas were taken verbatim. Need a floating wizard
+ * to check them for catastropic cancellation and other issues (teor?).
+ * Also: is 32bits of double from [0.0,1.0) enough?
+ */
+static double
+circpad_distribution_sample(circpad_distribution_t dist)
```
asn-d6: Not sure why we are passing by value here.. I think The Right Thing would be to pass a constified pointer to the dist.

src/test/test_circuitpadding.c
```diff
@@ -0,0 +1,975 @@
...
+   *    a. test non-update of RTT
+   * 3. Test client side circuit and non-application of RTT..
+   */
+  circpad_delay_t rtt_estimate;
+  (void)arg;
+
+  MOCK(circuitmux_attach_circuit, circuitmux_attach_circuit_mock);
+
+  dummy_channel.cmux = circuitmux_alloc();
+  relay_side = (circuit_t *)new_fake_orcirc(&dummy_channel,
```
asn-d6: I think we should use the TO_CIRCUIT() macro here instead of casting.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2491 @@
...
+{
+  double p = 0;
+
+  switch (dist.type) {
+    case CIRCPAD_DIST_NONE:
+      return 0;
```
asn-d6: We are not supposed to enter this functio with NONE so perhaps we can BUG() in this case.

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2491 @@
...
+      // param2 is upper bound, param1 is lower
+      /* The subtraction is exact as long as param2 and param1 are less than
+       * 2**53. The multiplication is accurate as long as (param2 - param1)
+       * is less than 2**52. (And when they are large, the low bits aren't
+       * important.) The result covers the full range of outputs, as long as
+       * p has a resolution of 1/2**32 or greater. */
+      p *= (dist.param2 - dist.param1);
+      p += dist.param1;
+      return p;
+    case CIRCPAD_DIST_LOGISTIC:
```
asn-d6: Please see my commit e3ed32e9f6 for some tests that cover this function. The tests fail because some of these probability distribution functions are busted. In particular the LOGISTIC distribution returns elements out of range, and same for the GEOMETRIC distribution,  whereas the  pareto distribution seems to always return the same value?

PADDING_TODO.txt
```diff
@@ -0,0 +1,67 @@
+TODO sketch for this branch, in rough priority order:
+
+- Clean up/fix XXX's and FIXMEs
+  - Test SMARTLIST_FOREACH_REVERSE_BEGIN()
```
asn-d6: See commit 0c014228ab for a test of the SMARTLIST_FOREACH_REVERSE_BEGIN macro

PADDING_TODO.txt
```diff
@@ -0,0 +1,67 @@
...
+  - Tests for circuit getting marked or freed before callback.
+  - Tests for detecting padding from wrong hop
+  - Tests for machine replacement on all conditions
+  - Tests for STATE_END teardown due to infinity, bins_empty, and/or length
+    with pending timers, new machines, etc.
+  - Test histogram spacing for larger length histograms
+  - Tests for consensus and machine rate limiting
```
asn-d6: Done in 94fe23d02b.

src/test/test_circuitpadding.c
mikeperry-tor: Err can we try more than one two datapoints here? The actual property I was going for was expoentially spaces bin cutoffs. Saying that the bins are doubling is perhaps poor phrasing, but it should actually happen after the third bin (in a histogram with enough bins).

src/core/or/circuitpadding.c
mikeperry-tor: Hrrm actually passing these on the stack means that inlining and/or fastcall (register argument passing) can take place (and does on my compiler). So it's actually much faster than pushing a pointer from the stack onto the stack, dereferencing twice, and then copying to registers to do the math. The reason is because this is a tiny struct with three arguments and the compiler figures that out. Not sure re pointer.

asn-d6: Sounds good. no need to do this.

src/test/test_circuitpadding.c
asn-d6: ACK. I'm on the process of making this test better and more robust. Here are the answers for a histogram with:
```
  burst_state->histogram_len = 15;
  burst_state->start_usec = 0;
  burst_state->range_usec = 100000;
```

Here we go:
```
 0 bin_width: 0
1 bin_width: 22
2 bin_width: 23
3 bin_width: 48
4 bin_width: 97
5 bin_width: 194
6 bin_width: 390
7 bin_width: 780
8 bin_width: 1562
9 bin_width: 3124
10 bin_width: 6249
11 bin_width: 12499
12 bin_width: 24999
13 bin_width: 4294917294
14 bin_width: 4294967295
```

src/core/or/circuitpadding.c
asn-d6: The prob distr functions were not busted. I was busted.
In any case, these tests are replaced by the new Riastradh code.

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,539 @@
...
+/**
+ * A circuit padding state machine state.
+ *
+ * This struct describes the histograms and parameters of a single
+ * state in the adaptive padding machine. Instances of this struct
+ * exist in global circpad machine definitions that come from torrc
+ * or the consensus, and are immutable.
+ */
+typedef struct circpad_state_t {
+  /** how long the histogram is (in bins). Histograms must have at least 2
```
asn-d6: Hm, is this accurate? Isn't it the case that the histogram has `histogram_len-1` bins?

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+        circpad_histogram_bin_to_usec(mi, i+1) > target_bin_us) {
+      return i;
+    }
+  }
+
+  return mi->histogram_len;
+}
+
+/**
+ * Find the index of the first bin whose lower bound is
```
asn-d6: We need to fix the documentation (and maybe naming) of this function and of `circpad_machine_remove_higher_token` since they dont return the first bin that is `lower`, they return the first bin that is `lower-or-equal` to the target.

src/core/or/circuitpadding.c
asn-d6: Done in 6d9d5660f9.

src/core/or/circuitpadding.c
asn-d6: Hmm, I'm not sure. `in_usec` seems to be a uint32_t in the lines below:

```
  timeout.tv_sec = in_usec/TOR_USEC_PER_SEC;
  timeout.tv_usec = (in_usec%TOR_USEC_PER_SEC);
```
And here is the assembly that gets generated for the following two lines:

```
mov    $0x431bde83,%r8d
mov    %ebp,%eax
mov    %ebp,%r9d
mov    $0x6,%edi
mul    %r8d
lea    0x269b56(%rip),%rcx        # 0x555555b713e7
mov    $0x400,%esi
mov    %edx,%r8d
lea    0x269ea0(%rip),%rdx        # 0x555555b71740 <__FUNCTION__.26195>
shr    $0x12,%r8d
mov    %r8d,%eax
mov    %rax,(%rsp)

imul   $0xf4240,%r8d,%eax
sub    %eax,%r9d
mov    %r9d,%eax
mov    %rax,0x8(%rsp)
```

So it seems to be implementing the division using right shifts `shr` and a multiplication `mul`? But how does that answer your question?

src/core/or/circuitpadding.c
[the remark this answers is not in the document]
asn-d6: I lilke the approach of a machine flag that says" don't close circuit if this machine is active". I think that makes sense and it would work in the case of IP circs.

src/core/or/circuitpadding.c
nmathewson: If both arguments are 32-bit values, it should only be using 32-bit math.  So, cool.  I think I probably just wasn't looking at the definition of circuitpad_delay_t when I wrote that question.

src/core/or/circuitpadding.c
asn-d6: I opened a ticket about this in https://trac.torproject.org/projects/tor/ticket/28777#ticket because I think trying to split the file now will be pretty hasty and won't give us a good interface. It will also cause lots of conflicts since me and Mike are both working on the same branch. Let's do it after the branch is merged IMO.

src/core/or/circuitpadding.c
asn-d6: I opened a ticket about this in https://trac.torproject.org/projects/tor/ticket/28777#ticket because I think trying to split the file now will be pretty hasty and won't give us a good interface. It will also cause lots of conflicts since me and Mike are both working on the same branch. Let's do it after the branch is merged IMO.

src/core/or/circuitpadding.h
asn-d6: Opened https://trac.torproject.org/projects/tor/ticket/28778 for this. Doing this right now will cause lots of conflicts between our branches so let's do it after merge.

src/lib/smartlist_core/smartlist_foreach.h
asn-d6: Done in 4c44d7b116 and 072e7c87e8 .

src/core/or/circuitpadding.c
asn-d6: Addressed in 9b4c261b2d.

src/core/or/circuitpadding.c
asn-d6: Addressed in 01ef88908a.

src/core/or/circuitpadding.c
asn-d6: Addressed in ba9b856347 .

src/core/or/circuitpadding.c
asn-d6: Addressed in 01ef88908a .

src/test/test_circuitpadding.c
asn-d6: Opened https://trac.torproject.org/projects/tor/ticket/28779#ticket for this since it would be hard to do this without causing too many conflicts and a hard to squash commit.

src/test/test_circuitpadding.c
asn-d6: Addressed in e48f1eb306.

src/core/or/circuitpadding.c
asn-d6: Addressed in 75b7da0ced.

src/core/or/circuitpadding.c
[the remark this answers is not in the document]
asn-d6: OK opened a ticket for the "dont close circ if machine is active" idea: https://trac.torproject.org/projects/tor/ticket/28780

So I think we can perhaps aim for the following to get a useful circ machine for 040 (unclear if we will have enough time tho):

- Do both tickets #28633 and #28780.
- For circuit purposes `CIRCUIT_PURPOSE_C_INTRO*` we introduce a simple padding machine that does some light randomized padding to obfuscate the introduction activity. We also setup the #28780 flag to not close the circuit before the machine is done.
- For circuit purposes `CIRCUIT_PURPOSE_C_*REND*` we introduce another simple padding machine that does some light randomized padding to obfuscate rendezvous setup/waiting.

This way we will slightly obfuscate the introduction and rendezvous circuits. The other remaining thing we could do is obfuscate the HS descriptor fetch but that might be harder, so perhaps we can skip this for 040.

How does that sound to you, Mike?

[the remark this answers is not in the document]
mikeperry-tor: This sounds reasonable. I believe that #28633 is supported already in the current code. #28780 will be a little tricky since we need to go through the hs code and circuit idle timeout code and add checks that say "if there's a machine here, don't close the circuit" and still remember to close the circuit once the machine is gone (which might be doable from the circuit idle handling).

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2370 @@
...
+  tor_assert(i < state->histogram_len);
+  tor_assert(histogram[i] > 0);
+
+  // Store this index to remove the token upon callback.
+  if (state->token_removal != CIRCPAD_TOKEN_REMOVAL_NONE) {
+    mi->chosen_bin = i;
+  }
+
+  if (i == state->histogram_len-1) {
+    if (state->token_removal != CIRCPAD_TOKEN_REMOVAL_NONE &&
```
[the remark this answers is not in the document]
mikeperry-tor: > I'm a bit confuse on how the token removal logic works. I don't think it corresponds to what `3.1.2` in `prop254` says. In particular, why is there an explicit rule here about removing tokens from the infinity bin? Why is this not handled in the token removal functions?

The infinity bin has to have its tokens removed as soon as "infinite" delay is chosen. This is because there is no callback for it. For all other bins, tokens are removed upon the padding callback, or if a non-padding cell is sent *before* the callback. We could have defined the infinity token removal to happen only if no non-padding cell ever is sent.. But those semantics are both more complicated, and not as useful for having a fixed number of "infinite" padding delays in a state. Does this make sense?

> Furthermore, we also remove tokens from bins in `circpad_send_padding_cell_for_callback()` regardless of what our token removal strategy is, even tho the spec says `If token removal is enabled, when a padding packet is sent, a token is removed from the bin corresponding to the target delay.`.

This is correct, and the implementation does in fact match the spec. So long as the token removal strategy is not NONE, then we will have a mutable histogram in the machine info. For all token removal strategies other than NONE, the padding case is the same -- remove from exactly the bin corresponding to the chosen delay. This bin should not be empty by definition -- otherwise padding could not have been scheduled for this bin's delay value.. Hence the asserts before token removal upon actually sending the padding cell.

> Also, the spec seems to imply that special token removal rules only apply when the exact bin has no tokens left. I don't see where this logic is in the code. I just see us applying the token removal rule, when we send a non-padding cell.

Every token removal rule other than NONE will try to remove from exactly the bin corresponding to the delay value that passed. It is only if this bin is empty do they apply their strategy. I could see an argument for factoring out this case and only applying the strategy-specific code if the target bin is in fact empty, but this mans either duplicating the target bin calculations, or passing the target bin weirdly between a bunch of the helper functions...

src/core/or/circuitpadding.h
asn-d6: Resolved in 22080a0be3427d9eb457b4be461308cf32c84f81.

src/core/or/circuitpadding.h
asn-d6: Made it an unsigned per the other comments in 22080a0be3427d9eb457b4be461308cf32c84f81.

src/core/or/circuitpadding.h
asn-d6: They were only mentioned in passing. I pushed a new commit in my `bug28142` brnach that specifies them further: https://github.com/asn-d6/torspec/commit/09814d9109d801a9995890dbde0dcca1f5370683

src/core/or/circuitpadding.c
asn-d6: Hmm, the timer code is hairy enough that I don't feel comfortable writing that method right now. I opened #28821 to do it in the future.

src/core/or/circuitpadding.c
asn-d6: Verifying Nick's logic here:
```
five_years_in_secs = 5*365*86400
uint64_max = 0xffffffffffffffff
In [14]: uint64_max/five_years_in_secs
Out[14]: 116988483471L
```

src/core/or/circuitpadding.h
asn-d6: Addressed in 80ab22476ff2c7e2a6ce91f74879aec18b81a9bc.

src/core/or/circuitpadding.h
asn-d6: Addressed in e1805053a48db1b489268c6ac3be5e3792fed00d.

src/core/or/circuitpadding.h
asn-d6: Made it unsigned as per other comments in 1e2564c5c7ab86384000cc74dc797b5d972fb065.

src/core/or/circuitpadding.h
asn-d6: Addressed in e1805053a48db1b489268c6ac3be5e3792fed00d .

src/core/or/circuitpadding.h
asn-d6: Addressed in e1805053a48db1b489268c6ac3be5e3792fed00d.

src/core/or/circuitpadding.h
asn-d6: Addressed in e1805053a48db1b489268c6ac3be5e3792fed00d.

src/core/or/circuitpadding.h
asn-d6: Addressed in e1805053a48db1b489268c6ac3be5e3792fed00d.

src/core/or/circuitpadding.h
asn-d6: Addressed in e1805053a48db1b489268c6ac3be5e3792fed00d.

src/core/or/circuitpadding.h
asn-d6: Addressed in e1805053a48db1b489268c6ac3be5e3792fed00d.

src/core/or/circuit_st.h
asn-d6: Addressed in 53fb50bd3cfccd93a3d17af18f2f9c0ad42a7cf9.

src/core/or/circuit_st.h
asn-d6: Addressed in 53fb50bd3cfccd93a3d17af18f2f9c0ad42a7cf9.

src/core/or/circuitpadding.h
asn-d6: Addressed in 1e2564c5c7ab86384000cc74dc797b5d972fb065.

src/core/or/circuitpadding.c
asn-d6: Addressed in 3ee2f69ddb5c9a02c4d15a6908ea4cf53eaf7c93.

src/core/or/circuitpadding.c
asn-d6: Addressed all of them in 0d7d10e3ad. Please review carefully! :)

src/core/or/circuitpadding.c
asn-d6: Addressed in e52653c14d44f87df4f8a256bcf69a2a5b6775b9.

src/core/or/circuitpadding.c
asn-d6: Addressed in f413a343fbf58e1729d8bc3c35db0d2222e86c3e. It caused some BUGs in the tests which I plugged with b0f2c71627544b87d30f0bfacf5a3d5049505d2d.

src/core/or/circuitpadding.c
asn-d6: Clamping code removed in f413a343fbf58e1729d8bc3c35db0d2222e86c3e.

src/core/or/circuitpadding.c
asn-d6: Addressed in ed9a4713b9e0c4ab8c3b7ae18573c0b809b708f0.

src/core/or/circuitpadding.c
asn-d6: Addressed in ed9a4713b9e0c4ab8c3b7ae18573c0b809b708f0.

src/test/test_circuitpadding.c
asn-d6: Addressed in 3c5af3253f70a672c86bf9f7022e408425695bf3.

src/core/or/circuitpadding.c
asn-d6: Done in 0a4e04a84f0b18343bd38732c78951c295c73a1d.

src/lib/smartlist_core/smartlist_foreach.h
asn-d6: Done in 4c44d7b116 and 1b706523ee.

src/core/or/circuitpadding.c
asn-d6: Addressed in d8a2e7dcd0.

PR #468, opened 2018-11-02
Serve bandwidth file used in the next vote

juga0:
When a directory authority is using a bandwidth file to obtain the
bandwidth values that will be included in the next vote, serve this
bandwidth file at /tor/status-vote/next/bandwidth.z.

### DIFF
--- a/src/feature/dircache/dircache.c
+++ b/src/feature/dircache/dircache.c
@@ -1447,7 +1447,6 @@ static int
 handle_get_next_bandwidth(dir_connection_t *conn,
                           const get_handler_args_t *args)
 {
-  (void)args;
   log_debug(LD_DIR, "Getting next bandwidth.");
   const or_options_t *options = get_options();
   const compress_method_t compress_method =
--- a/src/feature/dircache/dircache.c
+++ b/src/feature/dircache/dircache.c
@@ -357,12 +357,15 @@ static int handle_get_robots(dir_connection_t *conn,
                                 const get_handler_args_t *args);
 static int handle_get_networkstatus_bridges(dir_connection_t *conn,
                                 const get_handler_args_t *args);
+static int handle_get_next_bandwidth(dir_connection_t *conn,
+                                     const get_handler_args_t *args);
 
 /** Table for handling GET requests. */
 static const url_table_ent_t url_table[] = {
   { "/tor/", 0, handle_get_frontpage },
   { "/tor/status-vote/current/consensus", 1, handle_get_current_consensus },
   { "/tor/status-vote/current/", 1, handle_get_status_vote },
+  { "/tor/status-vote/next/bandwidth", 0, handle_get_next_bandwidth },
   { "/tor/status-vote/next/", 1, handle_get_status_vote },
   { "/tor/micro/d/", 1, handle_get_microdesc },
   { "/tor/server/", 1, handle_get_descriptor },
@@ -1435,6 +1438,27 @@ handle_get_networkstatus_bridges(dir_connection_t *conn,
     goto done;
   }
  done:
+  return 0;
+}
+
+/** Helper function for GET the bandwidth file used for the next vote */
+static int
+handle_get_next_bandwidth(dir_connection_t *conn,
+                          const get_handler_args_t *args)
+{
+  (void)args;
+  log_debug(LD_DIR, "Getting next bandwidth.");
+  const or_options_t *options = get_options();
+  if (options->V3BandwidthsFile) {
+    int lifetime = 60;
+    char *bandwidth =  read_file_to_str(options->V3BandwidthsFile, 0, NULL);
+    size_t len = strlen(bandwidth);
+    write_http_response_header(conn, len, NO_METHOD, lifetime);
+    connection_buf_add(bandwidth, len, TO_CONN(conn));
+    tor_free(bandwidth);
+  } else {
+    write_short_http_response(conn, 404, "Not found");
+  }
   return 0;
 }
 
--- a/src/feature/dircache/dircache.c
+++ b/src/feature/dircache/dircache.c
@@ -1451,7 +1451,8 @@ handle_get_next_bandwidth(dir_connection_t *conn,
   const or_options_t *options = get_options();
   if (options->V3BandwidthsFile) {
     int lifetime = 60;
-    char *bandwidth =  read_file_to_str(options->V3BandwidthsFile, 0, NULL);
+    char *bandwidth = read_file_to_str(options->V3BandwidthsFile,
+                                       RFTS_IGNORE_MISSING, NULL);
     size_t len = strlen(bandwidth);
     write_http_response_header(conn, len, NO_METHOD, lifetime);
     connection_buf_add(bandwidth, len, TO_CONN(conn));
--- a/src/feature/dircache/dircache.c
+++ b/src/feature/dircache/dircache.c
@@ -1453,13 +1453,15 @@ handle_get_next_bandwidth(dir_connection_t *conn,
     int lifetime = 60;
     char *bandwidth = read_file_to_str(options->V3BandwidthsFile,
                                        RFTS_IGNORE_MISSING, NULL);
-    size_t len = strlen(bandwidth);
-    write_http_response_header(conn, len, NO_METHOD, lifetime);
-    connection_buf_add(bandwidth, len, TO_CONN(conn));
-    tor_free(bandwidth);
-  } else {
-    write_short_http_response(conn, 404, "Not found");
+    if (bandwidth != NULL) {
+      site_t len = strlen(bandwidth);
+      write_http_response_header(conn, len, NO_METHOD, lifetime);
+      connection_buf_add(bandwidth, len, TO_CONN(conn));
+      tor_free(bandwidth);
+      return 0;
+    }
   }
+  write_short_http_response(conn, 404, "Not found");
   return 0;
 }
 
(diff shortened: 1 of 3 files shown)


src/feature/dircache/dircache.c
```diff
@@ -1438,6 +1441,27 @@ handle_get_networkstatus_bridges(dir_connection_t *conn,
...
+static int
+handle_get_next_bandwidth(dir_connection_t *conn,
+                          const get_handler_args_t *args)
+{
+  (void)args;
+  log_debug(LD_DIR, "Getting next bandwidth.");
+  const or_options_t *options = get_options();
+  if (options->V3BandwidthsFile) {
+    int lifetime = 60;
+    char *bandwidth =  read_file_to_str(options->V3BandwidthsFile, 0, NULL);
```
dgoulet-tor: This can fail so we have to check for it.

Also, I'm guessing you might want to pass `RFTS_IGNORE_MISSING` else if the file is missing, you'll have a lot of log warning that the file couldn't be open.

Double white space also.

src/feature/dircache/dircache.c
```diff
@@ -1438,6 +1441,27 @@ handle_get_networkstatus_bridges(dir_connection_t *conn,
...
+    size_t len = strlen(bandwidth);
```
dgoulet-tor: This explodes if `bandwidth` is NULL.

src/feature/dircache/dircache.c
```diff
@@ -1438,6 +1441,27 @@ handle_get_networkstatus_bridges(dir_connection_t *conn,
...
+    write_http_response_header(conn, len, NO_METHOD, lifetime);
```
dgoulet-tor: Spec mentions `bandwidth.z`. Aren't we suppose to compress here? For instance, see how `handle_get_descriptor()` does it.

src/feature/dircache/dircache.c
juga0: glups.
Question, is there any advantage on passing an struct stat to read_file_to_str and get the length from st_size instead of getting the length with strlen?

src/feature/dircache/dircache.c
juga0: would the connection spool be needed too?

juga0: i guess not because it's served only 1 file

src/feature/dircache/dircache.c
```diff
@@ -1438,6 +1441,27 @@ handle_get_networkstatus_bridges(dir_connection_t *conn,
...
+/** Helper function for GET the bandwidth file used for the next vote */
+static int
+handle_get_next_bandwidth(dir_connection_t *conn,
+                          const get_handler_args_t *args)
+{
+  (void)args;
+  log_debug(LD_DIR, "Getting next bandwidth.");
+  const or_options_t *options = get_options();
+  if (options->V3BandwidthsFile) {
+    int lifetime = 60;
```
teor2345: 60 seconds is a very short cache lifetime.
How often does sbws write the file? Once an hour?
If it's once an hour, use 30 minutes as the cache lifetime.

src/feature/dircache/dircache.c
teor2345: Not really:
https://github.com/torproject/tor/blob/537092cdbb7f4be0e6d68f4e5d65ca2a403375f9/src/lib/fs/files.c#L601-L617

read_file_to_str() makes sure that the sizes are consistent. If they're not, it fails:
https://github.com/torproject/tor/blob/537092cdbb7f4be0e6d68f4e5d65ca2a403375f9/src/lib/fs/files.c#L700-L710

src/feature/dircache/dircache.c
```diff
@@ -1449,13 +1449,19 @@ handle_get_next_bandwidth(dir_connection_t *conn,
   (void)args;
```
teor2345: args are used, so this line is not needed now.

juga0: last commit should fix that

src/test/test_dir_handle_get.c
```diff
@@ -2462,6 +2487,85 @@ test_dir_handle_get_status_vote_next_authority(void* data)
...
+
+  tt_assert(header);
+  tt_ptr_op(strstr(header, "HTTP/1.0 200 OK\r\n"), OP_EQ, header);
+  tt_assert(strstr(header, "Content-Encoding: deflate\r\n"));
+  tt_assert(strstr(header, "Content-Length: 167\r\n"));
+
+  tt_int_op(body_used, OP_EQ, strlen(body));
+  tt_str_op(content, OP_EQ, body);
+  goto done;
+
```
ahf: Nitpick: Indentation is a bit off here. 'goto labels' should have one space indentation and then afterwards the code should follow the indentation level before the label.

You also don't need the `goto done` at the end.

src/test/test_dir_handle_get.c
```diff
@@ -2462,6 +2487,85 @@ test_dir_handle_get_status_vote_next_authority(void* data)
...
+  done:
+    UNMOCK(get_options);
+    UNMOCK(connection_write_to_buf_impl_);
+    connection_free_minimal(TO_CONN(conn));
+    tor_free(header);
+    tor_free(body);
+    tor_free(expires);
+    or_options_free(mock_options); mock_options = NULL;
```
ahf: `or_options_free()` already sets the input to `NULL`.

src/test/test_dir_handle_get.c
```diff
@@ -2223,6 +2223,31 @@ test_dir_handle_get_status_vote_next_authority_not_found(void* data)
...
+
+  tt_int_op(0, OP_EQ, directory_handle_command_get(conn,
+    GET("/tor/status-vote/next/bandwdith"), NULL, 0));
+
+  fetch_from_buf_http(TO_CONN(conn)->outbuf, &header, MAX_HEADERS_SIZE,
+                      NULL, NULL, 1, 0);
+  tt_assert(header);
+  tt_str_op(NOT_FOUND, OP_EQ, header);
+
+  done:
```
ahf: Nitpick: Indentation is off here.

juga0:
last commit fix the nit-picking.
We should squash before merging

src/feature/dircache/dircache.c
```diff
@@ -1438,6 +1442,35 @@ handle_get_networkstatus_bridges(dir_connection_t *conn,
...
+  if (options->V3BandwidthsFile) {
+    char *bandwidth = read_file_to_str(options->V3BandwidthsFile,
+                                       RFTS_IGNORE_MISSING, NULL);
+    if (bandwidth != NULL) {
+      size_t len = strlen(bandwidth);
+      write_http_response_header(conn, len, compress_method,
+                                 BANDWIDTH_CACHE_LIFETIME);
+      if (compress_method != NO_METHOD)
+        conn->compress_state = tor_compress_new(1, compress_method,
+                                        choose_compression_level(len));
```
teor2345: choose_compression_level() takes the length of the compressed response, not the length of the input data.
Other code uses len/2 as an estimate:
https://github.com/torproject/tor/blob/8020d6fb05d9477e77c6ca554dc1288873f6115c/src/feature/dircache/dircache.c#L1024

src/feature/dircache/dircache.c
```diff
@@ -1438,6 +1442,35 @@ handle_get_networkstatus_bridges(dir_connection_t *conn,
...
+  const or_options_t *options = get_options();
+  const compress_method_t compress_method =
+    find_best_compression_method(args->compression_supported, 1);
+
+  if (options->V3BandwidthsFile) {
+    char *bandwidth = read_file_to_str(options->V3BandwidthsFile,
+                                       RFTS_IGNORE_MISSING, NULL);
+    if (bandwidth != NULL) {
+      size_t len = strlen(bandwidth);
+      write_http_response_header(conn, len, compress_method,
```
teor2345: len is the length of the response, or -1 if that length isn't known.

juga0: Shouldn't that line be?: `write_http_response_header(conn, compress_method != NO_METHOD ? -1 : len,`

teor2345: The code you suggest is better, but it will work either way.

src/feature/dircache/dircache.c
```diff
@@ -1456,13 +1456,19 @@ handle_get_next_bandwidth(dir_connection_t *conn,
...
-      if (compress_method != NO_METHOD)
+      ssize_t len = strlen(bandwidth);
+      write_http_response_header(conn, compress_method != NO_METHOD ? -1 : len,
+                                 compress_method, BANDWIDTH_CACHE_LIFETIME);
+      if (compress_method != NO_METHOD) {
         conn->compress_state = tor_compress_new(1, compress_method,
-                                        choose_compression_level(len));
-      connection_buf_add(bandwidth, len, TO_CONN(conn));
+                                        choose_compression_level(len/2));
+        log_debug(LD_DIR, "Calling connection_buf_add_compress.");
```
ahf: This log message should probably have a slightly different string to it. Something like "Compressing bandwidth cache file"

src/feature/dircache/dircache.c
```diff
@@ -1456,13 +1456,19 @@ handle_get_next_bandwidth(dir_connection_t *conn,
...
+        connection_buf_add_compress(bandwidth, len, conn, 0);
+        connection_buf_add_compress("", 0, conn, 1);
```
ahf: Maybe a comment here about why it's needed to call this with the empty string to finalize it? :-)

PR #478, opened 2018-11-05
Subsystems

nmathewson:


src/lib/err/torerr_sys.c
```diff
@@ -0,0 +1,39 @@
...
+#include "lib/err/torerr_sys.h"
+#include "lib/subsys/subsys.h"
+#include "lib/version/torversion.h"
+
+#include <stddef.h>
+
+static int
+torerr_subsys_init(void)
+{
+  configure_backtrace_handler(get_version());
```
dgoulet-tor: Shouldn't we keep the BUG that was removed and return -1 in case it fails?

src/app/main/subsysmgr.c
```diff
@@ -0,0 +1,130 @@
...
+    const subsys_fns_t *sys = tor_subsystems[i];
+    if (!sys->supported)
+      continue;
+    if (sys->level > target_level)
+      break;
+    if (sys_initialized[i])
+      continue;
+    int r = 0;
+    if (sys->initialize)
+      r = sys->initialize();
```
dgoulet-tor: Maybe I would add a debug log here stating what we are initializing. I think it might come handy later.

src/lib/err/torerr_sys.c
```diff
@@ -0,0 +1,39 @@
...
+  tor_log_reset_sigsafe_err_fds();
+  clean_up_backtrace_handler();
+}
+
+const subsys_fns_t sys_torerr = {
+  .name = "err",
+  .level = -100,
+  .supported = true,
+  .initialize = torerr_subsys_init,
+  .shutdown = torerr_subsys_shutdown
```
dgoulet-tor: I'm going to be a super nitpick here but I see these have all different names in the other subsystems added by this branch.

I would probably advocate for having a standard so it is easy to grep/find/identify in a code file.

Since they are static functions, I would vote for `subsys_<subsystem-name>_initialize()` or something like that?

src/lib/log/log_sys.c
```diff
@@ -0,0 +1,35 @@
...
+{
+  init_logging(0);
+  return 0;
+}
+
+static void
+shutdown_logging_subsys(void)
+{
+  logs_free_all();
+  escaped(NULL);
```
dgoulet-tor: Originally, the `escaped(NULL)` is before the free all. Will it impact anything?

src/lib/subsys/subsys.h
```diff
@@ -0,0 +1,63 @@
...
+   **/
+  int level;
+
+  /**
+   * Initialize any global components of this subsystem.
+   *
+   * This function MAY rely on any lower-level subsystem being initialized.
+   *
+   * This function MUST NOT rely on any runtime configuration information;
+   * it is only for global state or pre-configuration state.
```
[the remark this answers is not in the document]
nmathewson: The design here is that first the system initializes itself, and that configuration happens as a later step.  I'm trying to keep configuration and initialization separate.  I'll add a comment.

src/app/main/subsysmgr.c
```diff
@@ -0,0 +1,130 @@
...
+ * dynamically allocate in this module.)
+ **/
+static bool sys_initialized[128];
+
+/**
+ * Exit with a raw assertion if the subsystems list is inconsistent;
+ * initialize the subsystem_initialized array.
+ **/
+static void
+check_and_setup(void)
```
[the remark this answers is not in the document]
nmathewson: Do you have a better name in mind? I was hoping for something that validates the list and sets up the flag array.

src/app/main/subsysmgr.c
```diff
@@ -0,0 +1,130 @@
...
+subsystems_init_upto(int target_level)
+{
+  check_and_setup();
+
+  for (unsigned i = 0; i < n_tor_subsystems; ++i) {
+    const subsys_fns_t *sys = tor_subsystems[i];
+    if (!sys->supported)
+      continue;
+    if (sys->level > target_level)
+      break;
```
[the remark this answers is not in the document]
nmathewson: I'm not using the feature yet, but I intend that it should be possible to initialize everything up to a particular "target level", and then stop.  We would use this (for example) in stuff like subcommands, or in the unit tests, where we don't want to initialize all of Tor.

src/lib/subsys/subsys.h
```diff
@@ -0,0 +1,63 @@
...
+  int (*add_pubsub)(struct dispatch_connector_t *);
+
+  /**
+   * Free all resources held by this subsystem.
+   *
+   * This function is not allowed to fail.
+   **/
+  void (*shutdown)(void);
+
+} subsys_fns_t;
```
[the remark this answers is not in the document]
nmathewson: I've added a comment -- the answer is that all callbacks are optional.

src/app/main/subsysmgr.c
```diff
@@ -0,0 +1,130 @@
...
+  subsystems_shutdown_downto(MIN_SUBSYS_LEVEL - 1);
+}
+
+/**
+ * Shut down all the subsystems whose level is above <b>target_level</b>.
+ **/
+void
+subsystems_shutdown_downto(int target_level)
+{
+  check_and_setup();
```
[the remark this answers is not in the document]
nmathewson: It wipes sys_initialized only if subsystem_array_validated is false.  But if we have already called setup, then subsystem_array_validated will be true.

src/lib/err/torerr_sys.c
nmathewson: BUG isn't meaningful here -- without a backtrace handler, we can't bug.  But we can certainly return -1. Fixing.

src/lib/log/log_sys.c
nmathewson: It shouldn't; the escaped(NULL) is just meant to free the buffer held by escaped().

src/lib/err/torerr_sys.c
nmathewson: okay; I've done a commit to make them uniform.

src/lib/subsys/subsys.h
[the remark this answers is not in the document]
dgoulet-tor: So does that mean we are looking to have a `configuration` function call per-subsystem later? In other words, there needs to be a time where each subsystem can query the options and configure itself _early_ in the boot process.

src/app/main/subsysmgr.c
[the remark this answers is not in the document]
dgoulet-tor: Oh I see. Roger.

src/lib/subsys/subsys.h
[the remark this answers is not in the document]
dgoulet-tor: Ack.

src/lib/err/torerr_sys.c
dgoulet-tor: Ack.

src/lib/err/torerr_sys.c
dgoulet-tor: Neat!

src/lib/log/log_sys.c
dgoulet-tor: Ack.

PR #480, opened 2018-11-06
Adaptive padding 036 resquashed

asn-d6:


mikeperry-tor:
Can we close this PR? I believe it only has one commit not in my branch: https://github.com/torproject/tor/pull/480/commits/2f370d0419589f956b50b1909c97d60ca5174b57

PR #485, opened 2018-11-06
Ticket27047

juga0:


juga0:
Moved to https://github.com/torproject/tor/pull/492

PR #492, opened 2018-11-10
Ticket27047 02

juga0:


juga0:
appveyor is failing with tortls_openssl.c
I've changed to SHA1, since SHA256 might be too many characters to be something human readable.

juga0:
Move to https://github.com/torproject/tor/pull/497

PR #495, opened 2018-11-11
#27929: Create new periodic event for pruning old info about Tor routers

rl1987:
https://trac.torproject.org/projects/tor/ticket/27929

### DIFF
--- a/src/core/mainloop/mainloop.c
+++ b/src/core/mainloop/mainloop.c
@@ -1359,6 +1359,7 @@ CALLBACK(heartbeat);
 CALLBACK(hs_service);
 CALLBACK(launch_descriptor_fetches);
 CALLBACK(launch_reachability_tests);
+CALLBACK(prune_old_routers);
 CALLBACK(reachability_warnings);
 CALLBACK(record_bridge_stats);
 CALLBACK(rend_cache_failure_clean);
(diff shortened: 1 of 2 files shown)


src/core/mainloop/mainloop.c
```diff
@@ -2220,6 +2225,27 @@ retry_dns_callback(time_t now, const or_options_t *options)
   return RETRY_DNS_INTERVAL;
 }
 
+/**
+ * Periodic callback: prune routerlist of old information about Tor network.
+ */
+static int
+prune_old_routers_callback(time_t now, const or_options_t *options)
+{
+#define ROUTERLIST_PRUNING_INTERVAL (60) // 1 minute.
```
nmathewson: We could afford to make this much longer...

src/core/mainloop/mainloop.c
```diff
@@ -1454,6 +1457,7 @@ static periodic_event_item_t *fetch_networkstatus_event=NULL;
 static periodic_event_item_t *launch_descriptor_fetches_event=NULL;
 static periodic_event_item_t *check_dns_honesty_event=NULL;
 static periodic_event_item_t *save_state_event=NULL;
+static periodic_event_item_t *prune_old_routers_event=NULL;
```
nmathewson: Do we need this?  NAMED_CALLBACK and callbacks here are only for the callbacks that we want to turn on and off by hand.

src/core/mainloop/mainloop.c
nmathewson: Like, 60 minutes?

rl1987:
These two places are `router_reload_router_list_impl()` and `router_rebuild_store()`, both of which are related persisting routerlist into file store, and may not actually call `router_reload_router_list_impl()` depending on exact codepath. I imagine the reason why this was done from `check_descriptor_callback()` is that we want to prune routerlist at some periodic interval, without depending on secondary effect of routerlist persistence logic. That would be the motivation of having it in `check_descriptor_callback()` in the first place and having it in new periodic event now.

PR #509, opened 2018-11-15
Ticket28453: End of Life Tor Instructions

teor2345:


doc/HACKING/EndOfLifeTor.md
```diff
@@ -0,0 +1,52 @@
...
+   Send the announcement out with every new release announcement.
+
+1. A month before End of Life:
+   Send the announcement to tor-announce, tor-talk, tor-relays, and the
+   packagers.
+
+2. A few days before End of Life:
+   Get at least two of weasel/arma/Sebastian to
+   remove the old version number from their approved versions list.
+   Give them a few days to do this if you can.
```
dgoulet-tor: We could define a bit more in details this procedure because #27983 made it clear that the process is unclear. Historically, Roger would push a patch to the dirauth-conf git repository but lately it has been Nick asking weasel on IRC to do so and it would be pushed.

So I propose we continue towards that path with: "The stable maintainer notifies weasel/arma/Sebastian about the change considering it an official request". Then one of them pushes the update to the git repository and we move on.

doc/HACKING/EndOfLifeTor.md
```diff
@@ -0,0 +1,52 @@
...
+   - the jenkins builds
+   - tor's Travis CI cron jobs
+   - chutney's Travis CI tests (#)
+   - stem's Travis CI tests (#)
+
+2. Close the milestone in Trac. To do this, go to Trac, log in,
+   select "Admin" near the top of the screen, then select "Milestones" from
+   the menu on the left.  Click on the milestone for this version, and
+   select the "Completed" checkbox. By convention, we select the date as
+   the End of Life date.
```
dgoulet-tor: I would add a step or complement this with "Update wiki page": https://trac.torproject.org/projects/tor/wiki/org/teams/NetworkTeam/CoreTorReleases

Basically, move the milestone down into the EOL table.

PR #522, opened 2018-11-17
Bug #28485: Add missing library to build tor-print-ed-signing-cert.

ahf:
To succesful compile tor-print-ed-signing-cert.exe on Windows we
sometimes need to include the @TOR_LIB_GDI@ library.

See: https://bugs.torproject.org/28485


nmathewson:
cherry-picked to maint-0.3.5

PR #533, opened 2018-11-21
Adaptive padding prob distr tests

asn-d6:


src/test/test_prob_distr.c
```diff
@@ -0,0 +1,1197 @@
...
+		CHECK_RELERR(sf_logistic(x, 0, 1), sf_logistic(x/2, 0, .5));
+		CHECK_RELERR(icdf_logistic(p, 0, 1), icdf_logistic(p, 0,.5)*2);
+		CHECK_RELERR(isf_logistic(p, 0, 1), isf_logistic(p, 0, .5)*2);
+
+		/*
+		 * For p near 0 and p near 1/2, the arithmetic of
+		 * translating by 1 loses precision.
+		 */
+		if (fabs(p) > DBL_EPSILON && fabs(p) < 0.4) {
+			CHECK_RELERR(cdf_logistic(x, 0, 1),
```
asn-d6: Hmm why did we check the value of `p` if we are not using it at all it in the two checks below?

If you meant to use `p` instead of `x` in: `CHECK_RELERR(cdf_logistic(x, 0, 1), cdf_logistic(x*2 + 1, 1, 2));`
what does the value of `p` matter anyhow? Isn't this formula as true as `CHECK_RELERR(cdf_logistic(x, 0, 1), cdf_logistic(x*2, 0, 2));` for any `p`?

src/test/test_prob_distr.c
```diff
@@ -0,0 +1,1197 @@
...
+ */
+static void
+test_log_logistic(void *arg)
+{
+  (void) arg;
+
+	static const struct {
+		double x;
+		double p;
+		double np;
```
asn-d6: What does p and np stand for? Is it just random value placeholders to conduct the tests?

PR #537, opened 2018-11-22
Bug #28179, take two: Handle process input/output via event loop

ahf:
This is a rebased version of https://github.com/torproject/tor/pull/525 with fixes from @dgoulet-tor's review.

This is https://bugs.torproject.org/28179


src/feature/client/transports.c
```diff
@@ -520,34 +523,34 @@ proxy_prepare_for_restart(managed_proxy_t *mp)
...
-
-  process_environment_free(env);
-
-  if (retval == PROCESS_STATUS_ERROR) {
-    log_warn(LD_GENERAL, "Managed proxy at '%s' failed at launch.",
+  tor_assert(mp);
+
+  smartlist_t *env = create_managed_proxy_environment(mp);
+
+  /* Configure our process. */
```
dgoulet-tor: Are we missing a `process_set_protocol()` by any chance? As in, seems we are expecting "lines" or maybe the "raw" type is fine but either way, either doing it explicitly or comment explaining why we can't skip it?

src/lib/process/process_win32.c
```diff
@@ -474,28 +474,43 @@ process_win32_timer_callback(periodic_timer_t *timer, void *data)
...
+   * the call to process_win32_timer_test_process() returns true. */
+  bool done = true;
+
+  do {
+    const smartlist_t *processes = process_get_all_processes();
+
+    SMARTLIST_FOREACH_BEGIN(processes, process_t *, process) {
+      if (process_win32_timer_test_process(process)) {
+        done = false;
+        break;
```
dgoulet-tor: I looked and saw that if `true` is returned, then it means the exit callback was called thus the `process` was freed at that point.

1) I would comment out here that it is imperative that if `true` is returned, the loop must end (we can't continue in the list because a `smartlist_remove()` occurred.

2) Why is `done = false` when we break out of the smartlist loop and then the first thing that is done is `done = true` ?

ahf:
Closing.

PR #538, opened 2018-11-22
Initial manual page for tor-genkey

ppopth:
ticket18098 (review only)

### DIFF
--- a/doc/tor-genkey.1.txt
+++ b/doc/tor-genkey.1.txt
@@ -1,0 +1,48 @@
+// Copyright (c) The Tor Project, Inc.
+// See LICENSE for licensing information
+// This is an asciidoc file used to generate the manpage/html reference.
+// Learn asciidoc on http://www.methods.co.nz/asciidoc/userguide.html
+:man source:   Tor
+:man manual:   Tor Manual
+tor-genkey(1)
+==============
+Suphanat Chunhapanya
+
+NAME
+----
+tor-genkey - utility to generate various keys for tor
+
+SYNOPSIS
+--------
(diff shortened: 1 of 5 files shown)


doc/tor-genkey.1.txt
```diff
@@ -0,0 +1,48 @@
...
+-------
+**-t**, **--type** __keytype__::
+    Specify which key type you want to generate.
+
+**-o**, **--output** __filename__::
+    Output your key in __filename__.
+
+KEY TYPES
+---------
+**HsClientDescriptorAuth**::
```
dgoulet-tor: I propose we use `HS` instead of `Hs`.

Small fix, maybe: `HSDescriptorClientAuth` or `HSClientAuthDescriptor instead? Since this is "client authorization", we should just namespace it clearly so it is easy to understand/categorized. If we had let say the introduction authorization, then we would have also something like `HSClientAuthIntro` which starts the same way as the Descriptor one.

For usability, I would make it case insensitive in the implementation so probably also good to mention it in the manpage.

doc/tor-genkey.1.txt
```diff
@@ -0,0 +1,48 @@
...
+==============
+Suphanat Chunhapanya
+
+NAME
+----
+tor-genkey - utility to generate various keys for tor
+
+SYNOPSIS
+--------
+**tor-genkey** [-t|--type __keytype__] [-o|--out __filename__]
```
dgoulet-tor: Here is another thought now. What about a git alike style command?

`tor-genkey hs clientauth descriptor -o <filename>` type of command? (not final format, just to give an idea)

PR #542, opened 2018-11-25
Prop293

nmathewson:


src/feature/dirauth/dirvote.c
```diff
@@ -4620,7 +4620,8 @@ dirserv_generate_networkstatus_vote_obj(crypto_pk_t *private_key,
 
   v3_out->known_flags = smartlist_new();
   smartlist_split_string(v3_out->known_flags,
-                         "Authority Exit Fast Guard Stable V2Dir Valid HSDir",
+                         "Authority Exit Fast Guard Stable V2Dir Valid HSDir "
+                         "StaleDesc",
```
teor2345: Should this string be in alphabetical order for readability?
Does it matter?

nmathewson: It doesn't matter, since we sort the list later, but I agree that it helps clarity.  I'll add a commit to alphabetize it after my plane lands.

nmathewson:
vote_descriptors.inc and test_dir_common.c need to stay in sync with one another: the latter parses the former.

>  Can we automatically keep the flags in (some of) these files in sync?

fuzz_vrs.c should have a complete list, so it can fuzz all the flags; I've turned the list into a pair of constants to help keep them in sync.   (Will push once plane lands.)

Will work on tests.

src/test/test_voting_flags.c
```diff
@@ -0,0 +1,192 @@
...
+  .cleanup_fn = teardown_voting_flags_test,
+};
+
+#define T(name,flags)                                   \
+  { #name, test_voting_flags_##name, (flags), &voting_flags_setup, NULL }
+
+struct testcase_t voting_flags_tests[] = {
+  T(minimal, 0),
+  T(ipv6, TT_FORK),
+  // TODO: Add more of these tests.
```
teor2345: Can you open a ticket for the extra tests?
I'd do it, but I think you can describe what needs to be done much better.

PR #545, opened 2018-11-26
#28007: Address shellcheck warnings in scan-build.sh (2)

rl1987:
https://trac.torproject.org/projects/tor/ticket/28007

### DIFF
--- a/scripts/test/scan-build.sh
+++ b/scripts/test/scan-build.sh
@@ -54,6 +54,7 @@ else
    OUTPUTARG=""
 fi
 
+# shellcheck disable=SC2086
 scan-build \
     $CHECKERS \
     ./configure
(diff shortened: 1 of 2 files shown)


scripts/test/scan-build.sh
```diff
@@ -63,13 +64,15 @@ scan-build \
...
+# shellcheck disable=SC2086
 scan-build \
     $CHECKERS \
     -disable-checker deadcode.DeadStores \
     make -j5 -k ./src/ext/ed25519/ref10/libed25519_ref10.a
 
+# shellcheck disable=SC2086
 scan-build \
-    $CHECKERS $OUTPUTARG \
+    $CHECKERS "$OUTPUTARG" \
```
tlyu: Are we sure this is correct? The way we set `OUTPUTARG` above seems to depend on word splitting between the `-o` and `$SCAN_BUILD_OUTPUT`.

rl1987: Fixed in da264f7c766b332f596a92766f7625c4a17abf70.

PR #548, opened 2018-11-27
Explicitly specify path to OpenSSL for Appveyor.

ahf:
This patch explicitly specifies the path to our OpenSSL dependency and
disables the installation of an external OpenSSL version and instead
uses the OpenSSL version available from the MinGW environments.

See: https://bugs.torproject.org/28574


ahf:
Was merged as 36c32355349bff091b92c04be2823d96feebd5c6

PR #565, opened 2018-12-05
Call run_tor_main_loop() in ntmain.c, rather than do_main_loop().

nmathewson:
Fixes bug 28612; bugfix on 0.3.5.3-alpha.


nmathewson:
Squashed and merged

PR #566, opened 2018-12-05
[Tor Trac #27491]: Prefer IPv4 or IPv6 based on the number of failures

neelchauhan:
For the ticket [here](https://trac.torproject.org/projects/tor/ticket/27491).

### DIFF
--- a/src/app/config/or_state_st.h
+++ b/src/app/config/or_state_st.h
@@ -92,6 +92,10 @@ struct or_state_t {
   int IPv4AutoFail;
   int IPv6AutoFail;
 
+  /** When did we fail IPv4 or IPv6? */
+  time_t IPv4AutoFailTime;
+  time_t IPv6AutoFailTime;
+
   /** Count the number of failures for IPv4 and IPv6 respectively. */
   int IPv4Fails;
   int IPv6Fails;
--- a/src/app/config/or_state_st.h
+++ b/src/app/config/or_state_st.h
@@ -87,6 +87,14 @@ struct or_state_t {
 
   /** When did we last rotate our onion key?  "0" for 'no idea'. */
   time_t LastRotatedOnionKey;
+
+  /** Do we automatically fail IPv4 or IPv6? If so, set one of these to 1. */
+  int IPv4AutoFail;
+  int IPv6AutoFail;
+
+  /** Count the number of failures for IPv4 and IPv6 respectively. */
+  int IPv4Fails;
+  int IPv6Fails;
 };
 
 #endif
--- a/src/app/config/statefile.c
+++ b/src/app/config/statefile.c
@@ -133,8 +133,10 @@ static config_var_t state_vars_[] = {
   VAR("BuildtimeHistogram",           LINELIST_V, BuildtimeHistogram, NULL),
 
   V(IPv4AutoFail,                     UINT,     "0"),
-  V(IPv6AutoFail,                     UINT,     "0"),
+  V(IPv4AutoFailTime,                 ISOTIME,  NULL),
   V(IPv4Fails,                        UINT,     "1"),
+  V(IPv6AutoFail,                     UINT,     "0"),
+  V(IPv6AutoFailTime,                 ISOTIME,  NULL),
   V(IPv6Fails,                        UINT,     "1"),
 
   END_OF_CONFIG_VARS
--- a/src/app/config/statefile.c
+++ b/src/app/config/statefile.c
@@ -132,6 +132,11 @@ static config_var_t state_vars_[] = {
   VAR("CircuitBuildTimeBin",          LINELIST_S, BuildtimeHistogram, NULL),
   VAR("BuildtimeHistogram",           LINELIST_V, BuildtimeHistogram, NULL),
 
+  V(IPv4AutoFail,                     UINT,     "0"),
+  V(IPv6AutoFail,                     UINT,     "0"),
+  V(IPv4Fails,                        UINT,     "1"),
+  V(IPv6Fails,                        UINT,     "1"),
+
   END_OF_CONFIG_VARS
 };
 
--- a/src/core/or/connection_or.c
+++ b/src/core/or/connection_or.c
@@ -36,6 +36,8 @@
 #include "core/or/circuitstats.h"
 #include "core/or/command.h"
 #include "app/config/config.h"
+#include "app/config/or_state_st.h"
+#include "app/config/statefile.h"
 #include "core/mainloop/connection.h"
 #include "core/or/connection_or.h"
 #include "feature/control/control.h"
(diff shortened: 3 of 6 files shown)


src/app/config/or_state_st.h
```diff
@@ -87,6 +87,18 @@ struct or_state_t {
 
   /** When did we last rotate our onion key?  "0" for 'no idea'. */
   time_t LastRotatedOnionKey;
+
+  /** Do we automatically fail IPv4 or IPv6? If so, set one of these to 1. */
+  int IPv4AutoFail;
```
teor2345: This name is confusing:
* the variable actually tracks if we have ever failed an IPv4 connection because there was no route to a host.
* but maybe it should track the number of failed connections due to no route to host.

src/app/config/or_state_st.h
```diff
@@ -87,6 +87,18 @@ struct or_state_t {
...
+  int IPv6AutoFail;
+
+  /** When did we fail IPv4 or IPv6? */
+  time_t IPv4AutoFailTime;
```
teor2345: This name is confusing:
* the variable actually tracks the earliest time in the last week that we failed an IPv4 connection because there was no route to a host.
* but maybe we should track this information over smaller time periods, so we can forget the earliest failures more regularly.

src/app/config/or_state_st.h
teor2345: We should also think about the privacy implications of storing exact times and exact failure numbers on clients.

src/core/or/connection_or.c
```diff
@@ -1364,10 +1366,34 @@ void
...
 {
+  const or_options_t *options = get_options();
+
   control_event_or_conn_status(conn, OR_CONN_EVENT_FAILED, reason);
-  if (!authdir_mode_tests_reachability(get_options()))
+  if (!authdir_mode_tests_reachability(options))
     control_event_bootstrap_prob_or(msg, reason, conn);
   note_or_connect_failed(conn);
+
+  if (options->ClientAutoIPv6ORPort) {
```
teor2345: This option check is good for an experimental feature. When the feature stops being experimental, we might always want to keep this state, even if the option is off.

src/core/or/connection_or.c
```diff
@@ -1364,10 +1366,34 @@ void
...
+    or_state_t *state = get_or_state();
+    switch (tor_addr_family(&conn->real_addr)) {
+      case AF_INET:
+        if (reason == END_OR_CONN_REASON_NO_ROUTE) {
```
teor2345: We don't want to turn off an IP family just because our routing table doesn't have an entry for one IP address. Maybe we should track the number of failed connections due to no route to host.

src/core/or/policies.c
```diff
@@ -495,8 +500,39 @@ fascist_firewall_prefer_ipv6_impl(const or_options_t *options)
 MOCK_IMPL(int,
 fascist_firewall_rand_prefer_ipv6_addr, (void))
 {
-  /* TODO: Check for failures, and infer our preference based on this. */
-  return crypto_rand_int(2);
+  or_state_t *state = get_or_state();
+  time_t current_time = time(NULL);
+
+  /* Clear IPv4 and IPv6 failure entries if they are more than a week old. */
+  if (state->IPv4AutoFailTime &&
```
teor2345: Maybe we should track this information over smaller time periods, so we can forget the earliest failures more regularly. rephist.c has some code for tracking state over multiple time periods.

src/core/or/policies.c
```diff
@@ -495,8 +500,39 @@ fascist_firewall_prefer_ipv6_impl(const or_options_t *options)
...
+    state->IPv4AutoFail = 0;
+    state->IPv4AutoFailTime = 0;
+  }
+  if (state->IPv6AutoFailTime &&
+      state->IPv6AutoFailTime <= current_time - IP_FAIL_CLEAR_TIME) {
+    state->IPv6AutoFail = 0;
+    state->IPv6AutoFailTime = 0;
+  }
+
+  if (state->IPv4AutoFail && !state->IPv6AutoFail) {
```
teor2345: We don't want to turn off an IP family just because our routing table doesn't have an entry for one IP address. Maybe we should track the number of failed connections due to no route to host, and give them a higher weight.

src/core/or/policies.c
```diff
@@ -495,8 +500,39 @@ fascist_firewall_prefer_ipv6_impl(const or_options_t *options)
...
+  } else if (!state->IPv4AutoFail && state->IPv6AutoFail) {
+    /* If we fail IPv6 automatically, prefer IPv4. */
+    return 0;
+  } else if (state->IPv4AutoFail && state->IPv6AutoFail) {
+    /* If we fail both IPv4 and IPv6, reset the auto fail counters and then
+     * select randomly. */
+    state->IPv4AutoFail = state->IPv6AutoFail = 0;
+  }
+
+  int fail_sum = state->IPv4Fails + state->IPv6Fails;
```
teor2345: Adding two unsigned ints can overflow a signed int. Instead, use uint32_t and tor_add_u32_nowrap().

src/core/or/connection_or.c
```diff
@@ -1364,10 +1366,34 @@ void
...
+          state->IPv4AutoFail = 1;
+          if (!state->IPv4AutoFailTime)
+            state->IPv4AutoFailTime = time(NULL);
+        }
+        state->IPv4Fails++;
```
teor2345: Incrementing an int can overflow. Instead, use uint32_t and tor_add_u32_nowrap().

src/core/or/connection_or.c
```diff
@@ -1364,10 +1366,34 @@ void
...
+        break;
+      case AF_INET6:
+        if (reason == END_OR_CONN_REASON_NO_ROUTE) {
+          state->IPv6AutoFail = 1;
+          if (!state->IPv6AutoFailTime)
+            state->IPv6AutoFailTime = time(NULL);
+        }
+        state->IPv6Fails++;
```
teor2345: Incrementing an int can overflow. Instead, use uint32_t and tor_add_u32_nowrap().

src/app/config/statefile.c
```diff
@@ -132,6 +132,13 @@ static config_var_t state_vars_[] = {
   VAR("CircuitBuildTimeBin",          LINELIST_S, BuildtimeHistogram, NULL),
   VAR("BuildtimeHistogram",           LINELIST_V, BuildtimeHistogram, NULL),
 
+  V(IPv4AutoFail,                     UINT,     "0"),
```
teor2345: The state file variables are unsigned int, but the struct is int. Use uint32_t for both.

src/app/config/or_state_st.h
```diff
@@ -87,6 +87,18 @@ struct or_state_t {
...
+  time_t IPv6AutoFailTime;
+
+  /** Count the number of failures for IPv4 and IPv6 respectively. */
+  int IPv4Fails;
```
teor2345: The state file variables are unsigned int, but the struct is int. Use uint32_t for both.

src/core/or/policies.c
```diff
@@ -87,6 +89,9 @@ static const char *private_nets[] = {
   NULL
 };
 
+/** Time when we should clear IPv4/IPv6 failure entries. */
+#define IP_FAIL_CLEAR_TIME 60 * 60 * 24 * 7
```
teor2345: This time is far too long for mobile devices and laptops, which can change networks every few seconds. It's also far too long for typical users, who give up after 30 seconds.

src/app/config/or_state_st.h
teor2345: I opened a ticket for automatic checks that could avoid these kinds of issues in future:
https://trac.torproject.org/projects/tor/ticket/29115#ticket

src/core/or/policies.c
```diff
@@ -495,8 +500,39 @@ fascist_firewall_prefer_ipv6_impl(const or_options_t *options)
...
+    /* If we fail IPv6 automatically, prefer IPv4. */
+    return 0;
+  } else if (state->IPv4AutoFail && state->IPv6AutoFail) {
+    /* If we fail both IPv4 and IPv6, reset the auto fail counters and then
+     * select randomly. */
+    state->IPv4AutoFail = state->IPv6AutoFail = 0;
+  }
+
+  int fail_sum = state->IPv4Fails + state->IPv6Fails;
+  int chosen_number = crypto_rand_int(fail_sum);
```
teor2345: Asking for up to 32 bits of entropy for each new circuit could significantly slow down busy clients. I think that @nmathewson has made our PRNG fast enough to handle this kind of load.

But we might want to think about this problem differently.
Let's work out an acceptable failure rate for connections, for example, 1/8.
Then, we reduce the probability of IPv4/IPv6 to one of a small set of probabilities: 1/8, 2/8, 3/8, ... , 7/8.
(Deliberately excluding 0/8 and 8/8.)
This design is better, because it doesn't starve clients that have just changed networks.

PR #583, opened 2018-12-12
23588: Stop ignoring IPv6 link specifiers sent to v3 onion services

teor2345:
v3 onion service IPv6 support is still incomplete, see 23493 for details.


teor2345:
Rebased to master in #973.

PR #590, opened 2018-12-14
Stop re-checking our hardcoded dh parameters on every startup [28851]

nmathewson:
Closes ticket 28851.


nmathewson:
Squashed and merged

PR #599, opened 2018-12-17
Always initialize addr in parse_port_config()

nmathewson:
It was always analyzed before use, but scan-build wasn't able to
persuade itself of that.

Closes ticket 28881.


nmathewson:
Merged a version with a typo fix in the changes file

PR #608, opened 2018-12-20
#28179 win32 robustness

ahf:
I tried to run the test cases for process_t related code in a VM with limited amount of memory and where I also ran a "load heavy" task (compiled Tor with -jN in a while true loop) and made some discoveries:

1. The exit handler would sometimes be called *before* we received the "Broken Pipe" event from `ReadFileEx()`. We now wait until we have received EOF state for both standard out and error before we check whether the process have terminated.
2. We forgot to clean up the child process' end of the pipes after the process have been created.
3. There is no need to log ordinary "end of process" conditions as "warning" log severity.
4. We forgot to set `reached_eof` to true under certain pathological conditions. Together with (1) this could sometimes lead to an error in the slow tests where the process was terminated before all output from the child process had been received by Tor's event loop.
5. Add some additional error checking around ReadFileEx() and WriteFileEx() which is recommended by the MSDN API documentation.

To run a test in PowerShell until it fails one can do:

`while ($true) { .\test-slow.exe slow/process/... ; if ($LastExitCode -ne 0) { break : } }`

### DIFF
--- a/src/lib/process/process_win32.c
+++ b/src/lib/process/process_win32.c
@@ -720,30 +720,23 @@ process_win32_cleanup_handle(process_win32_handle_t *handle)
   tor_assert(handle);
 
 #if 0
-  /* FIXME(ahf): My compiler does not set _WIN32_WINNT to a high enough value
-   * for this code to be available. Should we force it? CancelIoEx() is
-   * available from Windows 7 and above. If we decide to require this, we need
-   * to update the checks in all the three I/O completion callbacks to handle
-   * the ERROR_OPERATION_ABORTED as well as ERROR_BROKEN_PIPE. */
-
-#if _WIN32_WINNT >= 0x0600
-  /* This code is only supported from Windows 7 and onwards. */
   BOOL ret;
   DWORD error_code;
 
-  /* Cancel any pending I/O requests. */
-  ret = CancelIoEx(handle->pipe, &handle->overlapped);
+  /* Cancel any pending I/O requests: This means that instead of getting
+   * ERROR_BROKEN_PIPE we get ERROR_OPERATION_ABORTED, but it doesn't seem
+   * like this is needed. */
+  ret = CancelIo(handle->pipe);
 
   if (! ret) {
     error_code = GetLastError();
 
     /* There was no pending I/O requests for our handle. */
     if (error_code != ERROR_NOT_FOUND) {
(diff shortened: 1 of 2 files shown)


src/lib/process/process_win32.c
```diff
@@ -525,6 +525,28 @@ process_win32_timer_test_process(process_t *process)
...
+   * This ensures that we *never* call the exit callback of the `process_t`,
+   * which potentially ends up calling `process_free()` on our `process_t`,
+   * before all data have been received from the process.
+   *
+   * We do NOT have a check here for whether standard in reached EOF since
+   * standard in's WriteFileEx() function is only called on-demand when we have
+   * something to write and is thus usually not awaiting to finish any
+   * operations. If we WriteFileEx() to a file that has terminated we'll simply
+   * get an error from ReadFileEx() or its completion routine and move on with
+   * life.  */
```
nmathewson: Is there any possibility that these checks might prevent a dead process from being freed forever?  I'm guessing no, but I feel like I should ask.

src/lib/process/process_win32.c
```diff
@@ -363,6 +370,24 @@ process_win32_write(struct process_t *process, buf_t *buffer)
     return 0;
   }
 
+  /* Here be dragons: According to MSDN's documentation for WriteFileEx() we
+   * should check GetLastError() after a call to WriteFileEx() even though the
+   * `ret` return value was successful. If everything is good, GetLastError()
+   * returns `ERROR_SUCCESS` and nothing happens.
+   *
```
nmathewson: Wow, what a horrible API they have!  But I guess we can't do anything about that.  (I wonder if it is intentional, or whether they had a bug and just documented it as the official behavior.)

src/lib/process/process_win32.c
ahf: I haven't been able to see that happen: even with processes that don't initialize properly (missing a .DLL for example) it seems like the console handles gets closed properly upon termination. I wanted to document this though such that it makes more sense for people looking at the code in the future.

src/lib/process/process_win32.c
ahf: I agree. In the documentation they mostly talk about buffer size issues, but I think that is only happening if you do I/O where you have multiple async I/O calls out at the same time that works on the same overlapped structure.

PR #611, opened 2018-12-20
Ticket28913: Make base32_decode() return number of bytes written.

nmathewson:


src/feature/rend/rendcache.c
```diff
@@ -854,7 +854,7 @@ rend_cache_store_v2_desc_as_client(const char *desc,
     *entry = NULL;
   }
   if (base32_decode(want_desc_id, sizeof(want_desc_id),
-                    desc_id_base32, strlen(desc_id_base32)) != 0) {
+                    desc_id_base32, strlen(desc_id_base32)) < 0) {
```
dgoulet-tor: I think we should be looking at `!= REND_DESC_ID_V2_LEN_BASE32` here because with `< 0` we could end up continuing with shorter or longer descriptor ID after which is undesirable I would say? (In reality, I don't think `base32_decode` can return a smaller size but for sake of correctness).

dgoulet-tor: If we go that way, this means that every places that does `< 0` should then check for the expected length and not only an explicit error (-1). Decoding something that we expect 60 bit and getting 58 bit is imo an error.

nmathewson: hm. that's probably right, but it's also a bunch more work. I don't suppose you'd like to write that patch maybe? ;)  if not, I'll get to it when I can

PR #612, opened 2018-12-20
#28847 implementation

ahf:
These patches makes Tor use the subsystem list for initializing and shutting down the process module.


nmathewson:
squashed and merged!

PR #613, opened 2018-12-21
ticket27167 Orconn tracker

tlyu:


src/feature/control/btrack_orconn.c
```diff
@@ -0,0 +1,196 @@
...
+ * Update a cached state of a best ORCONN progress we've seen so far.
+ *
+ * Return true if the new state is better than the old.
+ **/
+static bool
+bto_update_best(const bt_orconn_t *bto, bto_best_t *best, const char *type)
+{
+  if (bto->state < best->state)
+    return false;
+  best->gid = bto->gid;
```
nmathewson: Why do we make this change unconditionally?

tlyu: It might seem somewhat arbitrary at first. I think it's better to make the update because even if the best state stays the same, an additional ORCONN transitioning to that state is more recent information than we previously had.

I can add a comment to that effect if you would like.

src/feature/control/control_bootstrap.c
```diff
@@ -33,9 +33,24 @@ static const struct {
 } boot_to_str_tab[] = {
   { BOOTSTRAP_STATUS_UNDEF, "undef", "Undefined" },
   { BOOTSTRAP_STATUS_STARTING, "starting", "Starting" },
-  { BOOTSTRAP_STATUS_CONN_DIR, "conn_dir", "Connecting to directory server" },
```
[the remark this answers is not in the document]
tlyu: I _think_ this is an exhaustive enumeration; please let me know what you think. We might want to put this information somewhere less ephemeral than a pull request comment, though. I'm not sure where is best for it.
* `handshake`: repurposed to indicate handshaking with any relay; previously it only occurred internally, to be decoded by `control_event_bootstrap()` into `handshake_dir` and `handshake_or`
* `undef` and `starting`: these remain the same
* `conn_dir`: removed; I would say we don't actually want this, because at this part of the bootstrap process, it's more useful to let the user know as much information about the state of connectivity to any relay than to let the user know about connectivity to directory servers in particular
* `handshake_dir`: removed; similar to `conn_dir`; also potentially misleading because it could be for an application circuit, because of the way we previously decoded the `handshake` status in `control_event_bootstrap()`
* `onehop_create`: this remains, but is renumbered
* `requesting_status`, `loading_status`, `loading_keys`, `requesting_descriptors`, and `loading_descriptors`: these remain, though some are renumbered
* `conn_or`: replaced with `ap_conn`; previously, the directory code would report this as soon as it thought it had enough directory info to build a circuit, regardless of whether tor was actually about to (and able to) launch a connection to build circuits. I guess we could reuse this, but I like the `ap_conn` systematic naming of tags better.
* `handshake_or`: replaced with `ap_handshake`; previously, this would be for any handshake that started after we get enough directory info, regardless of whether it was for a directory server or an application
circuit (see previous comments about decoding status). I guess we could reuse this too, but see previous comments about `conn_or`
* `circuit_create`: this remains, but is renumbered
* `done`: this remains the same
Do we know which applications depend on specific meanings of the old tag names? Or do they use them to look up localized summary text? (I think Tor Launcher does this, falling back to the `SUMMARY=` text)

src/feature/control/btrack_orconn.c
nmathewson: Yes; thank you!

src/feature/control/control_bootstrap.c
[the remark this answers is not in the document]
nmathewson: I'm afraid I don't know which applications are . But this is very helpful information, and will be useful to the integrators.

Is there a patch for control-spec.txt?  It would be good to merge that at the same time as this, so the integrators can start hacking once we merge.

PR #621, opened 2018-12-26
#20006: Allow for client auth when responding to HSFETCH

rl1987:
Until now Tor could not respond to HSFETCH command properly if
target Onion Service had client auth enabled. Now it checks if
we have an entry for requested address in our map of configured auth
cookies. If one is found, auth cookie is used to request HSDesc.

https://trac.torproject.org/projects/tor/ticket/20006

### DIFF
--- a/src/feature/control/control.c
+++ b/src/feature/control/control.c
@@ -4473,7 +4473,7 @@ handle_control_hsfetch(control_connection_t *conn, uint32_t len,
   rend_auth_type_t auth_type = REND_NO_AUTH;
   rend_service_authorization_t *auth = NULL;
 
-  char *cookie = NULL;
+  const char *cookie = NULL;
 
   if (hsaddress) {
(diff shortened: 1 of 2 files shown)


src/feature/control/control.c
```diff
@@ -4470,8 +4470,23 @@ handle_control_hsfetch(control_connection_t *conn, uint32_t len,
     }
   }
 
-  rend_query = rend_data_client_create(hsaddress, desc_id, NULL,
-                                       REND_NO_AUTH);
+  rend_auth_type_t auth_type = REND_NO_AUTH;
+  rend_service_authorization_t *auth = NULL;
+
+  char *cookie = NULL;
```
dgoulet-tor: This should be `const`.

rl1987: Fixed in b93f4140f6258d572a9635b06dbfdf7619608b93.

PR #623, opened 2018-12-27
Adaptive padding 20181227

mikeperry-tor:


asn-d6:
This is the unsquashed equivalent of https://github.com/torproject/tor/pull/624

PR #624, opened 2019-01-02
Adaptive padding final

asn-d6:


src/lib/smartlist_core/smartlist_foreach.h
```diff
@@ -83,6 +83,14 @@
          ++var ## _sl_idx) {                                    \
       var = (sl)->list[var ## _sl_idx];
 
+#define SMARTLIST_FOREACH_REVERSE_BEGIN(sl, type, var)  \
```
nmathewson: I think this will be incompatible with SMARTLIST_DEL_CURRENT; we should maybe add a comment?

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,685 @@
...
+/* See LICENSE for licensing information */
+
+/**
+ * \file circuitpadding.h
+ * \brief Header file for circuitpadding.c.
+ **/
+#ifndef TOR_CIRCUITPADDING_H
+#define TOR_CIRCUITPADDING_H
+
+#include "circpad_negotiation.h"
```
nmathewson: We try to include stuff like this by full path; can fix post merge though.

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,685 @@
...
+} circpad_decision_t;
+
+/** The type for the things in histogram bins (aka tokens) */
+typedef uint32_t circpad_hist_token_t;
+
+/** The type for histogram indexes (needs to be negative for errors) */
+typedef int8_t circpad_hist_index_t;
+
+/** The type for absolute time, from monotime_absolute_usec() */
+typedef uint64_t circpad_time_t;
```
nmathewson: It is probably faster to use something else if we are going to be calling this a lot; the non-coarse monotime functions are slower than the coarse ones, and converting to usec sometimes takes an expensive division on 32-bit platforms.

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,685 @@
+/*
+ * Copyright (c) 2017, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
+ * \file circuitpadding.h
+ * \brief Header file for circuitpadding.c.
```
nmathewson: This could use a comment that explains WTF-PAD 101 for beginners, unless such a comment will appear somewhere else.  Most of the stuff below in this file only makes sense if you know how the rest of wtfpad works.

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,685 @@
...
+#define CIRCPAD_PURPOSE_ALL (0xFFFFFFFF)
+
+/**
+ * This type specifies all of the conditions that must be met before
+ * a client decides to initiate padding on a circuit.
+ *
+ * A circuit must satisfy every sub-field in this type in order
+ * to be considered to match the conditions.
+ */
+typedef struct circpad_machine_conditions_t {
```
nmathewson: Are all of these structures are stuff that users of the wtf-pad api will need, or are some of them internal only, or only used in a few places?  In the latter case, we now use a "foo_st.h" header for them.

src/core/or/circuitpadding.h
```diff
@@ -0,0 +1,685 @@
...
+ * machine. The mutable information must be kept separate because
+ * it exists per-circuit, where as the machines themselves are global.
+ * This separation is done to conserve space in the circuit structure.
+ *
+ * This is the per-circuit state that changes regarding the global state
+ * machine. Some parts of it are optional (ie NULL).
+ *
+ * XXX: Play with layout to minimize space on x64 Linux (most common relay).
+ */
+typedef struct circpad_machineinfo_t {
```
nmathewson: It still seems to me like the name of this structure and the name of the ones above might be swapped, but I won't insist.

src/core/or/circuit_st.h
```diff
@@ -177,6 +182,25 @@ struct circuit_t {
...
       circuitmap. */
   HT_ENTRY(circuit_t) hs_circuitmap_node;
+
+  /** Adaptive Padding state machines: these are immutable. The state machines
+   *  that come from the consensus are saved to a global structure, to avoid
+   *  per-circuit allocations. This merely points to the global copy.
+   *
+   *  Each element of this array corresponds to a different padding machine,
+   *  and we can have up to CIRCPAD_MAX_MACHINES such machines. */
+  const circpad_machine_t *padding_machine[CIRCPAD_MAX_MACHINES];
```
nmathewson: We need to make sure that the global structure itself never changes somewhere, and is never allocated or deallocated, right?

src/core/or/circuit_st.h
```diff
@@ -177,6 +182,25 @@ struct circuit_t {
...
+  /** Adaptive Padding machine info for above machines. This is the
+   *  per-circuit mutable information, such as the current state and
+   *  histogram token counts. Some of it is optional (aka NULL).
+   *  If a machine is being shut down, these indexes can be NULL
+   *  without the corresponding padding_machine being NULL, while we
+   *  wait for the other end to respond to our shutdown request.
+   *
+   *  Each element of this array corresponds to a different padding machine,
+   *  and we can have up to CIRCPAD_MAX_MACHINES such machines. */
+  circpad_machineinfo_t *padding_info[CIRCPAD_MAX_MACHINES];
```
nmathewson: I think this is okay so long as CIRCPAD_MAX_MACHINES stays small.

src/core/or/relay.c
```diff
@@ -524,6 +526,7 @@ relay_command_to_string(uint8_t command)
     case RELAY_COMMAND_INTRODUCE_ACK: return "INTRODUCE_ACK";
     case RELAY_COMMAND_EXTEND2: return "EXTEND2";
     case RELAY_COMMAND_EXTENDED2: return "EXTENDED2";
+    case RELAY_COMMAND_PADDING_NEGOTIATE: return "PADDING_NEGOTIATE";
```
nmathewson: Does this also need a case for PADDING_NEGOTIATED?

src/core/or/circuitpadding.c
```diff
@@ -0,0 +1,2467 @@
...
+
+  if (state->use_rtt_estimate)
+    start_usec = mi->rtt_estimate_usec+state->start_usec;
+  else
+    start_usec = state->start_usec;
+
+  if (usec <= start_usec)
+    return 0;
+
+  if (usec == start_usec+1)
```
nmathewson: Why does this need to be a special case?

nmathewson: (Maybe a comment would help)

PADDING_TODO.txt
```diff
@@ -0,0 +1,58 @@
+TODO sketch for this branch, in rough priority order:
```
nmathewson: Does this file get removed later on in the branch?  If not, then the commit message suggests we have an issue here. :)

doc/tor.1.txt
```diff
@@ -1020,6 +1020,21 @@ The following options are useful only for clients (that is, if
     The .exit address notation, if enabled via MapAddress, overrides
     this option.
 
+[[MiddleNodes]] **MiddleNodes** __node__,__node__,__...__::
```
nmathewson: If this is going to get merged, we should mark it as experimental-only for some reason, and tell people what they should use instead for non-experimental use.  We may also want to say explicitly that it might go away in the future.

src/lib/crypt_ops/crypto_rand.c
```diff
@@ -528,6 +528,17 @@ crypto_rand_unmocked(char *to, size_t n)
 #endif
 }
 
+/**
+ * Draw an unsigned 32-bit integer uniformly at random.
+ */
+uint32_t
+crypto_rand_uint32(void)
```
nmathewson: This function's behavior is inconsistent with crypto_rand_uint64(), which takes an argument. Maybe it should have a different name, like, crypto_rand_u32()?  Or maybe we should rename the other one?

src/lib/math/.may_include
```diff
@@ -3,3 +3,5 @@ orconfig.h
 lib/cc/*.h
 lib/log/*.h
 lib/math/*.h
+lib/testsupport/*.h
+lib/crypt_ops/*.h
```
nmathewson: This change raises lib/math to a higher level than it was at before, which means that these libraries will no longer be able to use it: container, trace, thread, term, memarea, encoding, sandbox, fs, meminfo, crypt_os, and time.  I think that's okay?

src/lib/math/fp.c
```diff
@@ -117,3 +117,28 @@ ENABLE_GCC_WARNING(double-promotion)
 ENABLE_GCC_WARNING(float-conversion)
 #endif
 }
+
+/* isinf() wrapper for tor */
+int
+tor_isinf(double x)
```
nmathewson: This could use a test.

src/lib/smartlist_core/smartlist_foreach.h
asn-d6: A comment about this was added in a336d816a68e5eaddd9d80f7179699274b367a1d.

src/core/or/circuitpadding.c
asn-d6: Good question. I'm not sure why this is the case.

Originally I thought it was so that we dont have a div-by-zero in
```
  bin = CIRCPAD_INFINITY_BIN(state) -
    tor_log2(2*histogram_range_usec/(usec-start_usec+1));
```
but the denominator there is different from what that check checks.

Mike, comment?

PADDING_TODO.txt
asn-d6: Mike what should we do with this file in your opinion?

src/core/or/circuitpadding.h
asn-d6: Fixed in 2b29bccb85

src/core/or/circuitpadding.h
asn-d6: Hmm, not quite sure how to fix this.

Not sure if using coarse time sources is acceptable here given that padding delays are can be pretty small and might need to be accurate.

Any ideas?

src/core/or/circuitpadding.h
asn-d6: Done in cdbd5c0af8.

src/core/or/circuitpadding.h
asn-d6: I think we can do this, yes. 

Folded this idea in https://trac.torproject.org/projects/tor/ticket/28777#comment:3 so that we can do it post-merge, if you think that's acceptable.

src/core/or/circuitpadding.h
asn-d6: OK, this might be a good moment to do this since there are not many pending branches on top of this right now: 5738a0ab6c f4938179c5

Doing this will also close ticket #28778.

#28778
--

src/core/or/circuit_st.h
nmathewson: Maybe add comments here and in the code that sets up the foo_padding_machines?

src/core/or/circuitpadding.h
nmathewson: So, how small are we talking here?  Libevent won't reliably give us a timeout more fine-grained than tens of milliseconds on many platforms.

Even if coarse time isn't workable, we should at least figure out how to minimize how often this is called, and try to use monotime_t.  But I think coarse time or timestamp-units is a better idea, probably.

Maybe the right answer is to define an abstraction that can be one or the other, and compile it conditionally?

src/core/or/circuit_st.h
asn-d6: Done in 94c33eabc7.

src/core/or/circuitpadding.h
asn-d6: This was discussed and resolved in the trac ticket: https://trac.torproject.org/projects/tor/ticket/28142#comment:35

We should fix this for good before enabling any machines.

PADDING_TODO.txt
asn-d6: I removed the TODO file in b269ab5 and added the remaining items as children of #28632.

PR #625, opened 2019-01-02
Detect openssl bug 7712 and work around it.

nmathewson:
In theory it would be better to detect this bug in advance, but this
approach is much simpler, and therefore safer to backport.

This closes tor issue 28973.


nmathewson:
Squashed and merged

PR #629, opened 2019-01-04
Actually close stdout pipe on error in process_unix_exec

katterjohn:
https://trac.torproject.org/projects/tor/ticket/28994


nmathewson:
Cherry-picked 136ebc4 to master. Thanks!

PR #630, opened 2019-01-04
Bug28780

asn-d6:


src/core/or/circuitlist.c
```diff
@@ -2199,6 +2199,11 @@ circuit_mark_for_close_, (circuit_t *circ, int reason, int line,
   tor_assert(line);
   tor_assert(file);
 
+  /* Check whether the circuitpadding subsystem wants to block this close */
+  if (!circpad_circuit_should_be_marked_for_close(circ)) {
+    return;
+  }
```
mikeperry-tor: Should we be marking the circuit dirty and/or setting other flags that say "don't attach new streams here"? We don't want to change behavior and cause circuits that really should be not used to get used.

src/core/or/circuitpadding.c
```diff
@@ -1199,6 +1257,17 @@ circpad_machine_transitioned_to_end(circpad_machineinfo_t *mi)
...
 
+  /** If this machine handles cirucit lifetime on its own, shut the circuit
+   *  down now if we must.
+   *
+   *  XXX We ignore should_negotiate_end in this case */
+  if (machine->manage_circ_lifetime) {
+    circuit_t *on_circ = mi->on_circ;
+    if (mi->circuit_was_asked_to_be_closed) {
+      circuit_mark_for_close(on_circ, END_CIRC_REASON_NONE);
+    }
```
mikeperry-tor: The XXX says "We ignore should_negotiate_end in this case" but we don't return here. I kinda think we should return here and not negotiate the end. Otherwise, circuits torn down due to circpad can have an end cell sent before they are closed, which seems like a (minor) information leak.

src/core/or/circuitlist.c
mikeperry-tor: Related: We should check the reason flag to see if this is a remote reason or some other condition that indicates the other side closed the circuit on us, or it is unusable. (This may require some grepping to verify ciircuit_mark_for_close() calls).

src/core/or/circuitpadding.c
asn-d6: Ye I should return here. That's a bug.

PR #633, opened 2019-01-05
#28989: Fix CID 1442277

rl1987:
https://trac.torproject.org/projects/tor/ticket/28989

### DIFF
--- a/src/test/test_hs_service.c
+++ b/src/test/test_hs_service.c
@@ -267,7 +267,7 @@ helper_clone_authorized_client(const hs_service_authorized_client_t *client)
 
 /* Helper: Return a newly allocated service object with the identity keypair
  * sets and the current descriptor. Then register it to the global map.
- * Caller should us hs_free_all() to free this service or remove it from the
+ * Caller should use hs_free_all() to free this service or remove it from the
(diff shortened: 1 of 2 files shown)


nmathewson:
Should the helper be called in the "done" block, maybe?  Remember that every tt_* macro will conditionally call "goto done" on failure, so doing cleanup at other points in the function is not necessarily useful.

src/test/test_hs_service.c
```diff
@@ -289,6 +289,20 @@ helper_create_service(void)
   return service;
 }
 
+/* Helper: Deallocate a given service object, it's child objects and
```
nmathewson: s/it's/its/

rl1987: Fixed - 151ce23dc22383ebc754931e288bc2ad604cece9.

rl1987:
In 0f0ad3c2530c62d94b339f6c752effa2d80c8443 I keep track of the last `hs_service_t` struct that was created in one of these inner lexical scopes and deallocate it in case one of the asserts failed. I suppose this lets Coverity realise that we deallocate it either way?

nmathewson:
squashed and merged

PR #634, opened 2019-01-07
#24953: In updateFallbackDirs.py, say 'fallback list' instead of 'whitelist'

rl1987:
https://trac.torproject.org/projects/tor/ticket/24953

### DIFF
--- a/changes/bug24953
+++ b/changes/bug24953
@@ -1,0 +1,3 @@
+  o Minor bugfixes (fallback scripts):
+    - In updateFallbackDirs.py, say "fallback list" instead of "whitelist".
+      Fixes bug 24953; bugfix on 0.3.5.1-alpha.
--- a/scripts/maint/updateFallbackDirs.py
+++ b/scripts/maint/updateFallbackDirs.py
@@ -1594,7 +1594,7 @@ class CandidateList(dict):
     """ Apply the fallback whitelist_obj to this fallback list,
         passing exact to is_in_whitelist(). """
     excluded_count = 0
-    logging.debug('Applying whitelist')
+    logging.debug('Applying fallback list')
     # parse the whitelist
     whitelist = self.load_relaylist(whitelist_obj)
     filtered_fallbacks = []
@@ -1609,14 +1609,14 @@ class CandidateList(dict):
       else:
           # exclude
           excluded_count += 1
-          log_excluded('Excluding %s: not in whitelist.',
+          log_excluded('Excluding %s: not in fallback list.',
                        f._fpr)
     self.fallbacks = filtered_fallbacks
     return excluded_count
 
   @staticmethod
   def summarise_filters(initial_count, excluded_count):
-    return '/* Whitelist excluded %d of %d candidates. */'%(
+    return '/* Fallback list excluded %d of %d candidates. */'%(
                                                 excluded_count, initial_count)
 
   # calculate each fallback's measured bandwidth based on the median


changes/bug24953
```diff
@@ -0,0 +1,3 @@
+  o Minor bugfixes (fallback scripts):
+    - In updateFallbackDirs.py, say "fallback list" instead of "whitelist".
```
teor2345:
```suggestion
    - In updateFallbackDirs.py, call the filter file a "fallback list" instead of a "whitelist" in check_existing mode.
```

changes/bug24953
```diff
@@ -0,0 +1,3 @@
...
+      Fixes bug 24953; bugfix on 0.3.5.1-alpha.
```
teor2345: check_existing mode was added in commit 3fd58aa98dbae47ecb4a171f71278a78cf1e488e in 0.3.0.3-alpha:
```suggestion
      Fixes bug 24953; bugfix on 0.3.0.3-alpha.
```

scripts/maint/updateFallbackDirs.py
```diff
@@ -1609,14 +1613,14 @@ def apply_filter_lists(self, whitelist_obj, exact=False):
...
-                       f._fpr)
+          log_excluded('Excluding %s: not in %s.',
+                       f._fpr, list_type)
     self.fallbacks = filtered_fallbacks
     return excluded_count
 
   @staticmethod
   def summarise_filters(initial_count, excluded_count):
-    return '/* Whitelist excluded %d of %d candidates. */'%(
+    return '/* Fallback list excluded %d of %d candidates. */'%(
```
teor2345: Please use check_existing here.

scripts/maint/updateFallbackDirs.py
```diff
@@ -2244,7 +2253,11 @@ def list_fallbacks(whitelist, exact=False):
   """ Fetches required onionoo documents and evaluates the
       fallback directory criteria for each of the relays,
       passing exact to apply_filter_lists(). """
-  print "/* type=fallback */"
+  if whitelist['check_existing']:
+      print "/* type=fallback */"
+  else:
+      print "/* type=whitelist */"
```
teor2345: This is the type of the output list, not the input list. The output list always contains fallbacks, regardless of the type of the input list.

(More precisely: there's a mode that generates potential fallback candidates. But I don't think we've used it for a while. If we ever start using it again, we could update the type for that mode. If we don't use it for another year or two, we'll take it out instead.)

Also, this change will break existing parsers:
https://gitweb.torproject.org/torspec.git/tree/dir-list-spec.txt#n140

changes/bug24953
rl1987: Fixed in bf749490432c4e82fc415434fa9f380225303b71.

changes/bug24953
rl1987: Fixed in bf749490432c4e82fc415434fa9f380225303b71.

scripts/maint/updateFallbackDirs.py
rl1987: Undid this change in 3d5af708569fe83f1829395dd28b6b67cf435243.

scripts/maint/updateFallbackDirs.py
rl1987: Did so in 5952762901df4eee120f6155999ece4796909050.

PR #635, opened 2019-01-07
Bug28634

asn-d6:


src/core/or/circuitpadding.c
```diff
@@ -539,6 +539,11 @@ circpad_machine_sample_delay(circpad_machine_runtime_t *mi)
     histogram_total_tokens = state->histogram_total_tokens;
   }
 
+  /* If we are out of tokens, don't schedule padding. */
```
mikeperry-tor: What triggered this condition? It seems OK as a safety check, but shouldn't our machines transition to the END state instead of relying on this?

src/core/or/circuitpadding_machines.c
```diff
@@ -103,88 +92,36 @@ setup_burst_state_for_hiding_intro_circuits(circpad_state_t *burst_state,
...
-   * We want a machine that will send up to 20 fake cells over a reasonable
-   * time frame, to obfuscate the fact that introduction circuits always send 4
-   * cells both ways */
-
-  /* Figure out the length of the BURST state so that it's randomized. See
-     definition of these constants for rationale. */
+  /* Figure out the length of the BURST state so that it's randomized. We will
+   * set the histogram such that we don't send any padding from the
+   * origin-side, so just tune these parameteres for the relay-side. */
   burst_state->length_dist.type = CIRCPAD_DIST_UNIFORM;
```
mikeperry-tor: I do not see the code that prevents us from using the histogram on the origin-side. It looks like the origin side uses the same state machine event setup, and thus will send the same padding as relay side?

src/core/or/circuitpadding_machines.c
```diff
@@ -194,29 +131,53 @@ circpad_machine_client_hide_intro_circuits(smartlist_t *machines_sl)
...
 
-  /* XXX Keep the circuit alive even after the introduction has been finished,
+  /* Keep the circuit alive even after the introduction has been finished,
    * otherwise the short-term lifetime of the circuit will blow our cover */
-  // client_machine->manage_circ_lifetime = 1;
+  client_machine->manage_circ_lifetime = 1;
 
-  /* Let the origin-side machine control the lifetime of the circuit. */
+  /* Let the client-side machine control the lifetime of the circuit.
+   * XXX This might act as a fingerprint at the end of the conn. */
```
mikeperry-tor: Good point with the XXX. Now that we have manage_circ_lifetime, I don't think we should use should_negotiate_end.. The circuit will stay open until the padding machine shuts down, and then it will time out and get closed by circuit_expire_old_circuits_clientside(). So I would remove this block.

src/core/or/circuitpadding.c
asn-d6: Hmm, this triggers for the origin-side rendezvous machine right now, because of this transition:
```
  machine->states[CIRCPAD_STATE_BURST].
    next_state[CIRCPAD_EVENT_PADDING_SENT] = CIRCPAD_STATE_BURST;
```
and the fact that the code `circpad_send_padding_cell_for_callback()` first does the `circpad_cell_event_padding_sent(circ);`  transition and attempts to send padding with zero tokens, and then calls `check_machine_token_supply()` to potentially bring the machine to `END`.

The BURST->BURST transition above is not necessary because the machine only has one token, but in previous iterations of the code it had more than one token and that's why I had that transition.

What do you think?

src/core/or/circuitpadding_machines.c
asn-d6: It's because I do this for the origin-side:
```
    burst_state->histogram_edges[0] = CIRCPAD_DELAY_INFINITE-1;
    burst_state->histogram_edges[1] = CIRCPAD_DELAY_INFINITE;
```
which is a hacky way to never send padding.

It's hacky because it's not actually an infinity bin, more like a infinity-1 bin. That was necessary since we recently enforced strictly increasing bin orders and I could not have two infinity edges in a row.

src/core/or/circuitpadding_machines.c
asn-d6: Hm, sounds good, but then what would close the relay-side of the circuit? I guess just because the origin-side circuit will close (by expiring), the relay-side of the circuit will also close due to the other end missing? Let me know if this is right and I will remove the block.

mikeperry-tor: Yes when the origin circuit is closed, it will cause a DESTROY to get sent towards that middle, which will then close and free its circuit.

PR #641, opened 2019-01-09
rend: stop warning when clients send multiple rend establish cells

teor2345:
Instead, make "Tried to establish rendezvous on non-OR circuit" into a protocol warning.

Bug fix on TODO.


teor2345:
Whoops, this branch is in torproject/tor, but it should be in teor2345/tor

PR #649, opened 2019-01-16
Log more messages when failing to decode RSA keys

nmathewson:
We log these messages at INFO level, except when we are reading a
private key from a file, in which case we log at WARN.

This fixes a regression from when we re-wrote our PEM code to be
generic between nss and openssl.

Fixes bug 29042, bugfix on 0.3.5.1-alpha.


nmathewson:
Rebased and merged!

nmathewson:
Actually I don't think I actually merged this.

PR #654, opened 2019-01-17
Ticket21377 035 01 autosquashed rebased maint035

juga0:
Follows https://github.com/torproject/tor/pull/468

### DIFF
--- a/src/feature/dircache/dircache.c
+++ b/src/feature/dircache/dircache.c
@@ -1470,7 +1470,6 @@ static int
 handle_get_next_bandwidth(dir_connection_t *conn,
                           const get_handler_args_t *args)
 {
-  (void)args;
   log_debug(LD_DIR, "Getting next bandwidth.");
   const or_options_t *options = get_options();
   const compress_method_t compress_method =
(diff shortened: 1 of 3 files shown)


src/feature/dircache/dircache.c
```diff
@@ -1461,6 +1465,42 @@ handle_get_networkstatus_bridges(dir_connection_t *conn,
...
+                                       RFTS_IGNORE_MISSING, NULL);
+    if (bandwidth != NULL) {
+      ssize_t len = strlen(bandwidth);
+      write_http_response_header(conn, compress_method != NO_METHOD ? -1 : len,
+                                 compress_method, BANDWIDTH_CACHE_LIFETIME);
+      if (compress_method != NO_METHOD) {
+        conn->compress_state = tor_compress_new(1, compress_method,
+                                        choose_compression_level(len/2));
+        log_debug(LD_DIR, "Compressing bandwidth file.");
+        connection_buf_add_compress(bandwidth, len, conn, 0);
```
dgoulet-tor: Just minutes ago, a patch went in tor that adds `connection_dir_buf_add()` which supports compression if a state exists. So we could simplify this whole if() by simply creating a compression state if need be and then calling that function.

PR #661, opened 2019-01-21
Fix intermittent failures of test_circuitpadding_wronghop().

asn-d6:
We fix it by disabling the scheduling of actual padding.
Fixes #29122.


nmathewson:
Squashed and merged!

PR #666, opened 2019-01-23
Add circuit time check before logging about relaxing circuit time

JMGuisadoG:
Fixes [[Tor trac 28698](https://trac.torproject.org/projects/tor/ticket/28698)].


nmathewson:
Cherry-picked as #710 and merged to 0.4.0 and forward.

PR #671, opened 2019-01-30
Bug28614 040

nmathewson:


changes/ticket28614
```diff
@@ -0,0 +1,8 @@
+  o Major bugfixes (windows, startup):
+    - When writing a consensus file to disk, always write in
+      "binary" mode so that we can safely map it into memory later.
+      Fixes part of bug 28614; bugfix on 0.4.0.1-alpha.
+    - When reading a consensus file from disk, detect whether it
+      was written in text mode, and re-read it in text mode if it
```
teor2345:
```suggestion
      was written in text mode, and re-read it in text mode.
```

src/feature/nodelist/networkstatus.c
```diff
@@ -1750,6 +1742,41 @@ networkstatus_set_current_consensus_from_ns(networkstatus_t *c,
...
+                           const char *source_dir)
+{
+  tor_mmap_t *map = tor_mmap_file(fname);
+  if (!map)
+    return 0;
+
+  int rv = networkstatus_set_current_consensus(map->data, map->size,
+                                               flavor, flags, source_dir);
+#ifdef _WIN32
+  if (rv < 0 && tor_memstr(map->data, map->size, "\r\n")) {
```
teor2345: We need to do this check *before* we parse the consensus, otherwise users will see a warning from networkstatus_set_current_consensus().

teor2345: In particular, I think we should either call:
```
networkstatus_set_current_consensus(map->data, ...
```
or
```
char *content = read_file_to_str(fname, RFTS_IGNORE_MISSING, NULL);
rv = networkstatus_set_current_consensus(content, ...
```
But not both.

PR #673, opened 2019-01-30
Bug23576: Put intro point IPv6 addresses in v3 onion service descriptors

teor-travel:


src/feature/hs/hs_common.c
```diff
@@ -1843,3 +1843,40 @@ hs_inc_rdv_stream_counter(origin_circuit_t *circ)
...
+link_specifier_dup(const link_specifier_t *src)
+{
+  link_specifier_t *dup = NULL;
+  uint8_t *buf = NULL;
+
+  if (BUG(!src)) {
+    goto err;
+  }
+
+  ssize_t est_encoded_len = link_specifier_encoded_len(src);
```
dgoulet-tor: Is it possible you meant `dst_encoded_len`? I have no idea what `est_` would mean here hehe?

src/feature/hs/hs_descriptor.c
```diff
@@ -1190,52 +1190,16 @@ decode_link_specifiers(const char *encoded)
...
-             link_specifier_getconstarray_un_ed25519_id(ls),
-             sizeof(hs_spec->u.ed25519_id));
-      break;
-    default:
-      tor_free(hs_spec);
-      goto err;
-    }
-
-    smartlist_add(results, hs_spec);
+    smartlist_add(results, link_specifier_dup(ls));
```
dgoulet-tor: Because `link_specifier_dup()` can return NULL, I would be a bit safe here and check `ls` even though theoretically at this point the link specifier should be valid and thus decodable.

src/feature/hs/hs_common.c
teor2345: "estimated", but that's wrong, it should be the actual length. (Or at least an upper bound on the length, depending on how TRUNNEL_DYNARRAY_LEN() works.)

teor2345: Fixed in 4cbb752f73.

src/feature/hs/hs_descriptor.c
teor2345: I revised all the link_specifier_dup() NULL tests in a3e9805dea.

I also noticed that hs_link_specifier_dup() and link_specifier_dup() do the same thing, but link_specifier_dup() uses the opaque API. So I removed hs_link_specifier_dup() in bf915e67cd.

PR #685, opened 2019-02-07
Fast rng

nmathewson:


src/lib/crypt_ops/crypto_rand_numeric.c
```diff
@@ -16,21 +49,9 @@
...
-   * distribution with clipping at the upper end of unsigned int's
-   * range.
-   */
-  cutoff = UINT_MAX - (UINT_MAX%max);
-  while (1) {
-    crypto_rand((char*)&val, sizeof(val));
-    if (val < cutoff)
-      return val % max;
-  }
+  return (int)crypto_rand_uint(max);
```
asn-d6: Hmm, this does not look right. Why don't we use the macro here and instead we are casting from unsigned to signed?

src/lib/crypt_ops/crypto_rand_fast.c
```diff
@@ -0,0 +1,255 @@
...
+ **/
+
+/* This library is currently implemented to use the same implementation
+ * technique as libottery, using AES-CTR-256 as our underlying stream cipher.
+ * It's backtracking-resistant immediately, and prediction-resistant after
+ * a while.
+ *
+ * Here's how it works:
+ *
+ * We generate psuedorandom bytes using AES-CTR-256.  We generate BUFLEN bytes
```
asn-d6: nit: typo psuedorandom

src/lib/crypt_ops/crypto_rand_fast.c
```diff
@@ -0,0 +1,255 @@
...
+ * at a time.  When we do this, we keep the first SEED_LEN bytes as the key
+ * and the IV for our next invocation of AES_CTR, and yield the remaining
+ * BUFLEN - SEED_LEN bytes to the user as they invoke the PRNG.  As we yield
+ * bytes to the user, we clear them from the buffer.
+ *
+ * Every RESEED_AFTER times we refill the buffer, we mix in an additional
```
asn-d6: Hmm, I think this is not quite right. Every RESEED_AFTER we **reseed**. When the available bytes are finished we **refill**. Perhaps we should specify these two different operations better in this file comment

src/lib/crypt_ops/crypto_rand_fast.c
```diff
@@ -0,0 +1,255 @@
...
+
+/* Alias for CRYPTO_FAST_RNG_SEED_LEN to make our code shorter.
+ */
+#define SEED_LEN (CRYPTO_FAST_RNG_SEED_LEN)
+
+/* The amount of space that we mmap for a crypto_fast_rng_t.
+ */
+#define MAPLEN 4096
+
+/* The number of random bytes that we can yield to the user after each
```
asn-d6: Might be worth specifying where the 4 comes from here. It's the IV IIUC.

src/lib/crypt_ops/crypto_rand_fast.c
```diff
@@ -0,0 +1,255 @@
...
+CTASSERT(KEY_BITS == 128 || KEY_BITS == 192 || KEY_BITS == 256);
+
+struct crypto_fast_rng_t {
+  /** How many more fills does this buffer have before we should mix
+   * in the output of crypto_rand()? */
+  uint16_t n_till_reseed;
+  /** How many bytes are remaining in cbuf.bytes? */
+  uint16_t bytes_left;
+  struct cbuf {
+    /** The seed that we will use the next time that we refill cbuf. */
```
asn-d6: This also contains the IV right? Perhaps we should mention this. Another approach would be to explicitly add another element in this struct holding the IV, but perhaps that would be more work than worth it.

src/lib/crypt_ops/crypto_rand_fast.c
```diff
@@ -0,0 +1,255 @@
...
+
+/**
+ * Helper: create a crypto_cipher_t object from SEED_LEN bytes of
+ * input.  The first KEY_LEN bytes are used as the stream cipher's key,
+ * and the remaining CIPHER_IV_LEN bytes are used as its IV.
+ **/
+static inline crypto_cipher_t *
+cipher_from_seed(const uint8_t *seed)
+{
+  return crypto_cipher_new_with_iv_and_bits(seed, seed+KEY_LEN, KEY_LEN*8);
```
asn-d6: nit: we could use KEY_BITS here since we have it anyway

src/lib/crypt_ops/crypto_rand_fast.c
```diff
@@ -0,0 +1,255 @@
...
+}
+
+/**
+ * Helper: extract bytes from the PRNG, refilling it as necessary. Does not
+ * optimize the case when the user has asked for a huge output.
+ **/
+static void
+crypto_fast_rng_getbytes_impl(crypto_fast_rng_t *rng, uint8_t *out, size_t n)
+{
+  while (n) {
```
asn-d6: nit: Perhaps we could rename the cryptic 'n' to something more useful like 'bytes_needed` or `bytes_still_needed` or something.

src/lib/crypt_ops/crypto_rand_fast.c
```diff
@@ -0,0 +1,255 @@
...
+    if (rng->bytes_left == 0)
+      crypto_fast_rng_refill(rng);
+
+    const size_t to_copy = MIN(rng->bytes_left, n);
+    uint8_t *copy_from = rng->buf.bytes +
```
asn-d6: I don't think there is an issue here but perhaps we can add some asserts before this, like `tor_assert(sizeof(rng->buf.bytes) >= rng->bytes_left);` to make sure the subtraction does not underflow?

src/lib/crypt_ops/crypto_rand_numeric.c
nmathewson: The macro only works on unsigned outputs; this function needs to return a signed output.  I'm adding a fixup as 2bf39d7209c5b7 to explain this better.

src/lib/crypt_ops/crypto_rand_fast.c
nmathewson: 54966b1509eafb is a fixup with the fix.

src/lib/crypt_ops/crypto_rand_fast.c
nmathewson: 44d7458ceac292 has a fixup on this one

src/lib/crypt_ops/crypto_rand_fast.c
nmathewson: 41a18ec696c040 is a fixup here -- I've changed 4 to sizeof(uint16_t)*2

src/lib/crypt_ops/crypto_rand_fast.c
nmathewson: a765b12584c7fe tries to explain this in a few more places.

src/lib/crypt_ops/crypto_rand_fast.c
nmathewson: 9d2b10fefc4d24 has the fix.

src/lib/crypt_ops/crypto_rand_fast.c
nmathewson: Renamed with bf26aa5de36e14

src/lib/crypt_ops/crypto_rand_fast.c
nmathewson: Sure.  It can't underflow as the code is today, but it couldn't hurt to make sure we don't mess it up in the future.  Fixup also in bf26aa5de36e14

src/lib/crypt_ops/crypto_rand_numeric.c
asn-d6: Ah makes sense. LGTM.

src/lib/crypt_ops/crypto_rand_fast.c
asn-d6: LGTM

src/lib/crypt_ops/crypto_rand_fast.c
asn-d6: LGTM

src/lib/crypt_ops/crypto_rand_fast.c
asn-d6: LGTM

src/lib/crypt_ops/crypto_rand_fast.c
asn-d6: LGTM

src/lib/crypt_ops/crypto_rand_fast.c
asn-d6: LGTM

src/lib/crypt_ops/crypto_rand_fast.c
asn-d6: LGTM

src/lib/crypt_ops/crypto_rand_fast.c
asn-d6: LGTM

nmathewson:
making a new squashed PR

PR #688, opened 2019-02-07
#29065: Fix shellcheck SC2006 warnings in test_switch_id.sh

rl1987:
https://trac.torproject.org/projects/tor/ticket/29065


nmathewson:
Squashed and merged.

PR #708, opened 2019-02-17
[Tor Trac #28525]: Make tor_addr_is_internal_() aware of RFC 6598 (Carrier Grade NAT/Large Scale NAT) IPv4 Ranges

neelchauhan:
For the ticket [here](https://trac.torproject.org/projects/tor/ticket/28525).

Sorry about this, but I did a new PR because reusing the old one (#529) would have added many unnecessary commits.


teor2345:
Replaced by #762 and #763.

PR #717, opened 2019-02-20
#29532: Update git pre-push hook so that only upstream branches can get pushed to origin

rl1987:
https://trac.torproject.org/projects/tor/ticket/29532

### DIFF
--- a/scripts/maint/pre-push.git-hook
+++ b/scripts/maint/pre-push.git-hook
@@ -11,29 +11,20 @@
 # The following sample script was used as starting point:
 # https://github.com/git/git/blob/master/templates/hooks--pre-push.sample
 
+echo "Running pre-push hook"
+
 z40=0000000000000000000000000000000000000000
 
 remote="$1"
-CUR_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 
-# Only allow pushing master, release-* and maint-* branches to origin.
-if [ "$remote" == "origin" ]
-then
-        if  [ "$CUR_BRANCH" != "master" ] && [[ $CUR_BRANCH != release-* ]] &&
-             [[ $CUR_BRANCH != maint-* ]]
+ref_is_upstream_branch() {
+        if [ "$1" == "refs/heads/master" ] ||
+                [[ "$1" == refs/heads/release-* ]] ||
+                [[ "$1" == refs/heads/maint-* ]]
         then
-                echo >&2 "Not pushing $CUR_BRANCH to origin"
-                exit 1
+                return 1
         fi
-fi
-
-if [ "$CUR_BRANCH" != "master" ] && [[ $CUR_BRANCH != release-* ]] &&
-        [[ $CUR_BRANCH != maint-* ]]
-then
-        exit 0
-fi
-
-echo "Running pre-push hook"
+}
 
 # shellcheck disable=SC2034
 while read -r local_ref local_sha remote_ref remote_sha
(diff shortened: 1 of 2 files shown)


scripts/maint/pre-push.git-hook
```diff
@@ -1,17 +1,32 @@
...
 
 z40=0000000000000000000000000000000000000000
 
+remote="$1"
 CUR_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
+# Only allow pushing master, release-* and maint-* branches to origin.
+if [ "$remote" == "origin" ]
+then
+        if  [ "$CUR_BRANCH" != "master" ] || [[ $CUR_BRANCH != release-* ]] ||
```
nmathewson: Should these `||` operators be `&&` ?

scripts/maint/pre-push.git-hook
```diff
@@ -1,17 +1,32 @@
...
-# This is git pre-push hook script to prevent "fixup!" and "squash!" commits
-# from ending up in upstream branches (master, release-* or maint-*).
-#
 # The following sample script was used as starting point:
 # https://github.com/git/git/blob/master/templates/hooks--pre-push.sample
 
 z40=0000000000000000000000000000000000000000
 
+remote="$1"
 CUR_BRANCH=$(git rev-parse --abbrev-ref HEAD)
```
nmathewson: I know this line isn't new, but does it still work if you're pushing something other than HEAD, as in "git push branchname reponame" ?

scripts/maint/pre-push.git-hook
rl1987: Yes, fixed in fb6e9ba26dc411ac79d0b6d3dabf71868ba83de5.

scripts/maint/pre-push.git-hook
rl1987: Good point. I tried to make it more general in 60c8af367799e2cb8f59a236e97fdb0ad04bc350.

scripts/maint/pre-push.git-hook
```diff
@@ -52,6 +43,19 @@ do
 			range="$remote_sha..$local_sha"
 		fi
 
+                if ref_is_upstream_branch "$local_ref" == 0 ||
```
nmathewson: Should we be checking local_ref at all, or only remote_ref ?

rl1987: I suppose we should check both, to cover weird situations of local regular branch being pushed to remote upstream branch, and vice versa?

nmathewson: In that case, I think we should explicitly check for match; pushing maint-0.3.5 to maint-0.3.4 is as bad as pushing bug12345 to maint-0.3.4.

rl1987: a94c3328829482e0da7a175e3dee089b9113d553 and 59c26154e80d589a069e54a9aa52eca2e14bf322 fixes that.

scripts/maint/pre-push.git-hook
```diff
@@ -44,8 +44,9 @@ do
 			range="$remote_sha..$local_sha"
 		fi
 
-                if ref_is_upstream_branch "$local_ref" == 0 ||
-                        ref_is_upstream_branch "$remote_ref"  == 0
+                if (ref_is_upstream_branch "$local_ref" == 0 ||
+                        ref_is_upstream_branch "$remote_ref"  == 0) &&
+                        "$local_ref" ne "$remote_ref"
```
nmathewson: Does this part need to be in [square brackets] ?

rl1987: Shellcheck does not complain about this, so I think not? Then again I'm really not sure.

rl1987: On second thought, let's use `[...]` for consistency with previous code. See e28bbc8bd5847aba774b5ea38e7e98fbe5dd11fd.

PR #719, opened 2019-02-21
Bug 29204: Inspect circuit queues before sending padding.

mikeperry-tor:
Mitigates OOM conditions at relays.

### DIFF
--- a/src/core/or/circuitpadding.c
+++ b/src/core/or/circuitpadding.c
@@ -982,7 +982,6 @@ circpad_send_padding_cell_for_callback(circpad_machine_state_t *mi)
   circuit_t *circ = mi->on_circ;
   int machine_idx = mi->machine_index;
   mi->padding_scheduled_at_usec = 0;
-  mi->is_padding_timer_scheduled = 0;
   circpad_statenum_t state = mi->current_state;
 
   // Make sure circuit didn't close on us
--- a/src/core/or/circuitpadding.c
+++ b/src/core/or/circuitpadding.c
@@ -61,6 +61,7 @@
 #include "core/or/crypt_path_st.h"
 #include "core/or/circuit_st.h"
 #include "core/or/origin_circuit_st.h"
+#include "core/or/or_circuit_st.h"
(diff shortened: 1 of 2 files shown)


src/core/or/circuitpadding.c
```diff
@@ -1027,10 +1030,17 @@ circpad_send_padding_cell_for_callback(circpad_machine_state_t *mi)
   } else {
     // If we're a non-origin circ, we can just send from here as if we're the
     // edge.
-    log_fn(LOG_INFO,LD_CIRC,
-          "Callback: Sending padding to non-origin circuit.");
-    relay_send_command_from_edge(0, mi->on_circ, RELAY_COMMAND_DROP, NULL,
-                                 0, NULL);
+    if (TO_OR_CIRCUIT(circ)->p_chan_cells.n <= circpad_max_circ_queued_cells) {
```
dgoulet-tor: I don't think this will work as expected. I think the problem is more on the outbuf side. If none of the cells go on the wire but we keep queuing padding cells directly in the outbuf, we'll ultimately fill it.

The above will keep sending those cells as long as the circuit queue has not too many cells but that won't work because the bloating in this case happens at the outbuf level.

Furthermore, the circuit cell queues already have a max check (in `append_cell_to_circuit_queue()`):

```
  if (PREDICT_UNLIKELY(queue->n >= max_circuit_cell_queue_size)) {
```

I think to fix this, we need a way to ask the channel "can you write on the wire?" and if no, we do not queue anything and if yes, we check if the circuit queue *and* outbuf have stuff. If yes, I'm guessing do not send padding... (Not entirely sure about this because in between traffic pattern you might want to send padding especially for quiet circuit even though everything is still in the outbuf?)

Then you might want to couple the above with a "last cell emitted timestamp" to have some max threshold when you definitely stop?...

Also, the TROVE we just got in tor today makes it that KIST now considers the outbuf size in order to make a decision if there is enough space to flush cells from the circuit queue.

If the circuit padding subsystem does send the padding cell through the circuit queue, then we might be fine here. If it writes it directly into the outbuf, then problem.

PR #720, opened 2019-02-21
Update pre-commit.git-hook for ticket 29553

nmathewson:
- handle older source layout
 - handle empty changes directories
 - "set -e" so that  we exit if there's a problem.

### DIFF
--- a/scripts/maint/pre-commit.git-hook
+++ b/scripts/maint/pre-commit.git-hook
@@ -10,16 +10,31 @@ workdir=$(git rev-parse --show-toplevel)
 
 cd "$workdir" || exit 1
 
-python scripts/maint/lintChanges.py ./changes/*
+set -e
 
-perl scripts/maint/checkSpace.pl -C \
-src/lib/*/*.[ch] \
-src/core/*/*.[ch] \
-src/feature/*/*.[ch] \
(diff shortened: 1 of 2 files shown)


scripts/maint/pre-commit.git-hook
```diff
@@ -10,16 +10,31 @@ workdir=$(git rev-parse --show-toplevel)
...
-src/app/*/*.[ch] \
-src/test/*.[ch] \
-src/test/*/*.[ch] \
-src/tools/*.[ch]
+if [ ! -z "ls ./changes/*" ]; then
+    python scripts/maint/lintChanges.py ./changes/*
+fi
+
+if [ -d src/lib ]; then
+    # This is the layout in 0.3.5
```
rl1987: Maybe we can use find(1) here to automagically get a list of files we want to check?

nmathewson: That could be a good idea!  We'd want to exclude src/ext, and include everything else.  Maybe:
```
 ./scripts/maint/checkSpace.pl -C $( find src/ -name "*.[ch]" -and -not path "src/trunnel/*" -and -not -path "src/ext/*" )
```
?

PR #724, opened 2019-02-25
#29537: Check for possible failures of roundtrip between void * and (unsigned) int

rl1987:
https://trac.torproject.org/projects/tor/ticket/29537

### DIFF
--- a/changes/ticket29537
+++ b/changes/ticket29537
@@ -1,0 +1,3 @@
+  o Testing:
+    - Check that all valid values of `int` and `unsigned int` can be
+      represented by `void *`. Resolves issue 29537.
--- a/changes/ticket29537
+++ b/changes/ticket29537
@@ -1,3 +1,3 @@
   o Testing:
-    - Check that all valid values of `int` and `unsigned int` can be
-      represented by `void *`. Resolves issue 29537.
+    - Check that representative subsets of values of `int` and `unsigned int`
+      can be represented by `void *`. Resolves issue 29537.
--- a/src/test/test_ptr_slow.c
+++ b/src/test/test_ptr_slow.c
@@ -33,7 +33,11 @@ test_int_voidstar_interop(void *arg)
   int a;
   (void)arg;
 
-  for (a = 0; a <= 1024; a++) {
+  for (a = -1024; a <= 1024; a++) {
+    assert_int_voidptr_roundtrip(a);
+  }
+
+  for (a = INT_MIN; a <= INT_MIN+1024; a++) {
     assert_int_voidptr_roundtrip(a);
   }
 
--- a/src/test/test_ptr_slow.c
+++ b/src/test/test_ptr_slow.c
@@ -10,46 +10,65 @@
 #include <stdint.h>
 #include <limits.h>
 
-/** Check that all values of int can be cast to void * and back. */
+/** Assert that <b>a</b> can be cast to void * and back. */
+static void
+assert_int_voidptr_roundtrip(int a)
+{
+  intptr_t ap = (intptr_t)a;
+  void *b = (void *)ap;
+  intptr_t c = (intptr_t)b;
+  void *d = (void *)c;
+
+  tt_assert(ap == c);
+  tt_assert(b == d);
+
+ done:
+  return;
+}
+
 static void
 test_int_voidstar_interop(void *arg)
 {
   int a;
   (void)arg;
 
-  for (a = INT_MIN; a < INT_MAX; a++) {
-   intptr_t ap = (intptr_t)a;
-   void *b = (void *)ap;
-   intptr_t c = (intptr_t)b;
-   void *d = (void *)c;
-
-   tt_assert(ap == c);
-   tt_assert(b == d);
+  for (a = 0; a <= 1024; a++) {
+    assert_int_voidptr_roundtrip(a);
   }
+
+  for (a = INT_MAX-1024; a < INT_MAX; a++) {
+    assert_int_voidptr_roundtrip(a);
+  }
+}
+
+static void
+assert_uint_voidptr_roundtrip(unsigned int a)
+{
+ intptr_t ap = (intptr_t)a;
+ void *b = (void *)ap;
+ intptr_t c = (intptr_t)b;
+ void *d = (void *)c;
+
+ tt_assert(ap == c);
+ tt_assert(b == d);
 
  done:
   return;
 }
 
-/** Check that all values of unsigned int can be cast to void * and back. */
 static void
 test_uint_voidstar_interop(void *arg)
 {
   unsigned int a;
   (void)arg;
 
-  for (a = 0; a < UINT_MAX; a++) {
-   intptr_t ap = (intptr_t)a;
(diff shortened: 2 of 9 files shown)


src/test/test_ptr_slow.c
```diff
@@ -0,0 +1,86 @@
...
+  (void)arg;
+
+  for (a = 0; a <= 1024; a++) {
+    assert_int_voidptr_roundtrip(a);
+  }
+
+  for (a = INT_MAX-1024; a < INT_MAX; a++) {
+    assert_int_voidptr_roundtrip(a);
+  }
+}
```
teor2345: This test gives good coverage of the low bits, but the high bits are identical within each loop. It might also be worth testing each bit individually.

src/test/test_ptr_slow.c
```diff
@@ -0,0 +1,86 @@
...
+static void
+assert_int_voidptr_roundtrip(int a)
+{
+  intptr_t ap = (intptr_t)a;
+  void *b = (void *)ap;
+  intptr_t c = (intptr_t)b;
+  void *d = (void *)c;
+
+  tt_assert(ap == c);
+  tt_assert(b == d);
```
teor2345: A smart optimiser may compile this function down to nothing.
(Has anyone checked what the current clang and gcc do?)

We can avoid most optimisations by placing each kind of cast in a function in another C file.
(Telling compilers not to optimise may not work: some compilers do some optimisations as a necessary part of translation.)

src/test/test_ptr_slow.c
```diff
@@ -0,0 +1,86 @@
...
+ done:
+  return;
+}
+
+static void
+test_int_voidstar_interop(void *arg)
+{
+  int a;
+  (void)arg;
+
```
teor2345: It's important that we check negative integers.

Let's check -1024…0 and INT_MIN…(INT_MIN+1024)?

rl1987: Did so in 280109473fe30d1e67f8db09ce279cc1f714a682.

src/test/test_ptr_slow.c
rl1987: Moved the cast operations to separate file in 0bc9ed9d38cb29783d000b6e22677ae16727976b.

src/test/test_ptr_slow.c
rl1987: 72e0dc0822deb8587b3cae2edf584ece13cb6bb8

src/test/test_ptr_slow.c
```diff
@@ -0,0 +1,103 @@
...
+#include "core/or/or.h"
+#include "test/test.h"
+#include "test/ptr_helpers.h"
+
+#include <stdint.h>
+#include <limits.h>
+
+/** Assert that <b>a</b> can be cast to void * and back. */
+static void
+assert_int_voidptr_roundtrip(int a)
```
teor2345: Is sizeof(int) always smaller than or equal to sizeof(void *) on all our supported platforms?
I think they are. But I'm not sure if that's guaranteed by the C spec.

Do we need to document this assumption, maybe using the SIZEOF macros?
Or does Tor already do a similar size check elsewhere?

src/test/test_ptr_slow.c
```diff
@@ -0,0 +1,103 @@
...
+
+  a = 1;
+  for (unsigned long i = 0; i < sizeof(int) * 8; i++) {
+    assert_int_voidptr_roundtrip(a);
+    a = (a << 1);
+  }
+}
+
+static void
+assert_uint_voidptr_roundtrip(unsigned int a)
```
teor2345: I have a similar size question for unsigned int.

src/test/test_ptr_slow.c
```diff
@@ -0,0 +1,103 @@
...
+  void *d = cast_intptr_to_voidstar(c);
+
+  tt_assert(ap == c);
+  tt_assert(b == d);
+
+ done:
+  return;
+}
+
+static void
```
teor2345: Add a function comment please.

src/test/test_ptr_slow.c
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  a = 1;
+  for (unsigned long i = 0; i < sizeof(int) * 8; i++) {
+    assert_int_voidptr_roundtrip(a);
+    a = (a << 1);
+  }
+}
+
+static void
```
teor2345: Add a function comment please.

src/test/test_ptr_slow.c
```diff
@@ -0,0 +1,103 @@
...
+ void *d = cast_uintptr_to_voidstar(c);
+
+ tt_assert(ap == c);
+ tt_assert(b == d);
+
+ done:
+  return;
+}
+
+static void
```
teor2345: Add a function comment please.

src/test/test_ptr_slow.c
rl1987: We can leverage the build system for this. See: 13b28063f9f2a3d2a587eabac358452a42277052

src/test/test_ptr_slow.c
rl1987: 7b30f8dc8c93ac215a1e7330fee1baa5f1437a98

src/test/test_ptr_slow.c
rl1987: 7b30f8dc8c93ac215a1e7330fee1baa5f1437a98

src/test/test_ptr_slow.c
rl1987: Addressed in 13b28063f9f2a3d2a587eabac358452a42277052.

src/test/test_ptr_slow.c
rl1987: Fixed in 7b30f8dc8c93ac215a1e7330fee1baa5f1437a98.

changes/ticket29537
```diff
@@ -0,0 +1,3 @@
+  o Testing:
+    - Check that all valid values of `int` and `unsigned int` can be
+      represented by `void *`. Resolves issue 29537.
```
tlyu: I think this is now "a representative subset of values of int and unsigned int", right?

rl1987: Fixed in 9529d99982eb64998e12db7704fa977bf02f541c.

src/test/test_ptr_slow.c
```diff
@@ -0,0 +1,115 @@
...
+#include <stdint.h>
+#include <limits.h>
+
+#if SIZEOF_INT > SIZEOF_VOID_P
+#error "sizeof(int) > sizeof(void *) - Tor cannot be built on this platform!"
+#endif
+
+#if SIZEOF_UNSIGNED_INT > SIZEOF_VOID_P
+#error "sizeof(unsigned int) > sizeof(void *) - Tor cannot be built on this \
+platform!"
```
teor2345: GitHub's syntax highlighting hates this split string literal, but it is correct according to the C standard.
Let's see if it breaks any compilers?

PR #725, opened 2019-02-25
Ticket26288 041 02

dgoulet-tor:


src/core/or/connection_edge.c
```diff
@@ -4540,6 +4540,25 @@ circuit_clear_isolation(origin_circuit_t *circ)
...
+ *
+ * This function can't fail and does nothing if conn is NULL. */
+void
+connection_edge_end_close(edge_connection_t *conn, uint8_t reason)
+{
+  if (!conn) {
+    return;
+  }
+
+  connection_edge_end(conn, reason);
```
nmathewson: I could have sworn that we already had a function that did this, but it looks like we don't.

src/core/or/sendme.c
```diff
@@ -213,3 +213,40 @@ sendme_stream_data_received(edge_connection_t *conn)
   tor_assert(conn);
   return --conn->deliver_window;
 }
+
+/* Called when a relay DATA cell is packaged on the given circuit. If
+ * layer_hint is NULL, this means we are the Exit end point else we are the
+ * Client. Update the package window and return its new value. */
+int
+sendme_circuit_data_packaged(circuit_t *circ, crypt_path_t *layer_hint)
```
nmathewson: I'd suggest adding "note" between sendme and stream/circuit in these functions' names, to make their purpose more clear.

src/core/or/sendme.c
```diff
@@ -48,6 +50,98 @@ get_accept_min_version(void)
...
+  tor_assert(payload);
+
+  cell = sendme_cell_new();
+  data = sendme_data_v1_new();
+
+  /* Building a payload for version 1. */
+  sendme_cell_set_version(cell, 0x01);
+
+  /* Copy the digest into the data payload. */
+  crypto_digest_get_digest(cell_digest,
```
nmathewson: Suggestion: instead of using a crypto_digest_t in all of these functions, it would be a good idea to use a relay_crypto_t, and put the function for generating the sendme tag into src/core/crypto/relay_crypto.c.  That way, we separate the code that knows how the tag is constructed from the code that uses the tag.  This might help us later on, when we want to change how circuit crypto works.

src/core/or/sendme.c
```diff
@@ -50,7 +53,124 @@ get_accept_min_version(void)
...
+ * accepted version from the consensus is known to us. */
+static bool
+cell_version_is_valid(uint8_t cell_version)
+{
+  int accept_version = get_accept_min_version();
+
+  /* Can we handle this version? */
+  if (accept_version > SENDME_MAX_SUPPORTED_VERSION) {
+    log_warn(LD_PROTOCOL, "Unable to handle SENDME version %u "
+                          "(from the consensus). We only support <= %d. "
```
nmathewson: Why does this say "from the consensus"?  it's coming from a cell, not the consensus.  Maybe we shouldn't warn here, but send protocol_warn instead?

src/core/or/circuit_st.h
```diff
@@ -110,6 +110,12 @@ struct circuit_t {
    * circuit-level sendme cells to indicate that we're willing to accept
    * more. */
   int deliver_window;
+  /** FIFO containing the digest of the cells that are just before a SENDME is
+   * sent by the client. It is done at the last cell before our package_window
+   * goes down to 0 which is when we expect a SENDME. The protocol doesn't
+   * allow more than 10 outstanding SENDMEs worth of data meaning this list
+   * should only contain at most 10 digests of 4 bytes each. */
```
nmathewson: We should document how these digests correspond to positions in the stream, exactly.  Like, the most recent digest in the list corresponds to the state up until the cell we received at...

Also, "10" and "4" should really be constants here.

src/core/or/sendme.c
```diff
@@ -507,3 +508,31 @@ sendme_stream_data_packaged(edge_connection_t *conn)
...
+    return;
+  }
+  /* Is this the last cell before a SENDME? The idea is that if the
+   * package_window reaches a multiple of the increment, after this cell, we
+   * should expect a SENDME. */
+  if (((circ->package_window - 1) % CIRCWINDOW_INCREMENT) != 0) {
+    return;
+  }
+
+  digest = tor_malloc_zero(4);
```
nmathewson: This should be a constant,

src/core/or/sendme.c
```diff
@@ -72,9 +73,33 @@ cell_v1_is_valid(const sendme_cell_t *cell)
     goto invalid;
   }
 
-  /* XXX: Match the digest in the cell to the previous cell. Needs to be
-   * implemented that is passed to this function and compared. For this, we
-   * need #26839 that is making tor remember the last digest(s). */
+  /* We shouldn't have received this SENDME if we have no digests. Log at
+   * protocol warning because it can be tricked by sending many SENDMEs
+   * without prior data cell. */
+  if (circ->sendme_last_digests == NULL ||
```
nmathewson: Maybe let's extract this digest-matching code into its own function?

src/core/or/sendme.c
```diff
@@ -72,9 +73,33 @@ cell_v1_is_valid(const sendme_cell_t *cell)
...
+  }
+
+  /* Pop the first element that was added (FIFO) and compare it. */
+  {
+    uint8_t *digest = smartlist_get(circ->sendme_last_digests, 0);
+    smartlist_del_keeporder(circ->sendme_last_digests, 0);
+
+    /* Compare the digest with the one in the SENDME. This cell is invalid
+     * without a perfect match. */
+    if (tor_memcmp(digest, sendme_data_v1_getconstarray_digest(data),
```
nmathewson: If you're only checking for inequality, prefer tor_memneq() to tor_memcmp().  tor_memcmp() is more complicated because it has to determine ordering.

src/core/crypto/relay_crypto.c
```diff
@@ -142,6 +142,13 @@ relay_decrypt_cell(circuit_t *circ, cell_t *cell,
           if (relay_digest_matches(thishop->crypto.b_digest, cell)) {
             *recognized = 1;
             *layer_hint = thishop;
+            /* Keep current digest of this cell for the possible SENDME. */
+            if (thishop->crypto.sendme_digest) {
+              crypto_digest_free(thishop->crypto.sendme_digest);
+            }
+            thishop->crypto.sendme_digest =
+              crypto_digest_dup(thishop->crypto.b_digest);
```
nmathewson: This approach seems needlessly wasteful to me.  The digest object is as large as the entire internal digest object's state, which is often much larger than the actual set of bytes you're transmitting.  Why not just store the digest itself, rather than the crypto_digest_t?

src/core/crypto/relay_crypto.c
```diff
@@ -142,6 +142,13 @@ relay_decrypt_cell(circuit_t *circ, cell_t *cell,
           if (relay_digest_matches(thishop->crypto.b_digest, cell)) {
             *recognized = 1;
             *layer_hint = thishop;
+            /* Keep current digest of this cell for the possible SENDME. */
+            if (thishop->crypto.sendme_digest) {
+              crypto_digest_free(thishop->crypto.sendme_digest);
```
nmathewson: Should we log a BUG warning if we drop this value without having sent it in a SENDME cell?

src/core/or/relay.c
```diff
@@ -577,6 +577,14 @@ relay_send_command_from_edge_,(streamid_t stream_id, circuit_t *circ,
   if (payload_len)
     memcpy(cell.payload+RELAY_HEADER_SIZE, payload, payload_len);
 
+  /* Add random bytes to the unused portion of the payload, to foil attacks
+   * where the other side can predict all of the bytes in the payload and thus
+   * compute authenticated sendme cells without seeing the traffic.  See
+   * proposal 289. */
+  crypto_fast_rng_getbytes(get_thread_fast_rng(),
```
nmathewson: I recall an old suggestion I made once: let's skip the first 4 bytes of the unused data.  Having some bytes that we know will be set to 0 if they are unused has saved us a lot of times in the past.

src/core/or/relay.c
```diff
@@ -1812,87 +1812,52 @@ connection_edge_process_relay_cell(cell_t *cell, circuit_t *circ,
...
-          circuit_resume_edge_reading(circ, layer_hint);
+      if (!rh.stream_id) {
+        /* Circuit level SENDME cell. */
+        ret = sendme_process_circuit_level(layer_hint, circ, rh.length);
+        if (ret < 0) {
+          return ret;
         }
+        /* Resume reading on any streams now that we've processed a valid
+         * SENDME cell that updated our package window. */
+        circuit_resume_edge_reading(circ, layer_hint);
```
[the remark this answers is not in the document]
dgoulet-tor: Yes. The processing of the circuit level cell above with `sendme_process_circuit_level()` returns an error if anything is wrong with the cell which closes the circuits. On success, we resume reading because our package window increased.

The ONLY difference is that we call `circuit_read_valid_data()` in the processing step if everything went well which is the same as we had before where on success we would resume reading and then increment the counters with the valid data function. Overall, it is the same behavior.

src/core/or/sendme.c
```diff
@@ -82,3 +84,99 @@ sendme_circuit_consider_sending(circuit_t *circ, crypt_path_t *layer_hint)
     }
   }
 }
+
+/* Process a circuit-level SENDME cell that we just received. The layer_hint,
```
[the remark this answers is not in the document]
dgoulet-tor: I've spent quite a bit of time to go over each single line of this SENDME code in order to do exactly what we currently do with this refactoring commit.

My certainty is 99% I would say because of the amount of tests I did and time I spent in that code.

src/core/or/connection_edge.c
dgoulet-tor: Same!!... Now we have it :).

src/core/crypto/relay_crypto.c
dgoulet-tor: Commit 3943d096cc060d7676e1549e37d9577951b3b68f now makes it that we overwrite... At this point, we do not know if we are dropping it or not :S.

src/trunnel/sendme.trunnel
```diff
@@ -0,0 +1,18 @@
...
+  u16 data_len;
+  u8 data[data_len];
+}
+
+/* SENDME version 0. No data. */
+
+/* SENDME version 1. Authenticated with digest. */
+struct sendme_data_v1 {
+  /* A 4 bytes digest. */
+  u8 digest[4];
```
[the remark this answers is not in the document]
dgoulet-tor: Done in ab4493bf2a86046f

src/trunnel/sendme.trunnel
```diff
@@ -0,0 +1,18 @@
+/* This file contains the SENDME cell definition. */
+
+struct sendme_cell {
+  /* Version field. */
+  u8 version IN [0x00, 0x01];
+
+  /* The data content depends on the version. */
+  u16 data_len;
+  u8 data[data_len];
```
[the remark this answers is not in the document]
dgoulet-tor: Done in ab4493bf2a86046f

src/core/or/sendme.c
```diff
@@ -14,40 +14,48 @@
...
+  while (conn->deliver_window <=
+         (STREAMWINDOW_START - STREAMWINDOW_INCREMENT)) {
+    log_debug(log_domain, "Outbuf %lu, queuing stream SENDME.",
+              TO_CONN(conn)->outbuf_flushlen);
     conn->deliver_window += STREAMWINDOW_INCREMENT;
     if (connection_edge_send_command(conn, RELAY_COMMAND_SENDME,
                                      NULL, 0) < 0) {
-      log_warn(LD_APP,"connection_edge_send_command failed. Skipping.");
-      return; /* the circuit's closed, don't continue */
+      log_warn(LD_APP, "connection_edge_send_command failed while sending "
```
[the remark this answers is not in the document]
dgoulet-tor: Fixup commit 836671bbb7

src/core/or/relay.c
```diff
@@ -1812,87 +1812,52 @@ connection_edge_process_relay_cell(cell_t *cell, circuit_t *circ,
...
-          circuit_resume_edge_reading(circ, layer_hint);
-
-          /* We count circuit-level sendme's as valid delivered data because
-           * they are rate limited.
-           */
-          if (CIRCUIT_IS_ORIGIN(circ)) {
-            circuit_read_valid_data(TO_ORIGIN_CIRCUIT(circ),
-                                    rh.length);
-          }
+    {
```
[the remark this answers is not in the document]
dgoulet-tor: Done in e15771097f96f24d

src/core/or/sendme.c
dgoulet-tor: Done in commit ad100e855024251a

src/core/or/sendme.c
dgoulet-tor: The use of `crypto_digest_t` has been removed with commit 3943d096cc060d76. We do not compute the digest since we don't keep the digest object anymore.

src/core/or/sendme.c
dgoulet-tor: Ah!!! Gooooood catch, it is reverse. The second value is the one from the consensus. And yes protocol warn would is better!

Fixup commit: ddad00ece03e3862

src/core/or/circuit_st.h
dgoulet-tor: Attempt in commit a1cdc72082

src/core/or/sendme.c
dgoulet-tor: Done in ab4493bf2a86046f

src/core/or/sendme.c
dgoulet-tor: Done in commit 246577c86601f68ff1276804442ac713cb72d2d3

src/core/or/sendme.c
dgoulet-tor: Done in commit 246577c86601f68ff1276804442ac713cb72d2d3

src/core/crypto/relay_crypto.c
dgoulet-tor: Done in commit 3943d096cc060d7676e1549e37d9577951b3b68f

src/core/or/relay.c
dgoulet-tor: Done in commit 0c23471a35dfdd4cd286a31ee8b9774008001b71

There is some more calculation to make sure we don't overflow here. Needs special attentions :).

src/trunnel/sendme.trunnel
```diff
@@ -1,18 +1,19 @@
...
-
-/* SENDME version 1. Authenticated with digest. */
-struct sendme_data_v1 {
-  /* A 4 bytes digest. */
-  u8 digest[4];
+  /* The data content depends on the version. */
+  union data[version] with length data_len {
+    0x00: ignore;
+    0x01: u8 v1_digest[TRUNNEL_SENDME_V1_DIGEST_LEN];
+  };
```
nmathewson: Do we want parsing to fail if the version is 2 or more?  If not we should add a default: case to the union.

src/core/crypto/relay_crypto.c
```diff
@@ -143,12 +143,9 @@ relay_decrypt_cell(circuit_t *circ, cell_t *cell,
             *recognized = 1;
             *layer_hint = thishop;
             /* Keep current digest of this cell for the possible SENDME. */
-            if (thishop->crypto.sendme_digest) {
-              crypto_digest_free(thishop->crypto.sendme_digest);
-            }
-            thishop->crypto.sendme_digest =
-              crypto_digest_dup(thishop->crypto.b_digest);
-
+            crypto_digest_get_digest(thishop->crypto.b_digest,
```
nmathewson: Hm. Why are we doing this on every cell that we decrypt?  It seems very wasteful; digests can be quite expensive.

src/core/crypto/relay_crypto.c
```diff
@@ -220,10 +217,9 @@ relay_encrypt_cell_inbound(cell_t *cell,
 {
   relay_set_digest(or_circ->crypto.b_digest, cell);
   /* Keep a record of this cell, we might use it for validating the SENDME. */
-  if (or_circ->crypto.sendme_digest) {
-    crypto_digest_free(or_circ->crypto.sendme_digest);
-  }
-  or_circ->crypto.sendme_digest = crypto_digest_dup(or_circ->crypto.b_digest);
+  crypto_digest_get_digest(or_circ->crypto.b_digest,
```
nmathewson: As above, do we really need to do this on every cell we encrypt?

I'm hoping that we can only generate digests when we actually need them.

src/core/or/sendme.c
```diff
@@ -555,16 +554,10 @@ sendme_note_cell_digest(circuit_t *circ)
...
-  if (TO_OR_CIRCUIT(circ)->crypto.sendme_digest) {
-    digest = tor_malloc_zero(TRUNNEL_SENDME_V1_DIGEST_LEN);
-    crypto_digest_get_digest(TO_OR_CIRCUIT(circ)->crypto.sendme_digest,
-                             (char *) digest, TRUNNEL_SENDME_V1_DIGEST_LEN);
-    if (circ->sendme_last_digests == NULL) {
-      circ->sendme_last_digests = smartlist_new();
-    }
-    smartlist_add(circ->sendme_last_digests, digest);
+  /* Add the digest to the last seen list in the circuit. */
+  digest = TO_OR_CIRCUIT(circ)->crypto.sendme_digest;
```
nmathewson: Let's have a function here instead that calculates the sendme digest for us from the relay_crypto.  That way:
  * we can get the digest through some different algorithm later on when other kinds of crypto are in use
  * we don't have to compute the digest with every cell
  * we don't have to make sendme.c know about the internals of relay_crypto.

src/core/or/relay.c
```diff
@@ -579,11 +581,20 @@ relay_send_command_from_edge_,(streamid_t stream_id, circuit_t *circ,
...
   /* Add random bytes to the unused portion of the payload, to foil attacks
    * where the other side can predict all of the bytes in the payload and thus
-   * compute authenticated sendme cells without seeing the traffic.  See
-   * proposal 289. */
+   * compute authenticated sendme cells without seeing the traffic. See
+   * proposal 289.
+   *
+   * We'll skip the first 4 bytes of unused data because having some unused
+   * zero bytes has saved us a lot of times in the past. */
+  random_bytes_len = RELAY_PAYLOAD_SIZE -
```
nmathewson: I'm a little scared by signed/unsigned things when we're checking for negative numbers.  Can we have a test for this?  It might help to put it in a separate function.  Or maybe it would be easier to wrap this part of the code in something like:
```
offset = RELAY_HEADER_SIZE + PAYLOAD_LEN + 4
if (offset < RELAY_PAYLOAD_SIZE) {
   len = RELAY_PAYLOAD_SIZE - offset;
   // ...
}
```

src/trunnel/sendme.trunnel
dgoulet-tor: I believe we do want to fail. A SENDME you can't parse (most likely unknown version) means you can't continue and thus the circuit is closed. This is the sendme cell validation code:

```
    if (sendme_cell_parse(&cell, cell_payload, cell_payload_len) < 0) {
      log_fn(LOG_PROTOCOL_WARN, LD_PROTOCOL,
             "Unparseable SENDME cell received. Closing circuit.");
      goto invalid;
    }
```

src/core/or/sendme.c
dgoulet-tor: Not sure I understand. Probably here you mean a function *in* `relay_crypto.c` that returns the SENDME digest so we can use it to match the one in the circuit (`circ->sendme_last_digests`)?

src/core/or/relay.c
dgoulet-tor: Sooooo... good thing you requested that because the code was actually not correct. The offset should have been tested against `CELL_PAYLOAD_SIZE` since we offset on `cell.payload`.

I've pushed a new commit that creates two helper function for the random padding and adds a unit test to test the offset value we use. I would recommend reviewing carefully here because it is very confusing to have `payload_len` not be the length of `cell.payload` but rather the *data* length.

See commit: 09f67b45e1f333a1d018d78d5311b0d54a8f2e7d

src/core/crypto/relay_crypto.c
dgoulet-tor: See above commits. Should fix this.

src/core/crypto/relay_crypto.c
dgoulet-tor: Indeed... and especially that we are doing exactly that in the `relay_digest_matches()` just before...

Ok so I took the approach of adding a function that knows when to expect a SENDME or not and if we are about to expect one, we'll keep the digest.

See extra commits:
- ff89dc2b0828ce45
- 3b1ea9548ee14e14
- e7d0a467de715896

src/core/or/sendme.c
nmathewson: right -- I mean that we shouldn't reach inside the crypto object here. Instead we should have an accessor function.

src/core/or/sendme.c
```diff
@@ -286,10 +286,37 @@ send_circuit_level_sendme(circuit_t *circ, crypt_path_t *layer_hint,
   return 0;
 }
 
+/* Put the crypto.b_digest in the sendme_digest. */
+static void
+note_cell_digest(const relay_crypto_t *crypto)
```
nmathewson: "Note" isn't quite right here -- let's try "record"?

nmathewson: (This should probably go into the relay_crypto module)

src/core/or/sendme.c
```diff
@@ -286,10 +286,37 @@ send_circuit_level_sendme(circuit_t *circ, crypt_path_t *layer_hint,
...
+}
+
 /*
  * Public API
  */
 
+/** Keep the current inbound cell digest for the next SENDME digest. This part
+ * is only done by the client as the circuit came back from the Exit. */
+void
+sendme_circuit_note_outbound_cell(or_circuit_t *or_circ)
```
nmathewson: This function's name makes it sound like it should get called for every cell.  Maybe sendme_circuit_record_outbound_digest() ?

src/core/or/sendme.c
```diff
@@ -286,10 +286,37 @@ send_circuit_level_sendme(circuit_t *circ, crypt_path_t *layer_hint,
...
+{
+  tor_assert(or_circ);
+  note_cell_digest(&or_circ->crypto);
+}
+
+/** Keep the current inbound cell digest for the next SENDME digest. This part
+ * is only done by the client as the circuit came back from the Exit. */
+void
+sendme_circuit_note_inbound_cell(crypt_path_t *cpath)
```
nmathewson: As above wrt function name

src/core/or/sendme.c
```diff
@@ -286,6 +286,29 @@ send_circuit_level_sendme(circuit_t *circ, crypt_path_t *layer_hint,
...
+ */
+
+/** Return true iff the next cell for the given cell window is expected to be
+ * a SENDME.
+ *
+ * We are able to know that because the package or deliver window value minus
+ * one cell (the possible SENDME cell) should be a multiple of the increment
+ * window value. */
+bool
+sendme_circuit_is_next_cell(int window)
```
nmathewson: Maybe next_cell_is_sendme would be a good name?

dgoulet-tor: I want to namespace it ... so it would end up a bit confusing like `sendme_next_cell_is_sendme()`. Also, the `circuit` in there I would like to keep because it cannot be used for a stream level SENDME check.

What about `sendme_circuit_cell_is_next()` ?

If you are ok with it, see fixup commit: 17636edb33]

src/core/or/sendme.c
dgoulet-tor: Same as above.

src/core/or/sendme.c
dgoulet-tor: Done in new commit:

src/core/or/sendme.c
dgoulet-tor: Moved it to `relay_crypto.c` and named `relay_crypto_record_sendme_digest()`

See new commit 4d6d633132

PR #739, opened 2019-02-27
[Tor Trac #17357]: rend_service_relaunch_rendezvous() ignores hs_service_requires_uptime_circ()

neelchauhan:
For the ticket [here](https://trac.torproject.org/projects/tor/ticket/17357).

### DIFF
--- a/src/feature/rend/rendservice.c
+++ b/src/feature/rend/rendservice.c
@@ -3012,6 +3012,10 @@ rend_service_relaunch_rendezvous(origin_circuit_t *oldcirc)
 {
   origin_circuit_t *newcirc;
   cpath_build_state_t *newstate, *oldstate;
+  const char *rend_pk_digest;
+  rend_service_t *service = NULL;
+
+  int flags = CIRCLAUNCH_NEED_CAPACITY | CIRCLAUNCH_IS_INTERNAL;
 
   tor_assert(oldcirc->base_.purpose == CIRCUIT_PURPOSE_S_CONNECT_REND);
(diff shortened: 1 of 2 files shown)


src/feature/rend/rendservice.c
```diff
@@ -3026,13 +3031,29 @@ rend_service_relaunch_rendezvous(origin_circuit_t *oldcirc)
   log_info(LD_REND,"Reattempting rendezvous circuit to '%s'",
            safe_str(extend_info_describe(oldstate->chosen_exit)));
 
+  /* Look up the service. */
+  rend_pk_digest = (char *) rend_data_get_pk_digest(oldcirc->rend_data, NULL);
+
+  base32_encode(serviceid, REND_SERVICE_ID_LEN_BASE32+1,
+                rend_pk_digest, REND_SERVICE_ID_LEN);
```
dgoulet-tor: If we can move this in the if() below because we only do that for logging.

src/feature/rend/rendservice.c
```diff
@@ -3026,13 +3031,29 @@ rend_service_relaunch_rendezvous(origin_circuit_t *oldcirc)
...
+
+  service = rend_service_get_by_pk_digest(rend_pk_digest);
+  if (!service) {
+    log_warn(LD_BUG, "Internal error: Trying to relaunch a rendezvous circ "
+                     "for an unrecognized service %s.", escaped(serviceid));
```
dgoulet-tor: At warning level here, we should `safe_str_client()` the serviceid.

PR #740, opened 2019-02-28
doc: update the man page and sample torrc for ExitRelay

teor2345:
We changed the default of ExitRelay in #21530 in 0.3.5.1-alpha, but
forgot to update the documentation.

Closes 29612.


teor2345:
Oops, wrong merge base.

PR #742, opened 2019-02-28
Bug29221 draft

asn-d6:


scripts/maint/practracker/metrics.py
```diff
@@ -0,0 +1,31 @@
...
+    for lineno, line in enumerate(f):
+        if not in_function:
+            # find the start of a function
+            m = re.match(r'^([a-zA-Z_][a-zA-Z_0-9]*),?\(', line)
+            if m:
+                func_name = m.group(1)
+                func_start = lineno
+                in_function = True
+        else:
+            # Fund the end of a function
```
nmathewson: Oops, this is my typo: this function should say "find", not "fund"

scripts/maint/practracker/metrics.py
```diff
@@ -8,6 +8,14 @@ def file_len(f):
         pass
     return i + 1
 
+def get_include_count(f):
+    """Get number of #include statements in the file"""
+    include_count = 0
+    for line in f:
+        if re.match(r' *# *include', line):
```
nmathewson: Maybe this should be `\s*` instead of ` *`, since tabs can also appear in this context. On the other hand, Tor doesn't allow tabs, so maybe we don't need to care.

scripts/maint/practracker/problem.py
```diff
@@ -0,0 +1,96 @@
...
+class IncludeCountProblem(Problem):
+    def __init__(self, problem_location, metric_value):
+        super(IncludeCountProblem, self).__init__("include-count", problem_location, metric_value)
+
+class FunctionSizeProblem(Problem):
+    def __init__(self, problem_location, metric_value):
+        super(FunctionSizeProblem, self).__init__("function-size", problem_location, metric_value)
+
+def get_old_problem_from_exception_str(exception_str):
+    try:
```
nmathewson: I'd suggest that we explicitly allow comments and empty lines in this file, and that we treat other unrecognized lines as errors.

scripts/maint/practracker/problem.py
```diff
@@ -0,0 +1,96 @@
...
+        return None
+
+    if problem_type == "file-size":
+        return FileSizeProblem(problem_location, metric_value)
+    elif problem_type == "include-count":
+        return IncludeCountProblem(problem_location, metric_value)
+    elif problem_type == "function-size":
+        return FunctionSizeProblem(problem_location, metric_value)
+    else:
+        print "Unknown exception line %s" % exception_str
```
nmathewson: Using stderr might make sense here; also, this won't work with python3. (Maybe you fix it later in the branch)

scripts/maint/practracker/metrics.py
```diff
@@ -16,21 +20,28 @@ def get_include_count(f):
...
 
-def function_lines(f):
+def get_function_lines(f):
     """
     Return iterator which iterates over functions and returns (function name, function lines)
     """
 
-    # XXX Buggy! Doesn't work with MOCK_IMPL and ENABLE_GCC_WARNINGS
+    # Skip lines with these terms since they confuse our regexp
+    REGEXP_CONFUSE_TERMS = ["MOCK_IMPL", "ENABLE_GCC_WARNINGS", "ENABLE_GCC_WARNING", "DUMMY_TYPECHECK_INSTANCE",
```
nmathewson: I'd suggest that you use a set here rather than an array.

scripts/maint/practracker/metrics.py
```diff
@@ -16,21 +20,28 @@ def get_include_count(f):
...
+                            "DISABLE_GCC_WARNING", "DISABLE_GCC_WARNINGS"]
+
     in_function = False
     for lineno, line in enumerate(f):
+        if any(x in line for x in REGEXP_CONFUSE_TERMS):
```
nmathewson: Instead of checking for the presence of the strings here, I'd suggest checking after we find the start of a function, to see if the function name is in the list of confusing terms.

scripts/maint/practracker/practracker.py
```diff
@@ -47,69 +47,88 @@
...
-        v = problem.IncludeCountProblem(fname, include_count)
-        ProblemVault.register_problem(v)
+        p = problem.IncludeCountProblem(fname, include_count)
+        return ProblemVault.register_problem(p)
+    return False
 
 def consider_function_size(fname, f):
-    for name, lines in metrics.function_lines(f):
+    """Consider the function sizes for 'f' and return True if a new issue was found"""
+    found_new_issues = False
```
nmathewson: Suggestion, possibly for a later ticket: use a counter here instead of a boolean. That way you can print the number of new issues on exit.

scripts/maint/practracker/util.py
```diff
@@ -14,7 +18,7 @@ def get_tor_c_files(tor_topdir, exclude_dirs):
 
             # Exclude the excluded paths
             full_path = os.path.join(root,filename)
-            if any(exclude_dir in full_path for exclude_dir in exclude_dirs):
+            if any(exclude_dir in full_path for exclude_dir in EXCLUDE_SOURCE_DIRS):
```
nmathewson: I don't like looking for the excluded path *everywhere* in the full_path string; it should only be excluded when it is right under TOP_SRCDIR.  In other words, if tor is in `~/tor/`, then `~/tor/src/trunnel/` should be excluded, but "~/src/tor/src/lib/handle_trunnel/src/trunnel` shouldn't be excluded

scripts/maint/practracker/practracker.py
```diff
@@ -131,8 +131,10 @@ def main():
     # 3) Go through all the files and report problems if they are not exceptions
     found_new_issues = consider_all_metrics(files_list)
 
+    # If new issues were found, try to give out some advice to the developer on how to resolve it.
     if (found_new_issues):
-        print("practracker FAILED")
+        new_issues_str = "practracker FAILED as indicated by the problem lines above. Please use the exceptions file ({}) to find any previous state of these problems. If you are unable to fix the underlying best-practices issue right now then you need to either update the relevant exception line or add a new one.".format(exceptions_file)
```
nmathewson: With your permission, I'll tweak this string a little post-merge.  The way it is written now, it suggests that the preferred solution to a bad practice is to add an exception, rather than to fix the problem.  I think there should be a new section in doc/HACKING about this, with suggestions for how to fix each kind of problem.

scripts/maint/practracker/metrics.py
asn-d6: Fixed in ac05cc7437.

scripts/maint/practracker/metrics.py
asn-d6: Included this improvement in the post-merge ticket #29746, since it's a good-to-have but doesn't actually influence results atm.

scripts/maint/practracker/problem.py
asn-d6: Included this improvement in the post-merge ticket #29746, since it's a good-to-have but doesn't actually influence results atm.

scripts/maint/practracker/problem.py
asn-d6: Yep, that line was fixed later in the branch, but I used stderr for another err message in ac05cc7437.

scripts/maint/practracker/metrics.py
asn-d6: Fixed in ac05cc7437.

scripts/maint/practracker/metrics.py
asn-d6: Added to post-merge ticket #29746.

scripts/maint/practracker/practracker.py
asn-d6: Added to post-merge ticket #29746.

scripts/maint/practracker/util.py
asn-d6: I tried to do this, but I think it should be done properly with tests to make sure of it's correctness.
Added to post-merge ticket #29746.

scripts/maint/practracker/practracker.py
asn-d6: Yes, no problem with tweaking that string.

scripts/maint/practracker/metrics.py
nmathewson: I think this one is actually important for correctness; I'll try a quick fix.  See e968b886aca66a in my bug29221_more

nmathewson:
eventually merged as #787

PR #744, opened 2019-03-01
Log the correct "auto" port number for listening sockets

teor2345:
When "auto" was used for the port number for a listening socket, the
message logged after opening the socket would incorrectly say port 0
instead of the actual port used.

Fixes bug 29144; bugfix on 0.3.5.1-alpha

Signed-off-by: Kris Katterjohn <katterjohn@gmail.com>


teor2345:
Oops, this is the wrong merge head. See #745.

PR #750, opened 2019-03-01
#29588:  Add post-merge git hook script to warn about git related scripts being updated

rl1987:
https://trac.torproject.org/projects/tor/ticket/29588


scripts/maint/post-merge.git-hook
```diff
@@ -13,12 +13,20 @@ check_for_diffs() {
...
+        fullpath="$git_toplevel/scripts/maint/$1"
+
+        git diff ORIG_HEAD HEAD --exit-code -- "$fullpath"
+}
+
 check_for_diffs "pre-push"
 check_for_diffs "pre-commit"
 check_for_diffs "post-merge"
 
+check_for_script_update "git-merge-forward.sh"
```
nmathewson: Can we use the strategy of "check_for_script_update" for all of the files?

scripts/maint/post-merge.git-hook
```diff
@@ -13,12 +13,20 @@ check_for_diffs() {
...
                         diff "$installed" "$latest"
-                        exit 1
                fi
         fi
 }
 
+check_for_script_update() {
+        fullpath="$git_toplevel/scripts/maint/$1"
+
+        git diff ORIG_HEAD HEAD --exit-code -- "$fullpath"
```
nmathewson: Will `--exit-code` make this script fail if a file has changed?  Maybe we should let the script succeed uniformly, and just warn the user.

On a similar note, we should probably display a helpful message here, and not just a raw diff.

scripts/maint/post-merge.git-hook
```diff
@@ -0,0 +1,24 @@
...
+        installed="$git_toplevel/.git/hooks/$1"
+        latest="$git_toplevel/scripts/maint/$1.git-hook"
+
+        if [ -e "$installed" ]
+        then
+               if ! cmp "$installed" "$latest" >/dev/null 2>&1
+               then
+                        echo "ATTENTION: $1 hook has changed:"
+                        echo "==============================="
+                        diff "$installed" "$latest"
```
nmathewson: If we keep this line, it should probably be `diff -u`

scripts/maint/post-merge.git-hook
nmathewson: Maybe we should move all of these files into a scripts/git directory, and apply this hook to every file in the directory.  That way, it won't need to list all of the hooks and scripts that we use

scripts/maint/post-merge.git-hook
rl1987: Fixed in 92b4d17a64e3dbb6fb6880d9452abdae95f67a5b.

scripts/maint/post-merge.git-hook
rl1987: Did so in 48a3384df80d3833e99adf246971bdb9a491521b.

scripts/maint/post-merge.git-hook
rl1987: Moving the script files to new directory in 3e895fbb504edea174d21b6e8c8b4be02cd636c7 and iterating through them in 7f87393c6c8bc9eee81ab2ef4643112cd012c5d9.

scripts/git/post-merge.git-hook
```diff
@@ -39,6 +39,7 @@ check_for_diffs "pre-push"
 check_for_diffs "pre-commit"
 check_for_diffs "post-merge"
 
-check_for_script_update "git-merge-forward.sh"
-check_for_script_update "git-pull-all.sh"
-check_for_script_update "git-push-all.sh"
+for file in ls $git_toplevel/scripts/ ; do
```
nmathewson: This won't work -- it will first set file to "ls" and then to "$ggit_toplevel/scripts/".

Please make sure you test your code; it will save time in the long run.

scripts/maint/post-merge.git-hook
```diff
@@ -28,7 +28,11 @@ check_for_diffs() {
 check_for_script_update() {
         fullpath="$git_toplevel/scripts/maint/$1"
 
-        git diff ORIG_HEAD HEAD --exit-code -- "$fullpath"
+        if ! git diff ORIG_HEAD HEAD --exit-code -- "$fullpath" >&2
```
nmathewson: What is >&2 doing here?  Why are we sending this to stderr?

PR #788, opened 2019-03-13
Bug28636

asn-d6:


src/lib/crypt_ops/crypto_rand_numeric.c
```diff
@@ -154,6 +154,28 @@ crypto_fast_rng_get_uint64(crypto_fast_rng_t *rng, uint64_t limit)
                   crypto_fast_rng_getbytes(rng, (void*)&val, sizeof(val)));
 }
 
+/**
+ * As crypto_rand_u32, but extract the result from a crypto_fast_rng_t.
+ */
+uint32_t
+crypto_fast_rng_get_u32(crypto_fast_rng_t *rng)
+{
+  IMPLEMENT_RAND_UNSIGNED(uint32_t, UINT32_MAX, UINT32_MAX,
```
nmathewson: This won't actually work so well -- the "limit" argument has to be one more than the largest possible output. 
Instead, just do something like
    uint32_t val;
    crypto_fast_rng_getbytes(rng, (void*)&val, sizeof(val)));
    return val;

src/core/or/circuitpadding.c
```diff
@@ -965,13 +966,19 @@ circpad_send_padding_cell_for_callback(circpad_machine_state_t *mi)
...
-  // Make sure circuit didn't close on us
+  /* Make sure circuit didn't close on us */
   if (mi->on_circ->marked_for_close) {
     log_fn(LOG_INFO,LD_CIRC,
            "Padding callback on a circuit marked for close. Ignoring.");
     return CIRCPAD_STATE_CHANGED;
   }
 
+  /* Don't send padding if we are currently in dormant mode. */
+  if (accounting_tor_is_dormant()) {
```
nmathewson: This isn't actually the kind of dormant that you need to look at here:  it isn't about bandwidth accounting.  Instead see src/core/mainloop/netstatus.c.

Also, I am not sure if this change is enough for what we need here -- we need to make sure that all the timers stop waking up while we are dormant, or else we will waste CPU.  Is this tweak sufficient to ensure that?

asn-d6: Yikes, I guess I hit the wrong dormant, using the one in `netstatus.c` seems also easy because it has an API for checking whether it's enabled.

However, I did not manage to find what's the purpose of this dormant mode, and from reading the `DormantClientTimeout` man page it seemed to me that the goal is reducing bandwidth, and not so much CPU etc.

If we also care about CPU, the simple patch above won't work perfectly, since it does not stop the timers from waking up or the special event callbacks from triggering (e.g. circpad_machine_event_circ_built()).

However:
a) This simple patch IIUC will stop extra padding timers from being installed on a circuit since it stops `circpad_send_padding_cell_for_callback()` in its tracks and that's where subsequent timers get installed through `circpad_cell_event_padding_sent() -> circpad_machine_spec_transition`. So only the first timer will trigger, no padding will be sent, and then no more padding or timers IIUC (checking with Mike would be good).
b) If dormant mode is enabled, then no additional circuits will be created, and hence we only care about padding and timers of the circuits we already have, and if point (a) from above is correct then there is not much problem.

src/lib/crypt_ops/crypto_rand_numeric.c
asn-d6: Pushed fixup as c83fedb284.

src/core/or/circuitpadding.c
asn-d6: Will ask Mike about this in today's meeting.

nmathewson: FWIW, the main motivation for dormant mode is to save resource consumption when Tor is not in use.  For all clients, this includes bandwidth, but for mobile users, it includes CPU, for which wakeups are relevant.

src/core/or/circuitpadding.c
```diff
@@ -1165,6 +1159,12 @@ circpad_machine_schedule_padding,(circpad_machine_runtime_t *mi))
   struct timeval timeout;
   tor_assert(mi);
 
+  /* Don't schedule padding if we are currently in dormant mode. */
+  if (net_is_completely_disabled()) {
```
nmathewson: This isn't a test for dormant mode either; instead look at is_participating_on_network()?

src/core/or/circuitpadding.c
```diff
@@ -1160,7 +1160,7 @@ circpad_machine_schedule_padding,(circpad_machine_runtime_t *mi))
   tor_assert(mi);
 
   /* Don't schedule padding if we are currently in dormant mode. */
-  if (net_is_completely_disabled()) {
+  if (is_participating_on_network()) {
```
nmathewson: This test is backwards.  We want to not schedule padding if we are _not_ participating on the network.

teor2345:
Squashed in #819.

PR #793, opened 2019-03-13
#29702:  Use empty torrc file when launching tor in test_rebind.py

rl1987:
https://trac.torproject.org/projects/tor/ticket/29702


teor2345:
Rebased to 0.3.5 as #924.

PR #803, opened 2019-03-18
Run `make reset-gcov` after the script target in Travis CI is done.

ahf:
See: https://bugs.torproject.org/29036


ahf:
This needs to be run against 0.2.9.

PR #804, opened 2019-03-18
#29073: Fix shellcheck warnings in linux-tor-prio.sh (2)

rl1987:
https://trac.torproject.org/projects/tor/ticket/29073


teor2345:
We'll merge #781 and remove this script, rather than fixing it.

PR #805, opened 2019-03-18
bwauth: Ignore bandwidth file lines with "vote=0"

juga0:
so that the relays that would be "excluded" from the bandwidth
file because of something failed can be included to diagnose what
failed, without still including these relays in the bandwidth
authorities vote.

Closes #29806.


juga0:
PR against wrong branch, created https://github.com/torproject/tor/pull/807 instead.

PR #809, opened 2019-03-19
bwauth: Ignore bandwidth file lines with "vote=0"

juga0:
so that the relays that would be "excluded" from the bandwidth
file because of something failed can be included to diagnose what
failed, without still including these relays in the bandwidth
authorities vote.

Closes #29806.

### DIFF
--- a/changes/ticket29806
+++ b/changes/ticket29806
@@ -1,0 +1,7 @@
+  o Minor features (bandwidth authority):
+    - Make bandwidth authorities to ignore relays that are reported in the
+      bandwidth file with the key-value "vote=0".
+      This change allows to report the relays that were not measured due
+      some failure and diagnose the reasons without being included in the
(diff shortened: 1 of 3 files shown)


changes/ticket29806
```diff
@@ -0,0 +1,7 @@
+  o Minor features (bandwidth authority):
+    - Make bandwidth authorities to ignore relays that are reported in the
+      bandwidth file with the key-value "vote=0".
+      This change allows to report the relays that were not measured due
+      some failure and diagnose the reasons without being included in the
```
teor2345: This change only affects the bandwidth in the vote:
```suggestion
      some failure and diagnose the reasons without the bandwidth being included in the
```

teor2345:
Merged from 0.3.4 and now in #821.

PR #810, opened 2019-03-19
B22029

asn-d6:


src/feature/dirauth/process_descs.c
```diff
@@ -121,6 +128,40 @@ add_fingerprint_to_dir(const char *fp, authdir_config_t *list,
   return;
 }
 
+/** Add the ed25519 key <b>edkey</b> to the smartlist of fingerprint_entry_t's
+ * <b>list</b>, or-ing the currently set status flags with
+ * <b>add_status</b>.
+ */
+/* static */ void
```
teor2345: I think this function can be declared static.

src/feature/dirauth/process_descs.c
```diff
@@ -121,6 +128,40 @@ add_fingerprint_to_dir(const char *fp, authdir_config_t *list,
...
+add_ed25519_to_dir(const char *edkey, authdir_config_t *list,
+                   router_status_t add_status)
+{
+  char *ed25519_key;
+  char d[DIGEST_LEN];
```
teor2345: d is 160 bits long, but digest256_from_base64() expects a 256-bit array.

A unit test would have discovered this memory safety issue.

src/feature/dirauth/process_descs.c
```diff
@@ -121,6 +128,40 @@ add_fingerprint_to_dir(const char *fp, authdir_config_t *list,
...
+  router_status_t *status;
+  tor_assert(edkey);
+  tor_assert(list);
+
+  ed25519_key = tor_strdup(edkey);
+  tor_strstrip(ed25519_key, " ");
```
teor2345: Strip spaces *before* checking the length of the string, not after.

src/feature/dirauth/process_descs.c
```diff
@@ -237,9 +296,9 @@ dirserv_router_get_status(const routerinfo_t *router, const char **msg,
   }
 
   /* Check for the more usual versions to reject a router first. */
-  const uint32_t r = dirserv_get_status_impl(d, router->nickname,
-                                             router->addr, router->or_port,
-                                             router->platform, msg, severity);
+  uint32_t r = dirserv_get_status_impl(d, router->nickname, router->addr,
```
teor2345: I don't understand why you removed this const.
If you're going to make unrelated changes, make them in a separate commit?

src/feature/dirauth/process_descs.c
```diff
@@ -272,6 +331,14 @@ dirserv_router_get_status(const routerinfo_t *router, const char **msg,
         }
         return FP_REJECT;
       }
+
+      /* Check status with ed25519 key. */
+      r = dirserv_get_status_impl(
+              (char *) router->cache_info.signing_key_cert->signing_key.pubkey,
```
teor2345: dirserv_get_status_impl() does not accept ed25519 keys.

teor2345: nickm says: also, signing_key_cert can be null, so we shouldn't dereference it unconditionally

src/feature/dirauth/process_descs.c
```diff
@@ -174,27 +215,58 @@ dirserv_load_fingerprint_file(void)
...
+    int is_ed25519 = (strlen(fingerprint) == BASE64_DIGEST256_LEN);
+    int is_rsa = (strlen(fingerprint) == HEX_DIGEST_LEN);
+    bool is_valid_fpr = true;
+
+    /* We only accept ed25519 or RSA keys */
+    if (!is_ed25519 && !is_rsa) {
+      is_valid_fpr = false;
+    }
+
+    tor_assert(is_ed25519 || is_rsa);
```
teor2345: This line will crash the authority without a useful warning if an approved_routers line is malformed.
Please don't do that.

teor2345: Instead, log a warning and continue with the next line in the file.

src/feature/dirauth/process_descs.c
```diff
@@ -121,6 +128,40 @@ add_fingerprint_to_dir(const char *fp, authdir_config_t *list,
   return;
 }
 
+/** Add the ed25519 key <b>edkey</b> to the smartlist of fingerprint_entry_t's
+ * <b>list</b>, or-ing the currently set status flags with
+ * <b>add_status</b>.
+ */
+/* static */ void
+add_ed25519_to_dir(const char *edkey, authdir_config_t *list,
```
teor2345: nickm says: (Also, passing around untyped arrays of bytes is questionable; it's probably best that anything that needs an ed25519 key should take the ed25519_public_key_t structure)

PR #818, opened 2019-03-20
#29824: Fix CID 1444119

rl1987:
Let's use the same function exit point for BUG() codepath that we're using
for every other exit condition. That way, we're not forgetting to clean up
the memarea.

https://trac.torproject.org/projects/tor/ticket/29824


teor2345:
Cherry-picked to 0.3.4 as #848.

PR #825, opened 2019-03-21
prob-distr: Silence some coverity warnings.

asn-d6:


src/lib/math/prob_distr.h
```diff
@@ -19,17 +19,105 @@ struct dist {
...
+ *
+ *   struct geometric mydist = {
+ *     DIST_BASE(&weibull_ops),
+ *     ...
+ *   };
+ */
 #define DIST_BASE(OPS)  { .ops = (OPS) }
+
+/** A compile-time type-checking macro for use with DIST_BASE_TYPED.
+ *
```
tlyu: I would add something like `This macro works by checking that &OBJ is a pointer type that is the same type (except for qualifiers) as (const TYPE *)&OBJ. It's a C constraint violation (which requires a diagnostic) if two pointers are different types and are subtracted. The sizeof() forces compile-time evaluation, and the multiplication by zero is to discard the result of the sizeof() from the expression.`

asn-d6: Useful addition. Pushed it on e86564e.

PR #826, opened 2019-03-21
[Tor Trac #24546]: Use tor_addr_is_v4() rather than family, or reject all v6-mapped IPv4 addresses (Revision 1)

neelchauhan:
For the ticket [here](https://trac.torproject.org/projects/tor/ticket/24546).


src/app/config/config.c
```diff
@@ -7653,8 +7653,7 @@ get_first_listener_addrport_string(int listener_type)
     if (cfg->server_cfg.no_listen)
       continue;
 
-    if (cfg->type == listener_type &&
-        tor_addr_family(&cfg->addr) != AF_UNSPEC) {
+    if (cfg->type == listener_type && tor_addr_is_unspec(&cfg->addr)) {
```
teor2345: This change tests for unspec, it used to test for not unspec.
Maybe you should use an automated tool like coccinelle for this commit?

src/core/mainloop/connection.c
```diff
@@ -3040,7 +3040,7 @@ connection_is_rate_limited(connection_t *conn)
   if (conn->linked)
     return 0; /* Internal connection */
   else if (! options->CountPrivateBandwidth &&
-           (tor_addr_family(&conn->addr) == AF_UNSPEC || /* no address */
+           (tor_addr_is_unspec(&conn->addr)         || /* no address */
```
teor2345: I don't understand the extra spaces here. But they're not a big deal.

src/test/test_addr.c
```diff
@@ -362,50 +362,6 @@ test_addr_ip6_helpers(void *arg)
   test_external_ip("2001::", 0);
   test_external_ip("ffff::", 0);
 
-  test_external_ip("::ffff:0.0.0.0", 1);
```
teor2345: Why did you delete all these tests?

neelchauhan: They were IPv6-mapped IPv4 addresses

src/app/config/config.c
neelchauhan: I fixed this and pushed it.
For future tickets similar to this (in terms of refactoring), I will look into coccinelle.

PR #827, opened 2019-03-21
bug28925 Correctly report PT vs proxy during bootstrap

tlyu:
Previously, or_connection_t did not record whether or not the
connection uses a pluggable transport. Instead, it stored the
underlying proxy protocol of the pluggable transport in
proxy_type. This made bootstrap reporting treat pluggable transport
connections as plain proxy connections.

Store a separate bit indicating whether a pluggable transport is in
use, and decode this during bootstrap reporting.

Fixes bug 28925; bugfix on 0.4.0.1-alpha.

### DIFF
--- a/src/core/mainloop/connection.c
+++ b/src/core/mainloop/connection.c
@@ -5368,7 +5368,7 @@ assert_connection_ok(connection_t *conn, time_t now)
  */
 int
 get_proxy_addrport(tor_addr_t *addr, uint16_t *port, int *proxy_type,
-                   const connection_t *conn)
+                   int *is_pt, const connection_t *conn)
 {
   const or_options_t *options = get_options();
 
@@ -5387,6 +5387,7 @@ get_proxy_addrport(tor_addr_t *addr, uint16_t *port, int *proxy_type,
       tor_addr_copy(addr, &transport->addr);
       *port = transport->port;
       *proxy_type = transport->socks_version;
+      *is_pt = 1;
       return 0;
     }
 
@@ -5423,11 +5424,13 @@ log_failed_proxy_connection(connection_t *conn)
 {
   tor_addr_t proxy_addr;
   uint16_t proxy_port;
-  int proxy_type;
+  int proxy_type, is_pt;
 
-  if (get_proxy_addrport(&proxy_addr, &proxy_port, &proxy_type, conn) != 0)
+  if (get_proxy_addrport(&proxy_addr, &proxy_port, &proxy_type, &is_pt,
+                         conn) != 0)
     return; /* if we have no proxy set up, leave this function. */
 
+  (void)is_pt;
   log_warn(LD_NET,
            "The connection to the %s proxy server at %s just failed. "
            "Make sure that the proxy server is up and running.",
--- a/src/core/mainloop/connection.c
+++ b/src/core/mainloop/connection.c
@@ -5361,17 +5361,20 @@ assert_connection_ok(connection_t *conn, time_t now)
 }
 
 /** Fills <b>addr</b> and <b>port</b> with the details of the global
- *  proxy server we are using.
- *  <b>conn</b> contains the connection we are using the proxy for.
+ *  proxy server we are using. Store a 1 to the int pointed to by
+ *  <b>is_put_out</b> if the connection is using a pluggable
+ *  transport; store 0 otherwise. <b>conn</b> contains the connection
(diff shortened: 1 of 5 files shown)


src/core/mainloop/connection.c
```diff
@@ -5368,7 +5368,7 @@ assert_connection_ok(connection_t *conn, time_t now)
  */
 int
 get_proxy_addrport(tor_addr_t *addr, uint16_t *port, int *proxy_type,
-                   const connection_t *conn)
+                   int *is_pt, const connection_t *conn)
```
nmathewson: By convention, let's name this `is_pt_out`.  Let's also document it.

src/core/mainloop/connection.c
```diff
@@ -5387,6 +5387,7 @@ get_proxy_addrport(tor_addr_t *addr, uint16_t *port, int *proxy_type,
       tor_addr_copy(addr, &transport->addr);
       *port = transport->port;
       *proxy_type = transport->socks_version;
+      *is_pt = 1;
```
nmathewson: Does anything set *is_pt to 0 when we are _not_ a PT?  It would make sense to do that IMO.

src/core/mainloop/connection.h
```diff
@@ -187,7 +187,7 @@ int connection_proxy_connect(connection_t *conn, int type);
 int connection_read_proxy_handshake(connection_t *conn);
 void log_failed_proxy_connection(connection_t *conn);
 int get_proxy_addrport(tor_addr_t *addr, uint16_t *port, int *proxy_type,
-                       const connection_t *conn);
+                       int *is_pt, const connection_t *conn);
```
nmathewson: as above about naming

src/core/or/connection_or.c
```diff
@@ -1516,13 +1524,15 @@ connection_or_connect, (const tor_addr_t *_addr, uint16_t port,
...
-  r = get_proxy_addrport(&proxy_addr, &proxy_port, &proxy_type, TO_CONN(conn));
+  r = get_proxy_addrport(&proxy_addr, &proxy_port, &proxy_type, &is_pt,
+                         TO_CONN(conn));
   if (r == 0) {
     conn->proxy_type = proxy_type;
     if (proxy_type != PROXY_NONE) {
       tor_addr_copy(&addr, &proxy_addr);
       port = proxy_port;
       conn->base_.proxy_state = PROXY_INFANT;
+      conn->is_pt = is_pt;
```
nmathewson: We should either use the "!!" trick here to force "is_pt" to 0 or 1, or we should document in get_proxy_addrport that is_pt will be set to 0 or 1.

src/core/or/connection_or.c
```diff
@@ -437,7 +437,15 @@ connection_or_state_publish(const or_connection_t *conn, uint8_t state)
 
   msg.type = ORCONN_MSGTYPE_STATE;
   msg.u.state.gid = conn->base_.global_identifier;
-  msg.u.state.proxy_type = conn->proxy_type;
+  if (conn->is_pt) {
+    /* Do extra decoding because conn->proxy_type indicates the proxy
+     * protocol that tor uses to talk with the transport plugin,
+     * instead of PROXY_PLUGGABLE. */
+    tor_assert(conn->proxy_type != PROXY_NONE);
```
nmathewson: Can we make this tor_assert_nonfatal() ? I'd hate to introduce a new crash bug at this stage in the development process.

src/core/mainloop/connection.c
tlyu: I think this happens downstream (in the `or_connection_t`) as a side effect of `tor_alloc_zero()`, which I think we rely on for the other flag bits in that structure. I can add an explicit setting of zero for robustness.

src/core/or/connection_or.c
tlyu: Sure.

src/core/mainloop/connection.h
tlyu: OK

src/core/or/connection_or.c
tlyu: I'd rather go the documentation route. We could also have `is_pt` be a `bool`, but that conflicts with surrounding practice.

nmathewson: ok

teor2345:
Squashed in #836, merged forward to master in #837.

PR #843, opened 2019-03-25
Practracker regen

nmathewson:


scripts/maint/practracker/practracker.py
```diff
@@ -113,6 +113,38 @@ def consider_metrics_for_file(fname, f):
 
     return found_new_issues
 
+HEADER="""\
```
teor2345: Nitpick: there's a typo in the commit message summary: excptions.
But it's clear what you meant.

PR #868, opened 2019-03-29
Bug28634 draft

asn-d6:


src/core/or/circuitpadding_machines.c
```diff
@@ -0,0 +1,351 @@
...
+   *
+   * So with about 200 milliseconds latency between each cell, we appropriately
+   * set the histogram edges to from 30 milliseconds to 100 milliseconds, to
+   * account for faster connections.
+   */
+
+  /* Histogram is: (30 msecs, 200 msecs, infinity). */
+  burst_state->histogram_len = 2;
+  burst_state->histogram_edges[0] = 30000;
+  burst_state->histogram_edges[1] = 200000;
```
mikeperry-tor: One problem we'll run in to in research is that this pattern is not going to be in any of their "open worlds" that they train their classifiers against. So even though an argument can be made that this pattern could very well look like an SSH session, or IRC window, or Signal Chat, they won't have that in their testing set and declare this defense broken as a result (because we've just transformed one very unique intro pattern into another "very unique" random pattern that doesn't match anything else, and so their classifiers will still work).

Another option is to make a second state for this machine on the relay side that it transitions to on  receipt of a cell from the client, and have that state send something that looks like a short HTTP response (ie like a handful of packets), before transitioning back to the first state. If we also set use_rtt_estimate=1 on that first state, we should have something that looks like a set of HTTP fetches, which might match something from their open world. I can give an example of this if that doesn't make sense.

Such a machine might be higher overhead, though. It really depends on how long the responses are and how many we do.

What do you think?

src/core/or/circuitpadding_machines.c
```diff
@@ -0,0 +1,351 @@
...
+/* Setup the BURST state of the machine that hides client-side rend
+ * circuits. If <b>is_client</b> is true, then this is the client-side part of
+ * this machine, otherwise it's the service-side of this machine. */
+static void
+setup_burst_state_for_hiding_rend_circuits(circpad_state_t *burst_state,
+                                           bool is_client)
+{
+  /* Token removal strategy for BURST state */
+  burst_state->token_removal =
+      CIRCPAD_TOKEN_REMOVAL_CLOSEST;
```
mikeperry-tor: Again, using CIRCPAD_TOKEN_REMOVAL_NONE will avoid lots of needless mallocs.

src/core/or/circuitpadding_machines.c
```diff
@@ -0,0 +1,351 @@
...
+  /* For this machine we only pad in client-side introduction circuits even
+   * after introduction has finished */
+  client_machine->conditions.purpose_mask =
+    circpad_circ_purpose_to_mask(CIRCUIT_PURPOSE_C_INTRODUCING)|
+    circpad_circ_purpose_to_mask(CIRCUIT_PURPOSE_C_INTRODUCE_ACK_WAIT)|
+    circpad_circ_purpose_to_mask(CIRCUIT_PURPOSE_C_INTRODUCE_ACKED);
+
+  /* XXX Keep the circuit alive even after the introduction has been finished,
+   * otherwise the short-term lifetime of the circuit will blow our cover */
+  // client_machine->manage_circ_lifetime = 1;
```
mikeperry-tor: Note that we also want to set client_machine->should_negotiate_end = 1 on the client side here, since the relay side will not be managing lifetime (as you say below).

src/core/or/circuitpadding_machines.c
```diff
@@ -0,0 +1,351 @@
...
+{
+  /* Token removal strategy for BURST state */
+  burst_state->token_removal =
+      CIRCPAD_TOKEN_REMOVAL_CLOSEST;
+
+  /* Histogram for BURST state:
+   *
+   * We want a machine that will send a bunch of fake cells over a reasonable
+   * time frame, to obfuscate the fact that introduction circuits always send 4
+   * cells both ways */
```
mikeperry-tor: This comment seems to be cut+paste from the introduction machine? Needs update.

src/core/or/circuitpadding_machines.c
```diff
@@ -0,0 +1,351 @@
...
+  burst_state->length_dist.param2 = is_client ?
+    HS_MACHINE_PADDING_MAXIMUM_CLIENT : HS_MACHINE_PADDING_MAXIMUM_SERVICE;
+
+  /* Histogram is: (30 msecs, 200 msecs, infinity).
+   * See analysis above.
+  */
+  burst_state->histogram_len = 2;
+  burst_state->histogram_edges[0] = 30000;
+  burst_state->histogram_edges[1] = 200000;
+
```
mikeperry-tor: I think that doing the HTTP get thing here might help also.

And again we have to be careful about creating a thing that is now obviously unique. If we start padding before at least 3 hops are built, for example, then only padded intro/rend circs will be the ones with mismatched EXTEND2/EXTENDED2 request/response pattern. We can enforce the "don't pad until three hops" situation with a small histogram that just counts non-padding packets as our first state..

src/core/or/circuitpadding.c
```diff
@@ -2602,6 +2602,36 @@ circpad_handle_padding_negotiated(circuit_t *circ, cell_t *cell,
...
+STATIC void
+machine_spec_free_(circpad_machine_spec_t *m)
+{
+  if (!m) return;
+
+  tor_free(m->states);
+  tor_free(m);
+}
+
+/** Free all memory allocated by the circuitpadding subsystem. */
```
mikeperry-tor: We should comment somewhere that circpad machine_specs must all be malloced now, if this function is used.

I think our unittests actually add some non-malloc'd machines from global static allocation to the smartlist..

src/core/or/circuitpadding_machines.c
asn-d6: I definitely agree we are going from a unique pattern to a different unique pattern here and this is a problem. That's why I kinda feel like this is like a poor man's obfs2 for circuit fingerprinting.

image

Just to go a bit deeper into how useful an obfs2 would be here, if we check the backgammon board above, we can see that the top rows of the board are all pretty weirdly looking general circuits. Most of them are either web circuits or directory fetching ones, that don't fit into the usual general pattern (which is the next-to-last one with 48 counts). Perhaps our silly machine could fit into those for now, and try to blend in with the chaos.

--- 
I think the HTTP-simulation could be a plausible way forward for something more sophisticated, but I'm afraid that it's gonna increase the bandwidth, and except if we do it really really methodically, it's still gonna look weird to an observer. I think to make it look semi-legit we would need to impose lots of overhead to every circuit, and also create a proper state machine that simulates HTTP.

I think our main opponent here is the overhead otherwise we could really experiment with such solutions. I did an initial overhead analysis here which left me a bit concerned: https://github.com/torproject/torspec/blob/eb77f96b58fac76a606c5264ac83f68f71902c00/proposals/ideas/xxx-padding-machines-onions.txt#L173
I would be curious to hear what you think, especially since you are much more experienced into thinking about overhead etc.

--- 

Finally, I uploaded the python script I use to make the backgammon board with some instructions. Now you can make your own: https://github.com/asn-d6/padanalyzer

Thanks for the review and the thoughts.

src/core/or/circuitpadding_machines.c
```diff
@@ -0,0 +1,351 @@
...
+}
+
+/* Setup the BURST state of the machine that hides client-side intro
+ * circuits. */
+static void
+setup_burst_state_for_hiding_intro_circuits(circpad_state_t *burst_state,
+                                            bool is_client)
+{
+  /* Token removal strategy for BURST state */
+  burst_state->token_removal = CIRCPAD_TOKEN_REMOVAL_CLOSEST;
```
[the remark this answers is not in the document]
asn-d6: Agreed! BTW, there is a fun bug that needs to be resolved before we can do this. In particular, this block of code triggers https://github.com/torproject/tor/blob/15e0f36fa0a17a82aaa9684895447237234b28c6/src/core/or/circuitpadding.c#L339
and destroys the histogram right when we transition from START to BURST.

Any ideas on how this should be fixed? Also I remember you mentioned that a simplification was possible in that function, but we never really did it. Perhaps this is a good time along with the bugfix?

src/core/or/circuitpadding_machines.c
mikeperry-tor: So one lightweight option is to use Ticket #30092 (PR #916)  to make some fraction of general circuits look like this crazyness we're going to do with this circuits.

Another option is to implement https://trac.torproject.org/projects/tor/ticket/29203 to allow us to specify some padding machines as reduced padding, and do both the HTTP approach and the janky approach.

A third option is to just implement whatever and monitor padding overhead results, and if they get too high, use our existing padding overhead limits to crank them back down. That might need a fix to https://trac.torproject.org/projects/tor/ticket/29231 though..

I'm starting to sense the need for a bunch of child tickets here.

src/core/or/circuitpadding_machines.c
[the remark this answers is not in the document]
mikeperry-tor: No, this only destroys the mutable copy of the histogram that is in the machine runtime from any previous state (in this case there was no histogram in START, so it should already be NULL). It also prevents us from allocating another mutable copy from the BURST state. If the machine runtime has no histrogram copy, the code still uses the readonly copy in the machine_spec_t. (This is part of the reason why I wanted to differentiate between mutable info and readonly info in the code/naming.. some stuff does not need to be in the runtime for it to run.. it only has to be there if it is mutable).

https://github.com/torproject/tor/blob/15e0f36fa0a17a82aaa9684895447237234b28c6/src/core/or/circuitpadding.c#L453

and

https://github.com/torproject/tor/blob/15e0f36fa0a17a82aaa9684895447237234b28c6/src/core/or/circuitpadding.c#L894

src/core/or/circuitpadding_machines.c
asn-d6: Yes, that's a good idea! If we let intro circs build the circuit up to the intro point before padding, then the extra **EXTEND**->EXTENDED roundtrip (bold is outgoing) will blend in with the **BEGIN**->CONNECTED roundtrip that most normal circuits have (if you look at the backgammon board). We can start padding after that and then we will blend much better since then the **NEGOTIATE_PADDING**->INTRO1 will also blend with the **DATA**->DATA that most normal circuits have. 

Similarly for rend circuits, we should wait until the rendezvous circuit is established so that the **ESTABLISH_REND**->REND_ESTABLISHED dance, blends in with the **BEGIN**->CONNECTED dance of normal circuits.

asn-d6: I have some good results by following this lead. I basically managed to hide the initial cell sequence of intro circs inside general circs by manipulating the allowed purposes: https://github.com/torproject/tor/pull/934/commits/90a81fca5736f165c9568375bf0205b5e9e9ca84#diff-c1ce03b78513b52ceac68cfe10c5f991R203

I tried to do the same for rendezvous circuits in that commit but it didn't work as well. It still stands out as pretty unique.

I'm attaching a backgammon board for the commit above when used on the real network. See how intro circuits are blending in with general circuits very well, but rendezvous circuits still stand out. I think I might be able to tweak the state machine transitions and perhaps get to where we want...

Figure_1

src/core/or/circuitpadding_machines.c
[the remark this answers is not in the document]
asn-d6: ACK thanks for the clarification. Also added some comments in 7dc7c02f13.

src/core/or/circuitpadding.c
asn-d6: Agreed. Added comments in 7dc7c02f13.

src/core/or/circuitpadding_machines.c
asn-d6: Fixed in 7dc7c02f13.

src/core/or/circuitpadding_machines.c
asn-d6: Done in 7dc7c02f13.

src/core/or/circuitpadding_machines.c
asn-d6: Done in 7dc7c02f13.

asn-d6:
PR overseeded in favor of https://github.com/torproject/tor/pull/635

PR #869, opened 2019-03-29
Ticket29241 040

nmathewson:


teor2345:
Backported to 0.3.5 in #911.

PR #883, opened 2019-04-01
29979: Removed the use of expect from CodingStandardsRust

elichai:
Hi,
After watching your talk (https://www.youtube.com/watch?v=WI4ApeHH9QE) I came to read the documentation here, 
and you're telling to use `expect()` instead of `unwrap()` but expect also makes the whole program panic and unwind.
so neither expect should be used unless justified.

### DIFF
--- a/doc/HACKING/CodingStandardsRust.md
+++ b/doc/HACKING/CodingStandardsRust.md
@@ -256,7 +256,7 @@ Here are some additional bits of advice and rules:
    or 2) should fail (i.e. in a unittest).
 
    You SHOULD NOT use `unwrap()` anywhere in which it is possible to handle the
-   potential error with either `expect()` or the eel operator, `?`.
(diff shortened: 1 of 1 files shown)


doc/HACKING/CodingStandardsRust.md
```diff
@@ -256,20 +256,20 @@ Here are some additional bits of advice and rules:
    or 2) should fail (i.e. in a unittest).
 
    You SHOULD NOT use `unwrap()` anywhere in which it is possible to handle the
-   potential error with either `expect()` or the eel operator, `?`.
+   potential error the eel operator, `?` or another non panicing way.
```
teor2345: Two typos:
```suggestion
   potential error with the eel operator, `?` or another non panicking way.
```

teor2345:
Thanks! Someone else will squash and merge in the next few days.

nmathewson:
Squashed and merged!

PR #884, opened 2019-04-02
#29060: Stop requiring bash in test-network.sh. Make it POSIX compliant

rl1987:
https://trac.torproject.org/projects/tor/ticket/29060

### DIFF
--- a/src/test/test-network.sh
+++ b/src/test/test-network.sh
@@ -27,10 +27,6 @@ if echo "$@" | grep -e "--quiet" > /dev/null; then
   ECHO=true
 fi
 
-echo "CHUTNEY_PATH = $CHUTNEY_PATH"
-echo "TOR_DIR = $TOR_DIR"
-echo "ECHO = $ECHO"
-
 # optional: $TOR_DIR is the tor build directory
 # it's used to find the location of tor binaries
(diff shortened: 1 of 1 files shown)


src/test/test-network.sh
```diff
@@ -18,32 +18,14 @@ ECHO="${ECHO:-echo}"
...
-      ECHO=true
-    ;;
-    *)
-      # maybe chutney's test-network.sh can handle it
-    ;;
-  esac
-  shift
-done
+CHUTNEY_PATH=$(echo "$@" | awk -F '--chutney-path ' '{sub(" .*","",$2); print $2}')
+TOR_DIR=$(echo "$@" | awk -F '--tor-dir ' '{sub(" .*","",$2); print $2}')
```
teor2345: How do these awk commands handle quoted strings with spaces?

rl1987: They return the first thing that comes after e.g. `--chutney-path ` until the whitespace character, so it this not handle paths with spaces in them properly. We're not losing any functionality here, as the script didn't know how to handle them previously either - further in the script we're suggesting the user to set `CHUTNEY_PATH` and `TOR_DIR` environment variables as workaround.

PR #891, opened 2019-04-03
Bug28223 diagnostic 035

nmathewson:


src/feature/nodelist/microdesc.c
```diff
@@ -221,6 +221,13 @@ dump_microdescriptor(int fd, microdesc_t *md, size_t *annotation_len_out)
   }
 
   md->off = tor_fd_getpos(fd);
+  const char *nulpos = memchr(md->body, 0, md->bodylen);
+  if (BUG(nulpos)) {
+    log_warn(LD_BUG, "About to dump a NUL into a microdescriptor file. "
+             "offset %"PRId64", bodylen %zu, nul position %zu",
+             (int64_t)md->off, md->bodylen,
+             (size_t)(nulpos - md->body));
```
teor2345: Can we log the escaped() context of the NUL?
Maybe 10 bytes either side of the NUL?

teor2345: Also, can we log some identifier from the microdescriptor?

teor2345: Should we call warn_if_nul_found() here?

src/feature/nodelist/microdesc.c
```diff
@@ -221,6 +221,13 @@ dump_microdescriptor(int fd, microdesc_t *md, size_t *annotation_len_out)
   }
 
   md->off = tor_fd_getpos(fd);
+  const char *nulpos = memchr(md->body, 0, md->bodylen);
+  if (BUG(nulpos)) {
+    log_warn(LD_BUG, "About to dump a NUL into a microdescriptor file. "
+             "offset %"PRId64", bodylen %zu, nul position %zu",
```
teor2345: Windows says:
"error: unknown conversion type character 'z' in format"
https://ci.appveyor.com/project/torproject/tor/builds/23576412/job/80p2k2cumplatqx9#L1468

src/feature/nodelist/microdesc.c
```diff
@@ -480,6 +487,17 @@ microdesc_cache_clear(microdesc_cache_t *cache)
   cache->bytes_dropped = 0;
 }
 
+static void
+warn_if_nul_found(const char *inp, size_t len, const char *description)
+{
+  const char *nul_found = memchr(inp, 0, len);
+  if (BUG(nul_found)) {
+    log_warn(LD_BUG, "Found unexpected NUL while reading %s, at "
+             "position %zu/%zu.",
```
teor2345: Windows probably doesn't like this z, either.

src/feature/nodelist/microdesc.c
nmathewson: I'll try to do the other stuff; at this stage, there is no identifier from the md that we could log.

PR #908, opened 2019-04-05
Bug29960 cant fail

teor2345:


src/lib/crypt_ops/crypto_format.c
```diff
@@ -246,14 +246,11 @@ int
 ed25519_signature_from_base64(ed25519_signature_t *sig,
                               const char *input)
 {
-
   if (strlen(input) != ED25519_SIG_BASE64_LEN)
     return -1;
-  char buf[ED25519_SIG_BASE64_LEN+3];
+  char buf[ED25519_SIG_BASE64_LEN+1];
   memcpy(buf, input, ED25519_SIG_BASE64_LEN);
```
nmathewson: I think we don't need to use 'buf' here at all any more; we could just use "input" directly.

teor2345: 👍

PR #940, opened 2019-04-12
Control command refactor v2

nmathewson:


src/feature/control/control.c
```diff
@@ -276,25 +276,36 @@ peek_connection_has_http_command(connection_t *conn)
 }
 
 /**
- * Helper: take a nul-terminated command of given length, and find where
- * the command starts and the argument begins.  Separate them with a NUL,
- * and return a pointer to the arguments.
+ * Helper: take a nul-terminated command of given length, and find where the
+ * command starts and the argument begins.  Separate them; allocate a new
```
tlyu: s/argument begins/arguments begin/, probably?

src/feature/control/control.c
```diff
@@ -276,25 +276,36 @@ peek_connection_has_http_command(connection_t *conn)
...
-  *data_len -= (cmd_len+1); /* skip the command and NUL we added after it */
-  while (TOR_ISSPACE(*args)) {
-    ++args;
-    --*data_len;
+  *current_cmd_out = tor_memdup_nulterm(incoming_cmd, cmd_len);
+  char *args = incoming_cmd+cmd_len;
+  tor_assert(*data_len>=cmd_len);
+  *data_len -= cmd_len;
+  if (is_multiline) {
+    while ((*args == '\t' || *args == ' ') && *data_len) {
```
tlyu: If this is to avoid counting newlines as whitespace, maybe add a comment saying so?

src/feature/control/control.c
nmathewson: Fixup in cd994ceb6ce9b2.  Also fixes an "and and" later in the comment.

src/feature/control/control.c
nmathewson: e60bf6e066459d9 should solve this.

src/feature/control/control_cmd_args_st.h
```diff
@@ -0,0 +1,44 @@
+/* Copyright (c) 2001 Matej Pfajfar.
+ * Copyright (c) 2001-2004, Roger Dingledine.
+ * Copyright (c) 2004-2006, Roger Dingledine, Nick Mathewson.
+ * Copyright (c) 2007-2019, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
+ * \file control_cmd_args_st.h
+ * \brief Definition for contron_cmd_args_t
```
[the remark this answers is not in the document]
nmathewson: Fixup in 8885f07350bed7

src/feature/control/control_cmd.h
```diff
@@ -39,6 +47,37 @@ STATIC int add_onion_helper_keyarg(const char *arg, int discard_pk,
...
+  unsigned int min_args;
+  /**
+   * Highest number of positional arguments that this command accepts.
+   * UINT_MAX for no limit.
+   **/
+  unsigned int max_args;
+  /**
+   * True iff this command wants to be followed by a multiline object.
+   **/
+  bool want_object;
```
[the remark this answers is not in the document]
nmathewson: 621ec24f5bb31d should fix this

src/feature/control/control_cmd_args_st.h
```diff
@@ -0,0 +1,44 @@
...
+   **/
+  struct smartlist_t *args;
+  /**
+   * Number of bytes in <b>object</b>; 0 if <b>object</b> is not set.
+   **/
+  size_t object_len;
+  /**
+   * A multiline object passed with this command.
+   **/
+  char *object;
```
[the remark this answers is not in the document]
nmathewson: 621ec24f5bb31d should fix this too

src/feature/control/control.c
nmathewson: Fixup in cd994ceb6ce9b2.  Also fixes an "and and" later in the comment.

src/feature/control/control_cmd.c
```diff
@@ -120,18 +138,42 @@ control_cmd_parse_args(const char *command,
...
+      for (line = result->kwargs; line; line = line->next) {
+        if (! string_array_contains_keyword(syntax->allowed_keywords,
+                                            line->key)) {
+          tor_asprintf(error_out, "Unrecognized keyword argument %s",
+                       escaped(line->key));
+        }
+      }
+    }
+  }
+
```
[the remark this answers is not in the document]
nmathewson: f1da12cacb7e5a extracts the checking part of this function, where I expect us to grow more complexity in the future.

src/feature/control/control_cmd.c
```diff
@@ -2379,6 +2486,28 @@ handle_single_control_command(const control_cmd_def_t *def,
...
+        tor_free(err);
+      } else {
+        if (BUG(err))
+          tor_free(err);
+        if (def->handler.parsed(conn, parsed_args))
+          rv = 0;
+        control_cmd_args_free(parsed_args);
+      }
+      break;
+    }
```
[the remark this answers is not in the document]
nmathewson: This function will get cleaned up a lot later on in branch.  Please let me know if you think it's still ugly by the time the branch is over :)

src/feature/control/control_cmd.c
```diff
@@ -40,61 +40,256 @@
...
+string_array_contains_keyword(const char **array, const char *kwd)
+{
+  for (unsigned i = 0; array[i]; ++i) {
+    if (! strcasecmp(array[i], kwd))
+      return true;
+  }
+  return false;
+}
+
+/** Helper for argument parsing: check wither the keyword arguments just
```
tlyu: Typo: s/wither/whether/

src/feature/control/control_cmd.c
```diff
@@ -40,61 +40,256 @@
...
+kvline_check_keyword_args(const control_cmd_args_t *result,
+                          const control_cmd_syntax_t *syntax,
+                          char **error_out)
+{
+  if (result->kwargs == NULL) {
+    tor_asprintf(error_out, "Cannot parse keyword argument(s)");
+    return -1;
+  }
+
+  if (syntax->allowed_keywords) {
```
tlyu: This is a little better now that it's in a helper function. I still think this is a great place for an early return. (and it makes even more sense because there's no need for a `goto`.

src/feature/control/control_cmd.c
nmathewson: Fixup in b8ffede1e0ff0226f49f2a6a1e7fac2f535a82e0

src/feature/control/control_cmd.c
nmathewson: I think in this case I'd like to keep it as-is -- I anticipate that there are likely to be more kinds of check in the future, such as required keywords.

tlyu: I prefer to take the extra nesting out now, and add it back when we need it due to future complexity. It's not a strong preference in this case.

nmathewson:
Now squashed as #980

PR #951, opened 2019-04-15
Do not warn about compatible OpenSSL upgrades

bmwiedemann:
Do not warn about compatible OpenSSL upgrades.

When releasing OpenSSL patch-level maintenance updates,
we do not want to rebuild binaries using it.
And since they guarantee ABI stability, we do not have to.

Without this patch, warning messages were produced
that confused users:
https://bugzilla.opensuse.org/show_bug.cgi?id=1129411

### DIFF
--- a/changes/bug30190
+++ b/changes/bug30190
@@ -1,0 +1,3 @@
+  o Minor bugfixes (lib):
+    do not log a warning for OpenSSL versions that should be compatible
+    Fixes bug 30190; bugfix on 0.2.4.2-alpha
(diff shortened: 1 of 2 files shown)


teor2345:
Thanks for this pull request!

We do our task tracking using our own Trac instance, I've opened a ticket here:
https://trac.torproject.org/projects/tor/ticket/30190#comment:2

I think we've had issues in the past where OpenSSL has promised version compatibility, but failed to deliver. I've assigned the review to one of our developers who should be able to provide more details.

We try to do reviews within a week.

PR #956, opened 2019-04-15
Ticket30007 refactor control response formatting

tlyu:
overview:
* split low level control reply formatting out of control_fmt.c
* create abstractions to format the control protocol response syntax so that individual events or commands don't roll their own formatting
* use Coccinelle to transform existing explicit formatting into calls to the new functions
* some manual cleanup, both of whitespace quirks introduced by Coccinelle and some manual edits that it's not economical to write a Coccinelle script to accomplish

### DIFF
--- a/changes/ticket30091
+++ b/changes/ticket30091
@@ -1,0 +1,4 @@
+  o Major features (controller protocol):
+    - Controller commands are now parsed using a generalized parsing
+      subsystem. Previously, each controller command was responsible for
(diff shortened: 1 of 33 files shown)


tlyu:
This includes the commits of #940. The manual merge commit is because GitHub doesn't seem willing to make a trial merge commit automatically for this branch.

tlyu:
Superseded by #984, which is a rebased version.

PR #965, opened 2019-04-17
Bug28693+30173+29203

mikeperry-tor:


src/core/or/circuitpadding.c
```diff
@@ -1122,7 +1122,7 @@ circpad_machine_reached_padding_limit(circpad_machine_runtime_t *mi)
    * machine condition check happens, but machine checks only happen on
    * certain circuit events, and if padding is disabled due to some
    * network overload or DoS condition, we really want to stop ASAP. */
-  if (circpad_padding_disabled) {
+  if (circpad_padding_disabled || !get_options()->CircuitPadding) {
```
asn-d6: Hmm, this seems a bit like overfitting the original purpose of this function, since it was meant to be about rate limiting and not torrc/consensus parameters.

How about we make a new function `is_padding_allowed()` where we do these new checks, and also call `reached_padding_limit()` and also call `!is_participating_on_network()` (see schedule_padding()`). Then we can put it on top of `circpad_machine_schedule_padding()` and have a cleaner entry check.

mikeperry-tor: Code fixup: cda61784d0ab68ce6650cfdcc97b12a9fd84ee9b
Test fixup: 8a1fcbc631c54d992655447e25d8ddd7930f5d14

PR #966, opened 2019-04-18
Bug28780 squashed expiretest

mikeperry-tor:


src/core/or/circuitpadding.c
```diff
@@ -241,6 +241,20 @@ circpad_circuit_should_be_marked_for_close(circuit_t *circ, int reason)
       log_info(LD_GENERAL, "Circuit %d is not marked for close because of a "
                " pending padding machine.", CIRCUIT_IS_ORIGIN(circ) ?
                TO_ORIGIN_CIRCUIT(circ)->global_identifier : 0);
+
+      /* If the machine has had no network events at all within the
+       * last circpad_delay_t timespan, it's in some deadlock state.
+       * It's Ok to close now. */
+      if (circ->padding_info[i]->last_cell_time_sec +
+         (CIRCPAD_DELAY_INFINITE/CIRCPAD_DELAY_UNITS_PER_SECOND)+1
```
asn-d6: I think we can simplify this a bit. Instead of defining the constant number `CIRCPAD_DELAY_UNITS_PER_SECOND` and the constant expression `(CIRCPAD_DELAY_INFINITE/CIRCPAD_DELAY_UNITS_PER_SECOND)+1`, instead let's define a single constant number with the result of the above expression and document exactly what it means.

src/core/or/circuitpadding.c
```diff
@@ -253,13 +233,24 @@ circpad_circuit_should_be_marked_for_close(circuit_t *circ, int reason)
...
-        return 1;
+
+        return 0; // abort timer reached; mark the circuit for close now
       }
-      return 0;
+
+      /* If we weren't marked dirty yet, let's pretend we're dirty now, to use
+       * that timer instead of the idle one (we're not supposed to look idle,
+       * after all) */
+      if (!circ->timestamp_dirty)
```
asn-d6: I really like this commit. But what does this timestamp_dirty thing does here? I was not expecting to see it and I'm not sure what `timestamp_dirty` does in general, or why we need to mod it. Perhaps some more comments?

asn-d6: Ah I guess it's so that `circuit_expire_old_circuits_clientside()` definitely applies.

I think we should find a place and write a detailed analysis of how we think that this "block mark for close, but close in the expiry function" codepath should work exactly.

src/core/or/circuitpadding.c
mikeperry-tor: 769dec7004ed28e6c8c24477b3dd18c1032b1259

src/core/or/circuitpadding.c
mikeperry-tor: More comments: f6d815f7c54f27a3c15f9536430ccb2b1b449dd6

mikeperry-tor: More comments: f6d815f7c54f27a3c15f9536430ccb2b1b449dd6

mikeperry-tor:
Superceded by PR #1015

PR #969, opened 2019-04-19
Bug29034 0.3.5

asn-d6:


src/feature/hs/hs_circuit.c
```diff
@@ -1268,4 +1273,16 @@ hs_circ_cleanup(circuit_t *circ)
...
+  /* This circuit will be repurposed, clean the ident and other HS info. */
+  if (will_repurpose_circ) {
+    if (CIRCUIT_IS_ORIGIN(circ)) {
+      origin_circuit_t *ocirc = TO_ORIGIN_CIRCUIT(circ);
+      crypto_pk_free(ocirc->intro_key);
+      rend_data_free(ocirc->rend_data);
+
+      hs_ident_circuit_free(ocirc->hs_ident);
+      ocirc->hs_ident = NULL;
+    }
```
dgoulet-tor: My two cents here. I would much prefer we go with a `hs_circ_repurpose()` kind of function instead which would do this and call `hs_circ_cleanup()`.

One of the reason I say that also is because `crypto_pk_free()` and `rend_data_free()` is v2 and `hs_circ_cleanup()` is v3 only...

PR #970, opened 2019-04-19
[Tor Trac #22029]: Allow ed25519 keys to be banned in the approved-routers file (Revision 3)

neelchauhan:
For the bug [here](https://trac.torproject.org/projects/tor/ticket/22029).


doc/tor.1.txt
```diff
@@ -3430,11 +3430,11 @@ __DataDirectory__**/hashed-fingerprint**::
 
 __DataDirectory__**/approved-routers**::
     Only used by authoritative directory servers. This file lists
-    the status of routers by their identity fingerprint.
-    Each line lists a status and a fingerprint separated by
+    the status of routers by their identity fingerprint or base64-encoded
+    ed25519 public key. Each line lists a status and a fingerprint separated by
```
teor2345: "status and a fingerprint/pubkey"

doc/tor.1.txt
```diff
@@ -3430,11 +3430,11 @@ __DataDirectory__**/hashed-fingerprint**::
...
     whitespace. See your **fingerprint** file in the __DataDirectory__ for an
     example line. If the status is **!reject** then descriptors from the
```
teor2345: "example fingerprint"?

doc/tor.1.txt
```diff
@@ -3430,11 +3430,11 @@ __DataDirectory__**/hashed-fingerprint**::
 
 __DataDirectory__**/approved-routers**::
     Only used by authoritative directory servers. This file lists
-    the status of routers by their identity fingerprint.
-    Each line lists a status and a fingerprint separated by
+    the status of routers by their identity fingerprint or base64-encoded
+    ed25519 public key. Each line lists a status and a fingerprint separated by
     whitespace. See your **fingerprint** file in the __DataDirectory__ for an
```
teor2345: Where is the example file containing a "base64-encoded ed25519 public key"?

teor2345: See https://trac.torproject.org/projects/tor/ticket/30642

src/test/test_dir.c
```diff
@@ -6654,6 +6654,54 @@ test_dir_dirserv_router_get_status(void *arg)
...
+  RESET_FP_LIST(list);
+
+  /* Try a rejected router */
+  add_rsa_fingerprint_to_dir(fp, list, FP_REJECT);
+  tt_assert(dirserv_would_reject_router(&rs));
+  RESET_FP_LIST(list);
+
+  add_ed25519_to_dir(&kp.pubkey, list, FP_REJECT);
+  tt_assert(dirserv_would_reject_router(&rs));
+
```
teor2345: Please RESET_FP_LIST(list); after every test.

Otherwise, the next test could have bugs.

src/test/test_dir.c
```diff
@@ -6654,6 +6654,54 @@ test_dir_dirserv_router_get_status(void *arg)
   routerinfo_free(ri);
 }
 
+static void
+test_dir_dirserv_would_reject_router(void *arg)
```
teor2345: What happens when dirserv_would_reject_router() is called on an empty fingerprint list?
What should happen?

src/test/test_dir.c
```diff
@@ -6573,6 +6573,87 @@ test_dir_dirserv_load_fingerprint_file(void *arg)
   dirserv_free_fingerprint_list();
 }
 
+#define RESET_FP_LIST(list) STMT_BEGIN \
+    dirserv_free_fingerprint_list(); \
+    authdir_init_fingerprint_list(); \
+    list = authdir_return_fingerprint_list(); \
+  STMT_END
+
+#define FP_REJECT 4
```
teor2345: Please include the header that defines this constant.

src/test/test_dir.c
```diff
@@ -6573,6 +6573,87 @@ test_dir_dirserv_load_fingerprint_file(void *arg)
...
+  /* Try a rejected router */
+  add_rsa_fingerprint_to_dir(fp, list, FP_REJECT);
+  ret = dirserv_router_get_status(ri, &msg, LOG_INFO);
+  tt_int_op(ret, OP_EQ, FP_REJECT);
+  RESET_FP_LIST(list);
+
+  add_ed25519_to_dir(&kp1.pubkey, list, FP_REJECT);
+  ret = dirserv_router_get_status(ri, &msg, LOG_INFO);
+  tt_int_op(ret, OP_EQ, FP_REJECT);
+
```
teor2345: Please RESET_FP_LIST(list); after every test.

Otherwise, the next test could have bugs.

src/test/test_dir.c
```diff
@@ -6573,6 +6573,87 @@ test_dir_dirserv_load_fingerprint_file(void *arg)
...
+#define RESET_FP_LIST(list) STMT_BEGIN \
+    dirserv_free_fingerprint_list(); \
+    authdir_init_fingerprint_list(); \
+    list = authdir_return_fingerprint_list(); \
+  STMT_END
+
+#define FP_REJECT 4
+
+static void
+test_dir_dirserv_router_get_status(void *arg)
```
teor2345: What happens when dirserv_router_get_status() is called on an empty fingerprint list?
What should happen?

src/feature/dirauth/process_descs.c
```diff
@@ -221,50 +218,31 @@ dirserv_load_fingerprint_file(void)
...
-    }
-
-    /* Decode RSA keys */
-    char digest_tmp[DIGEST_LEN];
-    if (is_rsa &&
-        base16_decode(digest_tmp, sizeof(digest_tmp), fingerprint,
-                      HEX_DIGEST_LEN) != sizeof(digest_tmp)) {
-      is_valid_fpr = false;
+    if (strlen(fingerprint) == BASE64_DIGEST256_LEN &&
+        digest256_from_base64((char *) ed25519_pubkey_tmp.pubkey,
```
teor2345: This is a logic error: digest256_from_base64() returns 0 on success.

teor2345: Your tests didn't pick this up: please add a commit with a test that fails on this code.
And then write a fix to make the test pass.

src/feature/dirauth/process_descs.c
```diff
@@ -141,7 +139,6 @@ add_ed25519_to_dir(const ed25519_public_key_t *edkey, authdir_config_t *list,
   if (ed25519_validate_pubkey(edkey) < 0) {
     char ed25519_base64_key[ED25519_BASE64_LEN+1];
     ed25519_public_to_base64(ed25519_base64_key, edkey);
-    log_warn(LD_DIRSERV, "Invalid ed25519 key \"%s\"", ed25519_base64_key);
```
teor2345: Please don't delete this warning: you're hiding failures for every caller of this function.

Instead, do a minimal length check in dirserv_load_fingerprint_file().

src/feature/dirauth/process_descs.c
```diff
@@ -108,8 +108,6 @@ add_rsa_fingerprint_to_dir(const char *fp, authdir_config_t *list,
   tor_strstrip(fingerprint, " ");
   if (base16_decode(d, DIGEST_LEN,
                     fingerprint, strlen(fingerprint)) != DIGEST_LEN) {
-    log_warn(LD_DIRSERV, "Couldn't decode fingerprint \"%s\"",
-             escaped(fp));
```
teor2345: Please don't delete this warning: you're hiding failures for every caller of this function.

Instead, do a minimal length check in dirserv_load_fingerprint_file().

src/feature/dirauth/process_descs.c
```diff
@@ -221,50 +218,31 @@ dirserv_load_fingerprint_file(void)
...
-    bool is_valid_fpr = true;
+    /* Check if fingerprint is RSA or ed25519 by verifying it. */
+    int ed25519_not_ok, rsa_not_ok;
 
-    /* We only accept ed25519 or RSA keys */
-    if (!is_ed25519 && !is_rsa) {
-      is_valid_fpr = false;
+    /* Attempt to add the RSA key. */
+    rsa_not_ok = add_rsa_fingerprint_to_dir(fingerprint, fingerprint_list_new,
+                                            add_status);
```
teor2345: This code is unsafe: a malicious relay could have an ed25519 base64 encoded public key, that also passes the hexdigest check.

Please check the length of the string before passing it to this function.

src/feature/dirauth/process_descs.c
teor2345: This code would be easier to read, if the decoding was outside the "if" expression.

src/feature/dirauth/process_descs.c
```diff
@@ -307,9 +306,11 @@ dirserv_router_get_status(const routerinfo_t *router, const char **msg,
   }
 
   /* Check for the more usual versions to reject a router first. */
```
teor2345: This comment doesn't make sense.
Let's fix it while we are here?

"First, check for the more common reasons to reject a router"

src/feature/dirauth/process_descs.c
```diff
@@ -307,9 +306,11 @@ dirserv_router_get_status(const routerinfo_t *router, const char **msg,
...
-  uint32_t r = dirserv_get_status_impl(d, router->nickname, router->addr,
-                                       router->or_port, router->platform,
-                                       msg, severity);
+  uint32_t r = dirserv_get_status_impl(d,
+      router->cache_info.signing_key_cert->signing_key.pubkey,
```
teor2345: What if the router does not have an ed25519 key?
This code will crash.

teor2345: if (router->cache_info.signing_key_cert) {
    /* This has an ed25519 identity key. */

src/feature/dirauth/process_descs.c
```diff
@@ -381,28 +374,27 @@ dirserv_would_reject_router(const routerstatus_t *rs)
 {
   uint32_t res;
 
-  res = dirserv_get_status_impl(rs->identity_digest, rs->nickname,
-                                rs->addr, rs->or_port,
-                                NULL, NULL, LOG_DEBUG);
+  res = dirserv_get_status_impl(rs->identity_digest,
+                                rs->ed25519_signing_key.pubkey, rs->nickname,
```
teor2345: What if the router does not have an ed25519 key?
This code will crash.

src/feature/dirauth/process_descs.c
```diff
@@ -381,28 +374,27 @@ dirserv_would_reject_router(const routerstatus_t *rs)
...
+                                rs->addr, rs->or_port, NULL, NULL, LOG_DEBUG);
 
   return (res & FP_REJECT) != 0;
 }
 
 /** Helper: As dirserv_router_get_status, but takes the router fingerprint
- * (hex, no spaces), nickname, address (used for logging only), IP address, OR
- * port and platform (logging only) as arguments.
+ * (hex, no spaces), ed25510 key, nickname, address (used for logging only),
```
teor2345: Typo: ed25519

src/feature/dirauth/process_descs.c
```diff
@@ -440,17 +432,18 @@ dirserv_get_status_impl(const char *id_digest, const char *nickname,
...
+  if (ed25519_key) {
+    status_by_digest = digest256map_get(fingerprint_list->status_by_digest256,
+                                        ed25519_key);
+    if (status_by_digest)
+      result |= *status_by_digest;
+  }
+
   if (result & FP_REJECT) {
     if (msg)
       *msg = "Fingerprint is marked rejected -- if you think this is a "
```
teor2345: We need to update this message to say "mentioning your RSA fingerprint and ed25519 identity"

src/feature/dirauth/process_descs.c
```diff
@@ -47,10 +47,9 @@ static void directory_remove_invalid(void);
 static was_router_added_t dirserv_add_extrainfo(extrainfo_t *ei,
                                                 const char **msg);
 static uint32_t
-dirserv_get_status_impl(const char *fp, const char *nickname,
-                        uint32_t addr, uint16_t or_port,
-                        const char *platform, const char **msg,
-                        int severity);
+dirserv_get_status_impl(const char *fp, const uint8_t *ed25519_key,
```
teor2345: The new argument should be "const ed25519_public_key_t *ed25519_public_key".

src/feature/nodelist/routerstatus_st.h
```diff
@@ -20,6 +21,8 @@ struct routerstatus_t {
   /** Digest of the router's most recent descriptor or microdescriptor.
    * If it's a descriptor, we only use the first DIGEST_LEN bytes. */
   char descriptor_digest[DIGEST256_LEN];
+  ed25519_public_key_t ed25519_signing_key; /**< The ed25519 signing key of
+                                             * this router. */
```
teor2345: This needs to be a pointer: "ed25519_public_key_t *ed25519_signing_key".
Some routers do not have an ed25519 key.

src/feature/dirauth/voteflags.c
```diff
@@ -608,6 +609,11 @@ set_routerstatus_from_routerinfo(routerstatus_t *rs,
   rs->dir_port = ri->dir_port;
   rs->is_v2_dir = ri->supports_tunnelled_dir_requests;
 
+  if (ri->cache_info.signing_key_cert) {
+    ed25519_pubkey_copy(&rs->ed25519_signing_key,
```
teor2345: This code leaves the key zeroed-out if the relay does not have an ed25529 key.
Instead, allocate memory for the key if it is non-NULL.
And free that memory in routerstatus_free_().

src/test/test_dir.c
teor2345: Hmm, there is no header that defines this constant.
Can you call a function that checks if the relay is rejected?

neelchauhan: I don't think this is possible without adding a lot more code because the function `dirserv_would_reject_router()` expects a `routerstatus_t` and I have a `routerinfo_t` making this impractical.

src/test/test_dir.c
neelchauhan: I have added this test. `dirserv_router_get_status()` returns 0 in this case.

src/test/test_dir.c
neelchauhan: I've added this.

src/test/test_dir.c
neelchauhan: I fixed it. Both the values return 0 for RSA and ed25519.

src/test/test_dir.c
neelchauhan: I did this.

src/test/test_dir.c
teor2345: I opened another ticket to make this change:
https://trac.torproject.org/projects/tor/ticket/30677

doc/tor.1.txt
neelchauhan: f464f6dc13b6eba171015e0a775457900885affd

doc/tor.1.txt
neelchauhan: 89f4a2500a62dbbbb58205b32e7d32c1b154e37e

src/test/test_dir.c
neelchauhan: c2d60099fd1a7291eb05292ae45665632fa2abf0

src/test/test_dir.c
neelchauhan: e312885f7079ad681bbfafe5d220aeade597c8a8

src/test/test_dir.c
neelchauhan: 4631fbd7f004c6b544e08095aa14ce77f60c8c63

src/feature/dirauth/process_descs.c
neelchauhan: 9520e28510e757b54c52b8b2177b63936d16de58

However, I didn't move the decoding outside the "if" expression.

src/test/test_dir.c
neelchauhan: 38a1cd8e1e5f1d9918bd17d1e2a79dcc21d96d90

src/feature/dirauth/process_descs.c
neelchauhan: b6465c0aec08febe5d40dd343838095cb96faf1d

src/feature/dirauth/process_descs.c
neelchauhan: b6465c0aec08febe5d40dd343838095cb96faf1d

src/feature/dirauth/process_descs.c
neelchauhan: b6465c0aec08febe5d40dd343838095cb96faf1d

src/feature/dirauth/process_descs.c
neelchauhan: b6465c0aec08febe5d40dd343838095cb96faf1d

src/feature/dirauth/process_descs.c
neelchauhan: 0d36d1f514e308d741f2ba232276115458aa86f8

src/feature/dirauth/process_descs.c
neelchauhan: b6465c0aec08febe5d40dd343838095cb96faf1d

src/feature/dirauth/process_descs.c
neelchauhan: 46d7d140e2f4927059f82525e265a7158e6314b5

src/feature/nodelist/routerstatus_st.h
neelchauhan: 0d36d1f514e308d741f2ba232276115458aa86f8

src/feature/dirauth/voteflags.c
neelchauhan: 0d36d1f514e308d741f2ba232276115458aa86f8

src/feature/dirauth/process_descs.c
teor2345: I opened a new ticket for the decoding refactor:
https://trac.torproject.org/projects/tor/ticket/30691

And the missing unit test cases:
https://trac.torproject.org/projects/tor/ticket/30692#ticket

src/feature/dirauth/process_descs.c
```diff
@@ -241,10 +304,18 @@ dirserv_router_get_status(const routerinfo_t *router, const char **msg,
...
   }
 
-  /* Check for the more common reasons to reject a router first. */
-  const uint32_t r = dirserv_get_status_impl(d, router->nickname,
-                                             router->addr, router->or_port,
-                                             router->platform, msg, severity);
+  /* First, check for the more common reasons to reject a router. */
+  if (router->cache_info.signing_key_cert) {
+    /* This has an ed25519 identity key. */
+    r = dirserv_get_status_impl(d,
```
nmathewson: These two calls to dirserv_get_status_impl are almost identical.  Instead of having two calls like this, it's cleaner to have a local variable that holds either the ed25519 public key or NULL, and then pass that variable to a single call.

src/feature/nodelist/routerstatus_st.h
```diff
@@ -20,6 +21,8 @@ struct routerstatus_t {
   /** Digest of the router's most recent descriptor or microdescriptor.
    * If it's a descriptor, we only use the first DIGEST_LEN bytes. */
   char descriptor_digest[DIGEST256_LEN];
+  ed25519_public_key_t *ed25519_signing_key; /**< The ed25519 signing key of
```
nmathewson: Can we possibly avoid adding this field? It's stored in microdescriptors right now, not routerstatuses at all.

The reason we shouldn't add it here is that we can't ensure that this field is actually set in every routerstatus_t for a relay with an ed25519 key.  We set it in dirauth_set_routerstatus_from_routerinfo, but note that we can't set it at all when we receive the routerstatus_t from a vote.

PR #981, opened 2019-04-25
Refactor mainloop.c so that periodic events can be moved out.

nmathewson:
(Also, do this for dirauth events.)

### DIFF
--- a/src/core/mainloop/mainloop.c
+++ b/src/core/mainloop/mainloop.c
@@ -1587,6 +1587,13 @@ STATIC void
 teardown_periodic_events(void)
 {
   periodic_events_destroy_all();
+  check_descriptor_event = NULL;
+  dirvote_event = NULL;
+  fetch_networkstatus_event = NULL;
+  launch_descriptor_fetches_event = NULL;
+  check_dns_honesty_event = NULL;
+  save_state_event = NULL;
+  prune_old_routers_event = NULL;
   periodic_events_initialized = 0;
 }
 
@@ -1621,13 +1628,6 @@ rescan_periodic_events(const or_options_t *options)
 {
   tor_assert(options);
 
-  /* Avoid scanning the event list if we haven't initialized it yet. This is
-   * particularly useful for unit tests in order to avoid initializing main
-   * loop events everytime. */
-  if (!periodic_events_initialized) {
-    return;
-  }
-
   periodic_events_rescan_by_roles(get_my_roles(options), net_is_disabled());
 }
 
@@ -1636,13 +1636,7 @@ rescan_periodic_events(const or_options_t *options)
 void
 periodic_events_on_new_options(const or_options_t *options)
 {
-  /* Only if we've already initialized the events, rescan the list which will
-   * enable or disable events depending on our roles. This will be called at
-   * bootup and we don't want this function to initialize the events because
-   * they aren't set up at this stage. */
-  if (periodic_events_initialized) {
-    rescan_periodic_events(options);
-  }
+  rescan_periodic_events(options);
 }
 
 /**
@@ -2107,7 +2101,7 @@ dirvote_callback(time_t now, const or_options_t *options)
 void
 reschedule_dirvote(const or_options_t *options)
 {
-  if (periodic_events_initialized && authdir_mode_v3(options)) {
+  if (authdir_mode_v3(options) && dirvote_event) {
     periodic_event_reschedule(dirvote_event);
   }
 }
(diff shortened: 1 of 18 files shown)


dgoulet-tor:
Commit 0d650e7958d011dc message has a typo with: `mainlook.c`

src/core/mainloop/mainloop.c
```diff
@@ -2778,6 +2771,13 @@ do_main_loop(void)
    */
   initialize_periodic_events();
```
dgoulet-tor: So this is the only thing I'm confused and I would like to have documented why or resolved if we can.

These calls seems to belong to the subsystem initialization function where `initialize_periodic_events()` is actually invoked. If we can't use the initialize subsystem function for this, it kind of looses its purpose a bit to have two callsite that initialize. Seems very error prone to me over time :S.

src/core/mainloop/periodic.c
```diff
@@ -24,6 +32,11 @@
  */
 static const int MAX_INTERVAL = 10 * 365 * 86400;
 
+/**
+ *
+ **/
```
dgoulet-tor: Some documentation got truncated? :)

src/core/mainloop/periodic.c
```diff
@@ -184,3 +197,123 @@ periodic_event_schedule_and_disable(periodic_event_item_t *event)
...
+ * Does not free the periodic_event_item_t object themselves, because we do
+ * not own them. */
+void
+periodic_events_destroy_all(void)
+{
+  if (! the_periodic_events)
+    return;
+
+  SMARTLIST_FOREACH_BEGIN(the_periodic_events, periodic_event_item_t *, item) {
+    periodic_event_destroy(item);
```
dgoulet-tor: Shouldn't the module be in charge of doing this? A subsystem adds the event it but doesn't destroy?

src/core/mainloop/mainloop.c
```diff
@@ -1575,14 +1558,15 @@ initialize_periodic_events(void)
...
-  /* Set up all periodic events. We'll launch them by roles. */
-  int i;
-  for (i = 0; periodic_events[i].name; ++i) {
-    periodic_event_setup(&periodic_events[i]);
+  for (int i = 0; mainloop_periodic_events[i].name; ++i) {
+    periodic_events_add(&mainloop_periodic_events[i]);
   }
 
+  /* Set up all periodic events. We'll launch them by roles. */
+  periodic_events_setup_all();
```
dgoulet-tor: Should this be done by the subsystem that is setup its own event?

src/core/mainloop/periodic.c
nmathewson: Thanks; I've added a commit to fix this.

src/core/mainloop/mainloop.c
nmathewson: I've removed these calls, and nothing broke. :)

src/core/mainloop/periodic.c
nmathewson: I've added a documentation commit that renames a bunch of methods and explains what's going on; I hope it will make the logic here look more reasonable.

src/core/mainloop/mainloop.c
nmathewson: See note above about documentation commit; I think that will address this issue too

src/feature/dirauth/dirauth_periodic.c
```diff
@@ -0,0 +1,142 @@
...
+   * stability info counts more, and save the stability information to disk as
+   * appropriate. */
+  time_t next = rep_hist_downrate_old_runs(now);
+  return safe_timer_diff(now, next);
+}
+
+DECLARE_EVENT(downrate_stability, AUTHORITIES, 0);
+
+void
+dirauth_add_periodic_events(void)
```
dgoulet-tor: We probably want to rename it to `register` here after the rename commit.

src/core/mainloop/periodic.c
```diff
@@ -307,12 +317,14 @@ periodic_events_rescan_by_roles(int roles, bool net_disabled)
...
+/**
+ * Invoked at shutdown: disconnect and unregister all periodic events.
  *
  * Does not free the periodic_event_item_t object themselves, because we do
- * not own them. */
+ * not own them.
+ */
 void
-periodic_events_destroy_all(void)
+periodic_events_disconnect_all(void)
```
dgoulet-tor: I do see the `periodic_event_destroy()` still a thing which is ONLY called by this `disconnect_all`. This is a bit confusing tbh since the function actually frees the event so "destroy" is valid. Since the `connect()` actually allocates the event, I think then we could safely rename the `_destroy()` to be `_disconnect()`.

And while at it, we might want to make it static since it is not used outside the module. Side note on that, all the specific `periodic_events_` seems only used within the module so if we want to narrow down the exposed API of `periodic.c`, we could static a bunch of those :).

nmathewson:
Squashed as #993

PR #984, opened 2019-04-26
Ticket30007 v2 refactor control response formatting

tlyu:
This is a rebased version of #956, rebased on #980. I verified that the Coccinelle changes are equivalent by diffing the output of the Coccinelle run against the rebased committed previous Coccinelle output.


tlyu:
Rebased to master. Re-verified Coccinelle output.

PR #997, opened 2019-04-30
Bug28693+30173+29203 rebased

mikeperry-tor:


src/core/or/circuitpadding.c
```diff
@@ -1104,6 +1104,24 @@ circpad_new_consensus_params(const networkstatus_t *ns)
          CIRCWINDOW_START_MAX, 0, 50*CIRCWINDOW_START_MAX);
 }
 
+/**
+ * Return true if padding is allowed by torrc and consensus.
+ */
+STATIC bool
+circpad_is_padding_allowed(void)
```
asn-d6: No need to have `circpad` as a function name prefix since it's a private function to this file.

src/core/or/circuitpadding.c
```diff
@@ -1104,6 +1104,24 @@ circpad_new_consensus_params(const networkstatus_t *ns)
...
+{
+  /* If padding has been disabled in the consensus, don't send any more
+   * padding. Technically the machine should be shut down when the next
+   * machine condition check happens, but machine checks only happen on
+   * certain circuit events, and if padding is disabled due to some
+   * network overload or DoS condition, we really want to stop ASAP. */
+  if (circpad_padding_disabled || !get_options()->CircuitPadding) {
+    return 0;
+  }
+
```
asn-d6: We can also pull in
```
  /* Don't schedule padding if we are currently in dormant mode. */
  if (!is_participating_on_network()) {
    log_info(LD_CIRC, "Not scheduling padding because we are dormant.");
    return CIRCPAD_STATE_UNCHANGED;
  }
```
from `circpad_machine_schedule_padding` in this function.

src/core/or/circuitpadding.c
mikeperry-tor: a151032662efdb2421187d83dade96f5a1634499 does this (and fixes the CI error)

src/core/or/circuitpadding.c
mikeperry-tor: I would like to keep this separate as it requires its own log line and is a different mechanism that causes it to happen.

PR #1001, opened 2019-05-06
#30361: address CID 1444908 and CID 1444769

rl1987:
https://trac.torproject.org/projects/tor/ticket/30361

### DIFF
--- a/src/feature/nodelist/microdesc.c
+++ b/src/feature/nodelist/microdesc.c
@@ -512,9 +512,9 @@ warn_if_nul_found(const char *inp, size_t len, int64_t offset,
 int
 microdesc_cache_reload(microdesc_cache_t *cache)
 {
-  struct stat st;
-  char *journal_content;
-  smartlist_t *added;
+  struct stat st = {0};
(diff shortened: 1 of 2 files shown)


src/feature/nodelist/microdesc.c
```diff
@@ -535,19 +537,27 @@ microdesc_cache_reload(microdesc_cache_t *cache)
...
-  if (journal_content) {
+  if (journal_content && st.st_size) {
     cache->journal_len = (size_t) st.st_size;
     warn_if_nul_found(journal_content, cache->journal_len, 0,
                       "reading microdesc journal");
+
+    /* Replace all NUL bytes in journal_content with space character. */
+    for (size_t i = 0; i < (size_t)st.st_size; i++)
+      if (journal_content[i] == '\0')
+        journal_content[i] = ' ';
```
ahf: Hm. I'm not sure I think this is the right solution to the problem. Ideally there should never be any NUL bytes around here, so can't we instead pretend that the file is at EOF when we have the first NUL?

PR #1002, opened 2019-05-06
More subsystems

nmathewson:


src/feature/relay/relay_sys.c
```diff
@@ -0,0 +1,46 @@
...
+#include "feature/relay/onion_queue.h"
+#include "feature/relay/relay_sys.h"
+#include "feature/relay/routerkeys.h"
+#include "feature/relay/router.h"
+
+#include "lib/subsys/subsys.h"
+
+static int
+subsys_relay_initialize(void)
+{
```
dgoulet-tor: Where are all the init functions? I know `dns_init()` is one of them. Maybe also `init_keys()` ?

nmathewson: dns_init and init_keys can't actually go here, since they require configuration.  Once we refactor the configuration system to work with our subsystems logic, that's when we'll want to move those.

PR #1007, opened 2019-05-08
Bug29085

mikeperry-tor:


mikeperry-tor:
Closed in favor of https://github.com/torproject/tor/pull/1014

src/core/or/circuitpadding.c
```diff
@@ -851,12 +896,59 @@ circpad_machine_remove_token(circpad_machine_runtime_t *mi)
...
+  /* Remove a token from the histrogram, if applicable */
+  circpad_machine_remove_token(mi);
+}
+
+static inline void
+circpad_machine_update_state_length(circpad_machine_runtime_t *mi)
+{
+  const circpad_state_t *state = NULL;
+
+  if (mi->state_length != CIRCPAD_STATE_LENGTH_INFINITE)
```
nmathewson: Is this test backwards?  In other places, INFINITE means "don't decrement it."  But here we decrement it only if it is infinite.

Should we be worried that the tests didn't catch this?

src/core/or/circuitpadding.c
```diff
@@ -851,12 +896,59 @@ circpad_machine_remove_token(circpad_machine_runtime_t *mi)
...
   }
 
+  /* Update any state packet length limits that apply */
+  circpad_machine_update_state_length(mi);
+
+  /* Remove a token from the histrogram, if applicable */
+  circpad_machine_remove_token(mi);
+}
+
+static inline void
```
nmathewson: Let's document this.

src/test/test_circuitpadding.c
```diff
@@ -584,12 +584,12 @@ test_circuitpadding_token_removal_higher(void *arg)
     tt_int_op(mi->histogram[bin_to_remove], OP_EQ, 2);
 
     mi->padding_scheduled_at_usec = current_time - 57;
-    circpad_machine_remove_token(mi);
+    circpad_cell_event_nonpadding_sent((circuit_t*)client_side);
```
nmathewson: It's better to use TO_CIRCUIT() for this kind of upcast: it's typesafe, and more compiler-friendly.

src/core/or/circuitpadding.c
```diff
@@ -442,7 +442,7 @@ circpad_machine_sample_delay(circpad_machine_runtime_t *mi)
     return circpad_distribution_sample_iat_delay(state, iat_delay_shift);
   } else if (state->token_removal != CIRCPAD_TOKEN_REMOVAL_NONE) {
     /* We have a mutable histogram. Do basic sanity check and apply: */
-    if (BUG(!mi->histogram) ||
+    if (BUG(!CIRCPAD_IS_TOKEN_REMOVAL_SUPPORTED(mi)) ||
```
nmathewson: This is not exactly the same as it was before -- histogram_len was not previously checked.  I'm assuming that's okay.

src/core/or/circuitpadding.h
```diff
@@ -92,6 +92,15 @@ typedef uint32_t circpad_delay_t;
  */
 #define CIRCPAD_INFINITY_BIN(mi)  ((mi)->histogram_len-1)
 
+/**
+ * Macro to check if we have mutable histograms in circpad_machine_runtime_t.
+ *
+ * Having hutable histograms in the circpad_machine_runtime_t is equivalent to
```
nmathewson: s/hutable/mutable/

src/core/or/circuitpadding.h
```diff
@@ -92,6 +92,15 @@ typedef uint32_t circpad_delay_t;
...
+ * circpad_state_t.token_removal != CIRCPAD_TOKEN_REMOVAL_NONE.
+ */
+#define CIRCPAD_IS_TOKEN_REMOVAL_SUPPORTED(mi) \
```
nmathewson: Since this macro uses mi twice, it is not perfectly safe. I'd suggest a static inline function instead.

src/core/or/circuitpadding.c
```diff
@@ -1482,7 +1482,12 @@ circpad_estimate_circ_rtt_on_received(circuit_t *circ,
            ", %d) after two back to back packets. Current RTT: %d",
            circ->n_chan ?  circ->n_chan->global_identifier : 0,
            circ->n_circ_id, mi->rtt_estimate_usec);
-       mi->stop_rtt_update = 1;
+      mi->stop_rtt_update = 1;
+
+      if (!mi->rtt_estimate_usec) {
+        log_fn(LOG_NOTICE, LD_CIRC,
```
nmathewson: should this be warn/bug ?  Should it have a ratelimit?

src/core/or/circuitpadding.c
mikeperry-tor: This was done in the new PR, #1004 (see 8cfd329).

src/core/or/circuitpadding.c
mikeperry-tor: Yes, this was why I introduced this macro, to ensure that we have equivalence between histogram == NULL, and histrogram_length == 0, and also state_t.token_removal == CIRCPAD_TOKEN_REMOVAL_NONE. All those should be set at the same time.

When I transform this into a static inline as per your other comment, I will make that function check all of the things. (The state check is more expensive since we don't always have a direct pointer into the state, and it requires several dereferences to get one).

src/core/or/circuitpadding.c
mikeperry-tor: Holy moly. Yes and a little. Our tests never tested state lengths that also wanted to count non-padding cells as part of the length of the state. Dag.

I will write a test that fails for this case and then fix this.

mikeperry-tor: Nice find! Here is the new test: de62ccd567fb0536bef5c66f8704f4c44086fbde

Here are two typo bugs it found (yours and one other): 0a10fc6208d7e93f1cc8cac0299ee7e78dd37b3c

And here is a token empty bug it found if a padding packet causes us to run out of tokens: ae719d348e1253ad126a0e57a85bdb7c85a11c70

src/test/test_circuitpadding.c
mikeperry-tor: Heh, not only that, these were all circuit_t * already. I fixed all the instances of this in the circpad tests here: fe6fc3cff894c13120a2fdc507aa09f6a9cc5bb2 (PR #1024)

src/core/or/circuitpadding.h
mikeperry-tor: Ok this is much better all around: 28a081dacca387d160ae02eda3a4176610e0ff12 (in PR #1024).

Thanks.

src/core/or/circuitpadding.c
mikeperry-tor: Yeah probably. And on the other side too: 4f07010263ef8841109279b82bd355c4d9051381

PR #1009, opened 2019-05-09
Add a --list-modules command

nmathewson:
Closes ticket 30452.


src/app/config/config.c
```diff
@@ -5198,6 +5199,17 @@ options_init_from_torrc(int argc, char **argv)
     return 1;
   }
 
+  if (config_line_find(cmdline_only_options, "--list-modules")) {
+    puts("dirauth: "
+#ifdef HAVE_MODULE_DIRAUTH
```
tlyu: I would somewhat rather duplicate the function call inside the `#ifdef` branches instead of splitting the function call across `#ifdef` branches. It's a a short string. I think duplication is slightly easier to read and ends up adding fewer lines.

nmathewson:
Closed in favor of #1020.

PR #1010, opened 2019-05-09
Ticket30452 041

nmathewson:


src/app/config/config.c
```diff
@@ -5185,6 +5186,17 @@ options_init_from_torrc(int argc, char **argv)
     return 1;
   }
 
+  if (config_line_find(cmdline_only_options, "--list-modules")) {
+    puts("dirauth: "
+#ifdef HAVE_MODULE_DIRAUTH
+         "yes"
+#else
+         "no"
+#endif
```
dgoulet-tor: I would personally really rather we avoid these #ifdef/#else in the code like this but rather add a function within the dirauth module that returns `true` if enabled or `false` if not compiled in.

Centralizing these ifdefs *only* within the module makes it much easier and less error prone to track what is dirauth specific or not to the module.

Maybe something that `dirauth_sys.h` could have a `dirauth_sys_enabled()` or sth like that?

nmathewson:
Closed in favor of #1021

PR #1012, opened 2019-05-09
Ticket30454 034 01

dgoulet-tor:


src/trunnel/hs/cell_introduce1.trunnel
```diff
@@ -32,7 +37,9 @@ struct trn_cell_introduce1 {
 /* INTRODUCE_ACK payload. See details in section 3.2.2. */
 struct trn_cell_introduce_ack {
   /* Status of introduction. */
-  u16 status IN [0x0000, 0x0001, 0x0002];
+  u16 status IN [TRUNNEL_HS_INTRO_ACK_STATUS_SUCCESS,
```
nmathewson: I question this IN.  Do we really want an unrecognized error code to be a parse failure?

dgoulet-tor: Indeed, I'm removing that.

PR #1015, opened 2019-05-11
Bug28780 squashed2

mikeperry-tor:


src/core/or/circuitlist.h
```diff
@@ -92,41 +92,43 @@
...
  *
  *  This type of circuit is *only* created preemptively and never
  *  on-demand. When an HS operation needs to take place (e.g. connect to an
  *  intro point), these circuits are then cannibalized and repurposed to the
  *  actual needed HS purpose. */
-#define CIRCUIT_PURPOSE_HS_VANGUARDS 23
+#define CIRCUIT_PURPOSE_HS_VANGUARDS 24
 
-#define CIRCUIT_PURPOSE_MAX_ 23
+#define CIRCUIT_PURPOSE_MAX_ 25
```
nmathewson: Shouldn't this be 24 ?

src/core/or/circuitpadding.c
```diff
@@ -161,6 +161,113 @@ circpad_circuit_machineinfo_free_idx(circuit_t *circ, int idx)
...
+ * network-driven cell events.
+ *
+ * After CIRCPAD_DELAY_INFINITE microseconds of silence on a circuit, this
+ * function will no longer hold it open (it will return 0 regardless of
+ * what the machines ask for, and thus circuit_expire_old_circuits_clientside()
+ * will close the circuit after roughly 1.25hr of idle time, maximum,
+ * regardless of the padding machine state.
+ */
+int
+circpad_is_using_circuit_for_padding(circuit_t *circ, int reason)
```
nmathewson: It's still wrong for a function with a name like this to make changes on a circuit.  It should really just return 1 or 0, and let the caller invoke another function that adjusts the dirty timestamp and the purpose.

nmathewson: (If it's going to have the current functionality, it really needs a different name.)

src/core/or/circuitpadding.c
```diff
@@ -161,6 +161,113 @@ circpad_circuit_machineinfo_free_idx(circuit_t *circ, int idx)
...
+      continue; // check next machine
+    } else {
+      log_info(LD_CIRC, "Circuit %d is not marked for close because of a "
+               " pending padding machine.", CIRCUIT_IS_ORIGIN(circ) ?
+               TO_ORIGIN_CIRCUIT(circ)->global_identifier : 0);
+
+      /* If the machine has had no network events at all within the
+       * last circpad_delay_t timespan, it's in some deadlock state.
+       * Tell the caller that we don't own it anymore. */
+      if (circ->padding_info[i]->last_cell_time_sec +
```
nmathewson: I like that there is some mechanism of deadlock prevention here, but whom do we expect to call this function in this case?

src/core/or/circuitpadding.c
```diff
@@ -161,6 +161,113 @@ circpad_circuit_machineinfo_free_idx(circuit_t *circ, int idx)
...
+
+    /* If the machine does not want to control the circuit close itself, then
+     * check the next machine */
+    if (!circ->padding_machine[i]->manage_circ_lifetime) {
+      continue; // check next machine
+    }
+
+    /* If the machine has reached the END state, we can close. Check next
+     * machine. */
+    if (mi->current_state == CIRCPAD_STATE_END) {
```
asn-d6: While we are doing fixups here, there is no  reason to have an `else` statement here IIUC. We can just have the `current_state` check here be yet another `continue` check, and continue without nesting.

mikeperry-tor:
Due to merge conflicts, this PR is closed. See #1023. I will still reply to the review comments here, though.

src/core/or/circuitlist.h
mikeperry-tor: Yes. Fixed in d9c864b7294273a5dd41e25fae1bd35571ebf52e (in PR #1023).

src/core/or/circuitpadding.c
mikeperry-tor: How about circpad_marked_circuit_for_padding()? d9c864b7294273a5dd41e25fae1bd35571ebf52e

src/core/or/circuitpadding.c
mikeperry-tor: Comment clarified in 98d995a68e054dd0a9d36fcbc368bd6c45e22d07 in PR #1023.

src/core/or/circuitpadding.c
mikeperry-tor: 17765cf2c9e85d8ac0c6b8585486dd7af51c9779 in PR #1023

PR #1026, opened 2019-05-15
Ticket30428 041 02

dgoulet-tor:


src/core/or/sendme.c
```diff
@@ -571,31 +594,25 @@ sendme_note_stream_data_packaged(edge_connection_t *conn)
...
-    return;
-  }
-  /* Is this the last cell before a SENDME? The idea is that if the
-   * package_window reaches a multiple of the increment, after this cell, we
-   * should expect a SENDME. */
-  if (!sendme_circuit_cell_is_next(circ->package_window)) {
-    return;
+  if (cpath) {
+    package_window = cpath->package_window;
+    sendme_digest = cpath_get_sendme_digest(cpath);
```
nmathewson: This looks like we're moving backwards here: like we are computing the sendme digest even when we are not going to record it. Is that correct?  Remember, these digests are not cheap to calculate.

(edited to add) -- but it looks like we add other conditions later?  I'm a bit confused, I'll have to come back to this branch.

src/core/crypto/relay_crypto.c
```diff
@@ -100,12 +100,20 @@ relay_crypto_get_sendme_digest(relay_crypto_t *crypto)
...
-relay_crypto_record_sendme_digest(relay_crypto_t *crypto)
+relay_crypto_record_sendme_digest(relay_crypto_t *crypto,
+                                  bool is_foward_digest)
 {
+  struct crypto_digest_t *digest;
+
   tor_assert(crypto);
-  crypto_digest_get_digest(crypto->b_digest, (char *) crypto->sendme_digest,
+
+  (is_foward_digest) ? (digest = crypto->f_digest) :
```
nmathewson: Please just use "if" for this kind of conditional.  Or use `digest = is_forward_digest ? __ : __;`

src/core/or/sendme.c
```diff
@@ -615,3 +607,55 @@ sendme_record_cell_digest_on_circ(circuit_t *circ, crypt_path_t *cpath)
...
+ * is expected to be a SENDME. */
+void
+sendme_record_received_cell_digest(circuit_t *circ, crypt_path_t *cpath)
+{
+  tor_assert(circ);
+
+  /* Only record if the next cell is expected to be a SENDME. */
+  if (!sendme_circuit_cell_is_next(cpath ? cpath->deliver_window :
+                                           circ->deliver_window)) {
+    goto end;
```
nmathewson: No reason not to just return here.

src/core/crypto/relay_crypto.c
dgoulet-tor: Fixup commit: 5298b96783]

src/core/or/sendme.c
dgoulet-tor: Indeed... Very good catch.

I've fix this in the fixup commit f78fb4349da19daf and added a comment to this part about the CPU cost.

PR #1027, opened 2019-05-15
sendme: Introduce consensus param sendme_dir_min_version

dgoulet-tor:
See section 4.5 of proposal 289 which describes why we need this parameter. To
summarize, this is to accomodate old clients not supporting authenticated
SENDMEs and booting up for the first time.

They need to be able to fetch a consensus to actually learn that the required
protocols is FlowCtrl=1 and thus not start up. And for that, we need to allow
a SENDME minimum version on a directory request circuit. That new consensus
parameter handles that.

Closes #30364

### DIFF
--- a/src/core/or/or_circuit_st.h
+++ b/src/core/or/or_circuit_st.h
@@ -61,6 +61,13 @@ struct or_circuit_t {
    *  statistics. */
   unsigned int circuit_carries_hs_traffic_stats : 1;
 
+  /** If set, this circuit got a BEGIN_DIR, meaning a directory request, and
+   * not for any hidden service documents. We use this flag for SENDME
+   * mechanism in order to know if this circuit will serve a consensus for
+   * which we need to obey a specific consensus parameter about the minimal
+   * SENDME version we accept. */
+  unsigned int begin_dir_seen : 1;
+
   /** Number of cells that were removed from circuit queue; reset every
    * time when writing buffer stats to disk. */
   uint32_t processed_cells;
--- a/src/core/or/relay.c
+++ b/src/core/or/relay.c
@@ -1670,6 +1670,11 @@ connection_edge_process_relay_cell(cell_t *cell, circuit_t *circ,
                "from intermediary node. Dropping.");
         return 0;
       }
+      if (circ->purpose == CIRCUIT_PURPOSE_OR) {
+        /* Flag that this circuit just got a BEGIN_DIR and it is not an hidden
+         * service document request. */
+        TO_OR_CIRCUIT(circ)->begin_dir_seen = 1;
+      }
       if (conn) {
         log_fn(LOG_PROTOCOL_WARN, domain,
                "Begin cell for known stream. Dropping.");
--- a/src/core/or/sendme.c
+++ b/src/core/or/sendme.c
@@ -39,6 +39,11 @@
 #define SENDME_ACCEPT_MIN_VERSION_MIN 0
 #define SENDME_ACCEPT_MIN_VERSION_MAX UINT8_MAX
 
+/* Minimum version allowed on a one-hop directory circuit. */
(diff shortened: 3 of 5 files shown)


src/core/or/relay.c
```diff
@@ -1670,6 +1670,11 @@ connection_edge_process_relay_cell(cell_t *cell, circuit_t *circ,
                "from intermediary node. Dropping.");
         return 0;
       }
+      if (circ->purpose == CIRCUIT_PURPOSE_OR) {
```
nmathewson: This code is reached for BEGIN _and_ BEGIN_DIR.

src/core/or/sendme.c
```diff
@@ -39,6 +39,11 @@
 #define SENDME_ACCEPT_MIN_VERSION_MIN 0
 #define SENDME_ACCEPT_MIN_VERSION_MAX UINT8_MAX
 
+/* Minimum version allowed on a one-hop directory circuit. */
```
nmathewson: This comment isn't accurate?

src/core/or/sendme.c
```diff
@@ -123,10 +139,17 @@ cell_v1_is_valid(const sendme_cell_t *cell, const circuit_t *circ)
...
-cell_version_is_valid(uint8_t cell_version)
+cell_version_is_valid(uint8_t cell_version, const circuit_t *circ)
 {
   int accept_version = get_accept_min_version();
 
+  /* We lookup a different consensus param when receiving a SENDME on a
+   * circuit that is for a directory request. It is only relevant if we are
+   * the directory server and thus on an OR circuit. */
+  if (CIRCUIT_IS_ORCIRC(circ) && CONST_TO_OR_CIRCUIT(circ)->begin_dir_seen) {
+    accept_version = get_dir_min_version();
```
nmathewson: This calculation is expensive enough that maybe we don't want to do it for every sendme?

src/core/or/or_circuit_st.h
```diff
@@ -61,6 +61,13 @@ struct or_circuit_t {
    *  statistics. */
   unsigned int circuit_carries_hs_traffic_stats : 1;
 
+  /** If set, this circuit got a BEGIN_DIR, meaning a directory request, and
+   * not for any hidden service documents. We use this flag for SENDME
+   * mechanism in order to know if this circuit will serve a consensus for
+   * which we need to obey a specific consensus parameter about the minimal
+   * SENDME version we accept. */
+  unsigned int begin_dir_seen : 1;
```
nmathewson: I think this might be backwards: If I understand correctly, we don't want somebody to be able to send a BEGIN_DIR in order to make all a circuit exempt from the minimum.  I think instead we wanted to have it so that if the circuit got a regular BEGIN, we know that it is _not_ a directory circuit?

src/core/or/or_circuit_st.h
```diff
@@ -61,6 +61,13 @@ struct or_circuit_t {
    *  statistics. */
   unsigned int circuit_carries_hs_traffic_stats : 1;
 
+  /** If set, this circuit got a BEGIN_DIR, meaning a directory request, and
+   * not for any hidden service documents. We use this flag for SENDME
```
nmathewson: There's nothing in this patch that checks for hidden service documents?

src/core/or/or_circuit_st.h
dgoulet-tor: Hmmm interesting, I haven't thought of gaming this by sending a `RELAY_BEGIN_DIR` and then performing other tasks than a directory request...

Actually, is this assertion always true?: if no BEGIN than must be a directory circuit?

In other words, it is safe and reliable to apply this minimal version for the SENDME as long as we haven't seen a `BEGIN` on it?

PR #1043, opened 2019-05-23
Ticket26846: Leave unused random bytes in relay cell payload

nmathewson:


src/core/or/relay.c
```diff
@@ -2027,6 +2031,84 @@ uint64_t stats_n_data_cells_received = 0;
...
+  circ->send_randomness_after_n_cells = CIRCWINDOW_START / 2 +
+    crypto_fast_rng_get_uint(get_thread_fast_rng(), CIRCWINDOW_START/2);
+}
+
+/**
+ * Any relay data payload containing fewer than this many real bytes is
+ * considered to have enough randomness to.
+ **/
+#define RELAY_PAYLOAD_LENGTH_FOR_RANDOM_SENDMES \
+  (RELAY_PAYLOAD_SIZE - CELL_PADDING_GAP - 16)
```
dgoulet-tor: What is this 16 bytes?

src/core/or/relay.c
```diff
@@ -2027,6 +2031,84 @@ uint64_t stats_n_data_cells_received = 0;
...
+ * Called when initializing a circuit, or when we have reached the end of the
+ * window in which we need to send some randomness so that incoming sendme
+ * cells will be unpredictable.  Resets the flags and picks a new window.
+ */
+void
+circuit_reset_sendme_randomness(circuit_t *circ)
+{
+  circ->have_sent_sufficiently_random_cell = 0;
+  circ->send_randomness_after_n_cells = CIRCWINDOW_START / 2 +
+    crypto_fast_rng_get_uint(get_thread_fast_rng(), CIRCWINDOW_START/2);
```
dgoulet-tor: So any circuit that transmits <= 250KB ( `(CIRCWINDOW_START / 2) * 512` ) will never see extra randomness added to its stream?

src/core/or/relay.c
nmathewson: I decided that 16 bytes would be a good amount of entropy to insert: 128 bits should be enough to make the sendme cell unpredictable.

src/core/or/relay.c
nmathewson: Unless it already has a cell with entropy added, yeah.

I copied this from Roger's branch, which added to each cell with with probability 1/CIRCWINDOW_START.  It might make more sense to use CIRCWINDOW_INCREMENT here though; what do you think?

dgoulet-tor: Yes, I was thinking that, if we limit within an increment window, this means that every SENDME cell is sure to at least had before some random added to the previous "stream of data".

PR #1056, opened 2019-05-29
Make get_proxy_type() connection-specific

nmathewson:
Previously, we were looking at our global settings to see what kind
of proxy we had.  But doing this would sometimes give us the wrong
results when we had ClientTransportPlugin configured but we weren't
using it for a particular connection.  In several places in the
code, we had added checks to see if we were _really_ using a PT or
whether we were using a socks proxy, but we had forgotten to do so
in at least once case.  Instead, since every time we call this
function we are asking about a single connection, it is probably
best just to make this function connection-specific.

Fixes bug 29670; bugfix on 0.2.6.2-alpha.


tlyu:
Temporarily closing to force Travis to rebuild.

PR #1061, opened 2019-05-30
Ticket15516 042 02

dgoulet-tor:


src/feature/hs/hs_dos.c
```diff
@@ -0,0 +1,45 @@
...
+ * Public API.
+ */
+
+/* Return true iff an INTRODUCE2 cell can be sent on the given service
+ * introduction circuit. */
+bool
+hs_dos_can_send_intro2(or_circuit_t *s_intro_circ)
+{
+  tor_assert(s_intro_circ);
+  tor_assert(TO_CIRCUIT(s_intro_circ)->purpose ==
```
asn-d6: Let's turn this into a nonfatal assert.

src/feature/hs/hs_dos.c
```diff
@@ -0,0 +1,45 @@
...
+             CIRCUIT_PURPOSE_INTRO_POINT);
+
+  /* This is called just after we got a valid and parsed INTRODUCE1 cell. The
+   * service has been found and we have its introduction circuit.
+   *
+   * First, the INTRODUCE2 bucket will be refilled (if any). Then, decremented
```
asn-d6: This might be a comment towards the token bucket API, but this unconditional refill seems a bit weird in a function that is supposed to decrement stuff. As it seems it's a conditional refill, but this is not obvious from the function name. We can either rename the token bucket function, or maybe let's clarify the cryptic "(if any)") part of the comment above.

src/feature/hs/hs_dos.c
```diff
@@ -0,0 +1,45 @@
...
+   * because we are about to send or not the cell we just got. Finally,
+   * evaluate if we can send it based on our token bucket state. */
+
+  /* Refill INTRODUCE2 bucket. */
+  token_bucket_ctr_refill(&s_intro_circ->introduce2_bucket, approx_time());
+
+  /* Decrement the bucket for this valid INTRODUCE1 cell we just got. Don't
+   * underflow else we end up with a too big of a bucket. */
+  if (token_bucket_ctr_get(&s_intro_circ->introduce2_bucket) > 0) {
```
asn-d6: That's another comment towards the token bucket API, but it's a bit weird that we have to explicitly check fort underflow like this. Is there any case where we would want the token bucket to underflow? If not, can't we handle this inside the decrement method?

src/feature/hs/hs_dos.h
```diff
@@ -0,0 +1,44 @@
...
+ **/
+
+#ifndef TOR_HS_DOS_H
+#define TOR_HS_DOS_H
+
+#include "core/or/or_circuit_st.h"
+
+#include "lib/evloop/token_bucket.h"
+
+#define HS_DOS_INTRODUCE_CELL_RATE_PER_SEC 25
```
asn-d6: DOCDOCDOC here but I guess this is the most hanging part of the branch so there is not much to document yet.

src/feature/hs/hs_intropoint.c
```diff
@@ -480,6 +484,15 @@ handle_introduce1(or_circuit_t *client_circ, const uint8_t *request,
     }
   }
 
+  /* Before sending, lets make sure this cell can be sent on the service
+   * circuit asking the DoS defenses. */
+  if (!hs_dos_can_send_intro2(service_circ)) {
+    log_info(LD_PROTOCOL, "Can't relay INTRODUCE1 v3 cell due to DoS "
```
asn-d6: Perhaps this log needs ratelimiting since it's gonna hit hard if it hits.

src/feature/rend/rendmid.c
```diff
@@ -180,6 +181,14 @@ rend_mid_introduce_legacy(or_circuit_t *circ, const uint8_t *request,
     goto err;
   }
 
+  /* Before sending, lets make sure this cell can be sent on the service
+   * circuit asking the DoS defenses. */
+  if (!hs_dos_can_send_intro2(intro_circ)) {
+    log_info(LD_PROTOCOL, "Can't relay INTRODUCE1 v2 cell due to DoS "
```
asn-d6: Perhaps this log needs ratelimiting since it's gonna hit hard if it hits.

src/feature/hs/hs_dos.c
dgoulet-tor: Probably want a new ticket for this.

src/feature/hs/hs_dos.c
dgoulet-tor: Yeah so I think a series of improvement could go towards the token bucket API. The DoS subsystem (dos.c) does way more to prevent over/underflows and handle clock jumps.

src/feature/hs/hs_dos.h
dgoulet-tor: This is temporary. I'm not even sure that value is valid yet.

src/feature/hs/hs_intropoint.c
dgoulet-tor: oh indeed!

Fixup ebb887afb9

src/feature/hs/hs_dos.c
dgoulet-tor: Fixup in 93597e17dc

asn-d6: LGTM.

src/feature/hs/hs_dos.h
asn-d6: I'd appreciate some more comments here regarding what 'rate' and 'burst' means in this context. Even a short one would be great.

src/feature/hs/hs_intropoint.c
asn-d6: LGTM!

src/feature/hs/hs_dos.c
asn-d6: Wanna open a  token bucket API ticket for this and the above comments? If we get a ticket, it's fine to not fix this right here.

src/feature/hs/hs_dos.c
```diff
@@ -7,14 +7,84 @@
...
+{
+  hs_dos_introduce_rate_per_sec = get_param_rate_per_sec(ns);
+  hs_dos_introduce_burst_per_sec = get_param_burst_per_sec(ns);
+}
+
 /*
  * Public API.
  */
 
+/* Return the INTRODUCE2 cell rate per second. */
```
asn-d6: I'm anxious that these getters will get called before `set_consensus_parameters()` gets called and they will return 0, resulting in an all-rejecting token bucket. Can we get some assurance than when the getters get called, they have been intiialized or they return the default val? What happens if we merge `hs_dos_get_intro2_rate()` and `get_param_rate_per_sec()` are you worried we are putting too much CPU on the fast path?

src/feature/hs/hs_dos.c
```diff
@@ -7,14 +7,84 @@
...
+/* Return the parameter for the introduction rate per sec. */
+static uint32_t
+get_param_rate_per_sec(const networkstatus_t *ns)
+{
+  return networkstatus_get_param(ns, "HSDoSIntroduceRate",
+                                 HS_DOS_INTRODUCE_CELL_RATE_PER_SEC,
+                                 0, INT32_MAX);
+}
+
+/* Return the parameter for the introduction burst per sec. */
```
asn-d6: Are we able to disable this feature using the consensus parameter? If yes, how does it work? Is this tested in the unittests? I don't see any special casing for this being 0.

src/feature/hs/hs_dos.c
```diff
@@ -7,14 +7,84 @@
...
+
+/* Consensus parameters. */
+static uint32_t hs_dos_introduce_rate_per_sec;
+static uint32_t hs_dos_introduce_burst_per_sec;
+
+/* Return the parameter for the introduction rate per sec. */
+static uint32_t
+get_param_rate_per_sec(const networkstatus_t *ns)
+{
+  return networkstatus_get_param(ns, "HSDoSIntroduceRate",
```
asn-d6: This seems to be a different consensus parameter than the one defined in prop305. How will these consensus parameter cooperate with the ones from prop305?

src/feature/hs/hs_dos.c
```diff
@@ -7,14 +7,84 @@
...
+#include "app/config/config.h"
+
 #include "core/or/circuitlist.h"
 
+#include "feature/nodelist/networkstatus.h"
+#include "feature/relay/routermode.h"
+
+#include "lib/evloop/token_bucket.h"
+
 #include "hs_dos.h"
```
asn-d6: I'd appreciate a file-level comment here with the aims of this subsystem and what we are hoping it achieves. I'm a bit behind on the latest experimental results with this feature, so having the goals/expectations clearly defined would be helpful to me.

src/feature/hs/hs_dos.c
teor2345: Here's one quick way to ensure that we never read zero-initialised values: initialise the static variables with the default values.

src/feature/hs/hs_intropoint.c
```diff
@@ -202,6 +203,9 @@ handle_verified_establish_intro_cell(or_circuit_t *circ,
   hs_circuitmap_register_intro_circ_v3_relay_side(circ, &auth_key);
   /* Repurpose this circuit into an intro circuit. */
   circuit_change_purpose(TO_CIRCUIT(circ), CIRCUIT_PURPOSE_INTRO_POINT);
+  /* Initialize the INTRODUCE2 token bucket for the rate limiting. */
+  token_bucket_ctr_init(&circ->introduce2_bucket, hs_dos_get_intro2_rate(),
+                        hs_dos_get_intro2_burst(), (uint32_t) approx_time());
```
teor2345: Introduction circuits can last for 18-24 hours, so it's a bit surprising that we copy the DoS values at the ESTABLISH_INTRO, and then don't change them, even if the torrc config or consensus parameters change.

That means operators will need to restart their intro point or onion service (or wait 24 hours) to overcome a DoS.

src/feature/hs/hs_dos.c
dgoulet-tor: Yes definitely. I've noted it down. Once merged, we need more robust token bucket checks.

src/feature/hs/hs_dos.h
dgoulet-tor: Tried to do that in fixup commit 143e36a09f

src/feature/hs/hs_dos.c
dgoulet-tor: Yes so prop305 does mentions a consensus parameter disabling this. It was written after this patch so because of that I've added a new commit that implements that parameter.

Commit: 4bc8e27ea8

src/feature/hs/hs_dos.c
dgoulet-tor: Yes. Fixup commit: daecc42b37

src/feature/hs/hs_intropoint.c
dgoulet-tor: _GREAT_ catch. 

So this patch for now only has consensus parameters support where torrc options (for the `ESTABLISH_INTRO` parameter comes with another ticket, prop305) aren't implemented yet.

This fix required a bit more code that is a function in the HS circuitmap module to "get all intro circuits" so we can go over them and adjust the token bucket parameters.

New commit takes care of this: d5e28d0d6560577e9e75826e6b3b1c72106ea7e2

src/feature/hs/hs_dos.c
dgoulet-tor: Fixup commit: 22232698ca

src/feature/hs/hs_dos.c
dgoulet-tor: Fixup commit: 118cbd4319

src/feature/hs/hs_dos.h
asn-d6: Thanks!

src/feature/hs/hs_dos.c
asn-d6: Ah that looks good. Perhaps we can also change the name of these constants to `HS_DOS_INTRODUCE_DEFAULT_CELL_RATE_PER_SEC` to denote that they are default values and not the final truth?

src/feature/hs/hs_dos.c
asn-d6: LGTM!

src/feature/hs/hs_dos.c
asn-d6: LGTM!

src/feature/hs/hs_dos.c
asn-d6: LGTM!

src/feature/hs/hs_intropoint.c
asn-d6: LGTM, but I will let teor take a look as well. Not resolving yet.

teor2345: Ok, looks good to me.

Let's make sure we remember to update all the circuits when the torrc options change :-)

src/feature/hs/hs_dos.c
dgoulet-tor: Commit 49c9f21255]

PR #1084, opened 2019-06-06
Ticket16564: reject bridge descriptors on relay authorities

teor2345:


src/feature/nodelist/routerlist.c
```diff
@@ -1516,6 +1516,8 @@ router_reject_same_desc_helper(routerinfo_t *router, const char **msg,
   return ROUTER_CONTINUE_CHECKS;
 }
 
+#ifdef HAVE_MODULE_DIRAUTH
```
nmathewson: I think that elsewhere in the dirauth api, we define these functions so that they are stubbed out when HAVE_MODULE_DIRAUTH is not defined, and so that other modules don't need to use HAVE_MODULE_DIRAUTH.  Does that make sense here?

PR #1085, opened 2019-06-06
30784: Windows put its timeval definitions in winsock2.h.

saurik:
In compat_winthreads.c, tor_cond_wait takes and uses an argument of type struct timeval without including winsock2.h (which is where timeval comes from, per the below documentation URL). This might work by accident with the particular toolchain normally used by the Tor project, but with MinGW this currently is causing the below compiler error.

https://docs.microsoft.com/en-us/windows/desktop/api/winsock/ns-winsock-timeval

```
tor/src/lib/thread/compat_winthreads.c:160:22: error: incomplete definition of type 'struct timeval'
    ms_orig = ms = tv->tv_sec*1000 + (tv->tv_usec+999)/1000;
                   ~~^
tor/src/lib/thread/threads.h:27:8: note: forward declaration of 'struct timeval'
struct timeval;
       ^
tor/src/lib/thread/compat_winthreads.c:160:41: error: incomplete definition of type 'struct timeval'
    ms_orig = ms = tv->tv_sec*1000 + (tv->tv_usec+999)/1000;
                                      ~~^
../.././vpn/tor/src/lib/thread/threads.h:27:8: note: forward declaration of 'struct timeval'
struct timeval;
       ^
2 errors generated.
```

### DIFF
--- a/src/lib/thread/compat_winthreads.c
+++ b/src/lib/thread/compat_winthreads.c
@@ -12,6 +12,7 @@
 
 #ifdef _WIN32
 
+#include <winsock2.h>
 #include <windows.h>
 #include <process.h>
 #include "lib/thread/threads.h"


saurik:
(Based on your CI setup, it seems like you are also using MinGW, but you might be using a different version of MinGW or a different Tor configuration or a different version of an upstream dependency or something, leading to the accidental "this file doesn't end up with winsock2.h being included" issue.)

teor2345:
Thanks for this fix.

We track bugs using our own Trac instance, I have opened a ticket here:
https://trac.torproject.org/projects/tor/ticket/30784

saurik:
So, I'm the reason this ticket exists (having filed a pull request on GitHub), and I've now spent some more time trying to figure out why this would work for you but not for me; I had originally thought it was because I was using MinGW (and so tend to run into issues with subtle differences in header files), but when I saw your CI system was using MinGW I figured I should dig deeper... and I tracked it down to my toolchain setup (incorrectly) having hardcoded WIN32_LEAN_AND_MEAN, which meant that the #include you already have for <windows.h> wasn't sufficient, but it should have been, so I'm sorry: Im just going to go ahead and close this ticket as invalid. (Again: sorry.)

PR #1087, opened 2019-06-06
Ticket30382 042 01

dgoulet-tor:


src/app/config/config.c
```diff
@@ -7193,6 +7193,9 @@ parse_port_config(smartlist_t *out,
         } else if (!strcasecmp(elt, "KeepAliveIsolateSOCKSAuth")) {
           socks_iso_keep_alive = ! no;
           continue;
+        } else if (!strcasecmp(elt, "ExtendedErrors")) {
```
asn-d6: We need to document this new option in the man page.

src/feature/hs/hs_cache.c
```diff
@@ -395,19 +395,29 @@ lookup_v3_desc_as_client(const uint8_t *key)
...
+  ret = hs_client_decode_descriptor(desc_str, service_identity_pk, &desc);
+  if (ret != HS_DESC_DECODE_OK &&
+      ret != HS_DESC_DECODE_NEED_CLIENT_AUTH &&
+      ret != HS_DESC_DECODE_BAD_CLIENT_AUTH) {
+    /* In the case of a missing or bad client authorization, we'll keep the
+     * descriptor in the cache because those credentials can arrive later. */
     goto end;
   }
-  tor_assert(desc);
+  /* Make sure we do have a descriptor if decoding was successful. */
```
asn-d6: Isn't it the case we also have the descriptor in case of NEED_CLIENT_AUTH and BAD_CLIENT_AUTH, it just doesn't have `->desc`?

src/feature/hs/hs_cache.c
```diff
@@ -635,9 +648,19 @@ cache_store_as_client(hs_cache_client_descriptor_t *client_desc)
   tor_assert(client_desc);
 
   /* Check if we already have a descriptor from this HS in cache. If we do,
-   * check if this descriptor is newer than the cached one */
+   * check if this descriptor is newer than the cached one only if we have a
+   * decoded descriptor. We do keep non-decoded descriptor that requires
+   * client authorization. */
   cache_entry = lookup_v3_desc_as_client(client_desc->key.pubkey);
   if (cache_entry != NULL) {
+    /* Signalling an undecrypted descriptor. We'll always replaced the one we
```
asn-d6: nit: s/replaced/replace

src/feature/hs/hs_descriptor.c
```diff
@@ -2580,17 +2576,17 @@ hs_desc_decode_descriptor(const char *encoded,
...
 
   ret = hs_desc_decode_superencrypted(desc, &desc->superencrypted_data);
-  if (ret < 0) {
+  if (ret != HS_DESC_DECODE_OK) {
     goto err;
   }
 
   ret = hs_desc_decode_encrypted(desc, client_auth_sk, &desc->encrypted_data);
-  if (ret < 0) {
+  if (ret != HS_DESC_DECODE_OK) {
```
asn-d6: I don't quite understand the code flow here that ends up with us storing the failed-to-decrypt descriptors. If we go to `err` when we fail to decrypt here, and the err frees the descriptor, how do we end up with the descriptor in the cache?

I guess this works in the end because the unittests say so, but I really have trouble seeing this in the code... Is there a way to clarify this in the code with comments or anything?

src/feature/hs/hs_descriptor.c
```diff
@@ -2277,12 +2274,14 @@ desc_decode_encrypted_v3(const hs_descriptor_t *desc,
...
       log_warn(LD_REND, "Client authorization for requested onion address "
                         "is invalid. Can't decrypt the descriptor.");
+      ret = HS_DESC_DECODE_BAD_CLIENT_AUTH;
     } else {
       /* Inform at notice level that the onion address requested can't be
        * reached without client authorization most likely. */
       log_notice(LD_REND, "Fail to decrypt descriptor for requested onion "
                         "address. It is likely requiring client "
                         "authorization.");
+      ret = HS_DESC_DECODE_NEED_CLIENT_AUTH;
```
asn-d6: Blocker: This does not seem to work smoothly for me. In particular, it takes 2 minutes before the SOCKS error gets propagated to my torsocks. This can't work for a browser tab, since the user will have long closed it.
```
         Jun 12 13:21:28.352 [notice] Fail to decrypt descriptor for requested onion address. It is likely requiring client authorization.
         Jun 12 13:21:38.389 [notice] Fail to decrypt descriptor for requested onion address. It is likely requiring client authorization.
         Jun 12 13:21:55.468 [notice] Fail to decrypt descriptor for requested onion address. It is likely requiring client authorization.
         Jun 12 13:22:16.417 [notice] Fail to decrypt descriptor for requested onion address. It is likely requiring client authorization.
         Jun 12 13:22:37.447 [notice] Fail to decrypt descriptor for requested onion address. It is likely requiring client authorization.
         Jun 12 13:22:55.453 [notice] Fail to decrypt descriptor for requested onion address. It is likely requiring client authorization.
         Jun 12 13:23:17.476 [notice] Fail to decrypt descriptor for requested onion address. It is likely requiring client authorization.
         Jun 12 13:23:28.315 [notice] Tried for 120 seconds to get a connection to 6eqv54nzmho6tmgbwfm76cciveryhunmz62c3tm4ez6wsbzricz5ysad:5858. Giving up. (waiting for rendezvous desc)
         ...
         1560335008 ERROR torsocks[10662]: Socks5 server replied an unknown code 244 (in socks5_recv_connect_reply() at socks5.c:559)
```

src/feature/hs/hs_descriptor.c
```diff
@@ -2277,12 +2274,14 @@ desc_decode_encrypted_v3(const hs_descriptor_t *desc,
        * authorization is failing. */
       log_warn(LD_REND, "Client authorization for requested onion address "
                         "is invalid. Can't decrypt the descriptor.");
+      ret = HS_DESC_DECODE_BAD_CLIENT_AUTH;
     } else {
       /* Inform at notice level that the onion address requested can't be
        * reached without client authorization most likely. */
       log_notice(LD_REND, "Fail to decrypt descriptor for requested onion "
```
asn-d6: We get this message 6 times everytime there is a missing client auth. Can this be fixed to only receive it once? Or zero?

src/feature/hs/hs_descriptor.h
```diff
@@ -69,6 +69,17 @@ typedef enum {
   HS_DESC_AUTH_ED25519 = 1
 } hs_desc_auth_type_t;
 
+/* Error code when decoding a descriptor. */
```
asn-d6: Perhaps let's document the individual error codes here?

src/test/test_hs_client.c
```diff
@@ -1007,6 +1029,104 @@ test_close_intro_circuits_new_desc(void *arg)
...
+static void
+test_socks_hs_errors(void *arg)
+{
+  int ret;
+  char *desc_encoded = NULL;
+  ed25519_keypair_t service_kp;
+  ed25519_keypair_t signing_kp;
+  entry_connection_t *socks_conn = NULL;
+  dir_connection_t *dir_conn = NULL;
+  hs_descriptor_t *desc;
```
asn-d6: This being non-NULL causes compilation errors that break travis.

src/feature/hs/hs_client.c
```diff
@@ -100,6 +101,45 @@ fetch_status_should_close_socks(hs_client_fetch_status_t status)
...
 }
 
+/* Return a newly allocated list of all the entry connections that matches the
+ * given service identity pk. If service_identity_pk is NULL, all entry
+ * connections with an hs_ident are returned.
+ *
+ * Caller must free the returned list but does NOT have ownership of the
+ * object inside thus they have to remain untouched. */
+static smartlist_t *
+find_entry_conns(const ed25519_public_key_t *service_identity_pk)
```
asn-d6: Suggestion: Let's rename to `find_entry_conns_for_onion_service` or something, to make it a bit more HS-centric? Because initially I thought it just returned everything.

src/feature/hs/hs_client.c
```diff
@@ -262,21 +289,20 @@ close_all_socks_conns_waiting_for_desc(const ed25519_public_key_t *identity_pk,
...
 
-  SMARTLIST_FOREACH_BEGIN(conns, connection_t *, base_conn) {
+  SMARTLIST_FOREACH_BEGIN(entry_conns, entry_connection_t *, entry_conn) {
     hs_client_fetch_status_t status;
-    const edge_connection_t *edge_conn =
-      ENTRY_TO_EDGE_CONN(TO_ENTRY_CONN(base_conn));
+    edge_connection_t *edge_conn = ENTRY_TO_EDGE_CONN(entry_conn);
+    connection_t *base_conn = &edge_conn->base_;
 
     /* Ignore non HS or non v3 connection. */
```
asn-d6: I think there is no reason for this check anymore now that the new helper function is used and already checks this.

src/feature/hs/hs_cache.c
```diff
@@ -756,27 +780,41 @@ hs_cache_lookup_as_client(const ed25519_public_key_t *key)
...
+ *                                  still stored.
+ *
+ *  Any other codes means indicate where the error occured and the descriptor
+ *  was not stored. */
+hs_desc_decode_status_t
 hs_cache_store_as_client(const char *desc_str,
                          const ed25519_public_key_t *identity_pk)
 {
+  hs_desc_decode_status_t ret;
   hs_cache_client_descriptor_t *client_desc = NULL;
```
asn-d6: We need to update the documentation of `hs_cache_client_descriptor_t` and mentioned that `desc` can be NULL since it currently claims the opposite. It would also be nice to say a few more things about the logic of it being NULL in those comments, so that there is a central place that explains it.

src/feature/hs/hs_descriptor.c
dgoulet-tor: Well I think you are just not seeing the top layer here with the HS cache.

The HS cache (client side), has a top level object that contains the raw blob desc _and_ the decoded descriptor.

In case of a bad client auth, we can NOT have the decoded descriptor which is why this code path returns an error. Remember, everything in `hs_descriptor.c` is about a _decoded_ descriptor object.

So on error of decoding a descriptor, we will keep the undecoded blob in the top level cache object making the "desc == NULL" in that object.

src/feature/hs/hs_descriptor.c
dgoulet-tor: Yeah, this is the SOCKS timeout that mcs/brade reported... It needs to be investigated separately for now. Ticket is about to be opened about this.

src/feature/hs/hs_descriptor.h
dgoulet-tor: Done.

src/test/test_hs_client.c
dgoulet-tor: Fixed in the code.

src/feature/hs/hs_client.c
dgoulet-tor: Well it is in `hs_client.c` ? ... I mean, it can't really be looking for any other kind of entry connections? Especially when it takes a "service identity key" ?

src/feature/hs/hs_client.c
dgoulet-tor: Indeed!

src/feature/hs/hs_cache.c
dgoulet-tor: Done.

src/feature/hs/hs_cache.c
dgoulet-tor: Fixed

src/feature/hs/hs_cache.c
dgoulet-tor: Yes, this is why the if() check before?

src/app/config/config.c
dgoulet-tor: Yup, will be there in the new branhc.

PR #1089, opened 2019-06-08
Make evloop into a subsystem.

nmathewson:
Note that the event base object is _not_ created from the initialize
function, since it is configuration-dependent.  This will wait until
configuration is integrated into subsystems.

Closes ticket 30806.


src/lib/evloop/evloop_sys.c
```diff
@@ -0,0 +1,46 @@
...
+
+#include "orconfig.h"
+#include "lib/subsys/subsys.h"
+#include "lib/evloop/compat_libevent.h"
+#include "lib/evloop/evloop_sys.h"
+
+static int
+subsys_evloop_initialize(void)
+{
+  if (tor_init_libevent_rng() < 0)
```
dgoulet-tor: We are losing the `log_warn()` here, is it something desirable?

PR #1103, opened 2019-06-12
Chutney ci v3 029

nmathewson:


.travis.yml
```diff
@@ -40,6 +40,10 @@ matrix:
     # - env: HARDENING_OPTIONS=""
     ## We check asciidoc with distcheck, to make sure we remove doc products
     - env: DISTCHECK="yes" ASCIIDOC_OPTIONS="" SKIP_MAKE_CHECK="yes"
+    # We also try running a hardened clang build with chutney on Linux.
+    - env: CHUTNEY="yes" SKIP_MAKE_CHECK="yes" CHUTNEY_ALLOW_FAILURES="2"
+      compiler: clang
+      os: linux
```
teor2345: This line is redundant, but harmless, because linux is the first os listed in the config file.

I don't think we need to change it.

PR #1113, opened 2019-06-15
Ticket30893

nmathewson:


src/app/config/config.c
```diff
@@ -8177,10 +8177,9 @@ getinfo_helper_config(control_connection_t *conn,
       switch (var->type) {
         case CONFIG_TYPE_STRING: type = "String"; break;
         case CONFIG_TYPE_FILENAME: type = "Filename"; break;
-        case CONFIG_TYPE_UINT: type = "Integer"; break;
+        case CONFIG_TYPE_POSINT: type = "Integer"; break;
         case CONFIG_TYPE_UINT64: type = "Integer"; break;
         case CONFIG_TYPE_INT: type = "SignedInteger"; break;
-        case CONFIG_TYPE_PORT: type = "Port"; break;
```
tlyu: These changes delete `CONFIG_TYPE_PORT`. Are we still using it elsewhere? (We don't seem to be, but I could have missed something).

nmathewson: We aren't using it anywhere else, no.

src/app/config/confparse.h
```diff
@@ -57,8 +55,8 @@ typedef enum config_type_t {
 typedef union {
   char **STRING;
   char **FILENAME;
-  int *UINT; /* yes, really: Even though the confparse type is called
-              * "UINT", it still uses the C int type -- it just enforces that
+  int *POSINT; /* yes, really: Even though the confparse type is called
+              * "POSINT", it still uses the C int type -- it just enforces that
```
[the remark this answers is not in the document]
nmathewson: I'm happy to clean this comment up, but I'd rather do so as part of #30864, since that branch already moves this declaration around.

How about "This is indeed an int, not a unsigned int. For historical reasons, many configuration values are restricted to the range [0,INT_MAX]." ?

PR #1116, opened 2019-06-18
Ticket30864

nmathewson:


scripts/maint/checkSpace.pl
```diff
@@ -177,7 +177,7 @@ sub msg {
                     $1 ne "elsif" and $1 ne "WINAPI" and $2 ne "WINAPI" and
                     $1 ne "void" and $1 ne "__attribute__" and $1 ne "op" and
                     $1 ne "size_t" and $1 ne "double" and $1 ne "uint64_t" and
-                    $1 ne "workqueue_reply_t") {
+                    $1 ne "workqueue_reply_t" and $1 ne "bool") {
```
tlyu: This list of keywords, types, etc. is getting rather unwieldy to have in the expression of an if-statement. Maybe make a list or something?

Makefile.am
```diff
@@ -74,6 +75,7 @@ TOR_UTIL_TESTING_LIBS = \
         src/lib/libtor-geoip-testing.a \
 	src/lib/libtor-process-testing.a \
         src/lib/libtor-buf-testing.a \
```
tlyu: You might not see it in github's rendering, but this line and the libto-geoip line above use spaces instead of tabs. Dunno if you want to fix it here or as a separate commit or ticket.

scripts/maint/checkSpace.pl
nmathewson: Opened trac#30990 for this.

Makefile.am
nmathewson: Added #30991 for an automatic fix; I think if we don't automate this, I will continue messing it up in the future.

src/lib/encoding/confline.c
```diff
@@ -33,14 +33,23 @@ config_line_append(config_line_t **lst,
...
   newline->key = tor_strdup(key);
   newline->value = tor_strdup(val);
   newline->next = NULL;
+
+  config_line_append_line(lst, newline);
+}
+
+/** Helper: append <b>newline</b> to the end of <b>lst</b>. */
+void
+config_line_append_line(config_line_t **lst,
```
tlyu: I think this name isn't great. Is the main difference between `config_line_append()` and `config_line_append_line()` that the first one duplicates stuff? Maybe rename the first to `config_line_append_alloc()` and the helper to `config_line_append()`, or `config_line_append()` and `config_line_append_noalloc()`?

Also in terms of pre-existing technical debt, this is a linked list implementation that isn't using the `tor_queue.h` macros. Maybe we should fix that at some point? (I think it's fine if you open a new ticket for this.)

src/lib/confmgt/var_type_def_st.h
```diff
@@ -0,0 +1,147 @@
+/* Copyright (c) 2001 Matej Pfajfar.
+ * Copyright (c) 2001-2004, Roger Dingledine.
+ * Copyright (c) 2004-2006, Roger Dingledine, Nick Mathewson.
+ * Copyright (c) 2007-2019, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
+ * @file typedvar.h
```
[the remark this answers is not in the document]
nmathewson: Fixed in e3ccf37e2545ad

src/lib/encoding/confline.c
nmathewson: Wrt  config_line_append_line, here's an easier solution: I think we can revert the commit that introduced this.  I wrongly thought that we would need this function, but looking through the branch, it seems that we never actually started using it.

I agree that a new ticket is the right place for changing config_line_t to use SLIST.  I've done so with trac#31074.

PR #1122, opened 2019-06-19
hs-v3: Disallow single hop client to post/get a descriptor

dgoulet-tor:
Closes #24964

### DIFF
--- a/src/feature/dircommon/directory.c
+++ b/src/feature/dircommon/directory.c
@@ -7,6 +7,10 @@
 
 #include "app/config/config.h"
 #include "core/mainloop/connection.h"
+#include "core/or/circuitlist.h"
+#include "core/or/connection_edge.h"
+#include "core/or/connection_or.h"
+#include "core/or/channeltls.h"
 #include "feature/dircache/dircache.h"
 #include "feature/dircache/dirserv.h"
 #include "feature/dirclient/dirclient.h"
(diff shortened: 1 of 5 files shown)


teor2345:
Closing and re-opening pull request, because Appveyor didn't even run on it.

src/feature/dircommon/directory.c
```diff
@@ -167,6 +175,67 @@ connection_dir_is_encrypted(const dir_connection_t *conn)
...
+  if (!connection_dir_is_encrypted(dir_conn)) {
+    return false;
+  }
+
+  /*
+   * Buckle up, we'll do a deep dive into the connection in order to get the
+   * final connection channel of that connection in order to figure out if
+   * this is a client or relay link.
+   *
+   * We go: dir_conn -> linked_conn -> edge_conn -> on_circuit -> p_chan.
```
asn-d6: I'm slightly afraid of this function giving a false negative and hence marking a connection as non-anonymous when it is, since that will have serious reachability issues for onion service. The problem is that the ladder from dir_conn -> p_chan is so complicated. Can we BUG out even more on the checks below if we consider them invariants, so that if this function does not work as intended we learn it soon enough?

asn-d6: Perhaps we can make this dir_conn -> p_chan ladder a function of its own in case someone else wants to learn how to do it?

teor2345: Sounds like a good idea.

In the original code, I had info logs. We should switch to info logs for the failures that actually happen.
https://github.com/teor2345/tor-old/commit/10290066c8ee6b4aa40ec048222fdd4f572ef8d9#diff-c56fd972333216da3bb1852bcc89f57dR1587

teor2345: Maybe we should make each step into a function, and then the whole thing into a function?

That way, we can do the casts and the mark for close / closed checks.

PR #1123, opened 2019-06-19
Ticket30914

nmathewson:


src/app/config/config.c
```diff
@@ -253,23 +254,45 @@ static config_abbrev_t option_abbrevs_[] = {
  * members with CONF_CHECK_VAR_TYPE. */
 DUMMY_TYPECHECK_INSTANCE(or_options_t);
 
-/** An entry for config_vars: "The option <b>name</b> has type
+/** An entry for config_vars: "The option <b>varname</b> has type
```
teor2345: Please document macro argument initvalue.

src/app/config/statefile.c
```diff
@@ -71,8 +71,10 @@ static config_abbrev_t state_abbrevs_[] = {
 DUMMY_TYPECHECK_INSTANCE(or_state_t);
 
 /*XXXX these next two are duplicates or near-duplicates from config.c */
```
teor2345: Will these duplicates will be refactored in some future PR?

Same question for shared random and other modules.

src/lib/conf/conftypes.h
```diff
@@ -0,0 +1,139 @@
...
+                             * mixed with other keywords. */
+  CONFIG_TYPE_LINELIST_V,   /**< Catch-all "virtual" option to summarize
+                             * context-sensitive config lines when fetching.
+                             */
+  // XXXX this doesn't belong at this level of abstraction.
+  CONFIG_TYPE_ROUTERSET,    /**< A list of router names, addrs, and fps,
+                             * parsed into a routerset_t. */
+  CONFIG_TYPE_OBSOLETE,     /**< Obsolete (ignored) option. */
+  CONFIG_TYPE_EXTENDED,     /**< Extended type; definition will appear in
+                             * pointer. */
```
teor2345: Are extended types also going to become part of a higher level of abstraction?

src/lib/confmgt/type_defs.c
```diff
@@ -0,0 +1,727 @@
...
+
+static int
+double_parse(void *target, const char *value, char **errmsg,
+             const void *params)
+{
+  (void)params;
+  (void)errmsg;
+  double *v = (double*)target;
+  // XXXX This is the preexisting behavior, but we should detect errors here.
+  *v = atof(value);
```
teor2345: When are we going to detect errors here?

src/test/test_confparse.c
```diff
@@ -46,15 +48,17 @@ typedef struct test_struct_t {
 
 static test_struct_t test_struct_t_dummy;
 
-#define VAR(name,conftype,member,initvalue)                             \
-  { name, CONFIG_TYPE_##conftype, offsetof(test_struct_t, member),      \
+#define VAR(varname,conftype,member,initvalue)                          \
```
teor2345: These are also duplicates, which should get removed at some future time.
(Unless they are necessary for low-level tests. If they are, they should get a comment saying so.)

src/test/test_confparse.c
```diff
@@ -805,6 +832,77 @@ test_confparse_extra_lines(void *arg)
...
+  tt_assert(ok);
+
+  /* Try a couple of unitless values. */
+  tt_int_op(config_parse_interval("10", &ok), OP_EQ, 10);
+  tt_assert(ok);
+  tt_u64_op(config_parse_interval("15.0", &ok), OP_EQ, 15);
+  tt_assert(ok);
+
+  /* u64 overflow */
+  /* XXXX our implementation does not currently detect this. See bug 30920. */
```
teor2345: Are we going to fix #30920 as part of this refactor?
Preserving existing behaviour is also ok.

src/test/test_confparse.c
```diff
@@ -805,6 +832,77 @@ test_confparse_extra_lines(void *arg)
...
+  /* i32 overflow */
+  tt_int_op(config_parse_interval("1000 months", &ok), OP_EQ, -1);
+  tt_assert(!ok);
+  tt_int_op(config_parse_msec_interval("4 weeks", &ok), OP_EQ, -1);
+  tt_assert(!ok);
+
+  /* bad units */
+  tt_u64_op(config_parse_memunit("7 nybbles", &ok), OP_EQ, 0);
+  tt_assert(!ok);
+  // XXXX these next two should return -1 according to the documentation.
```
teor2345: Is this comment still true?

src/app/config/config.c
nmathewson: This happens later in the #29211 branch, in c553750e32d1bf669a3e8308fa44319954a627ca when we move all these macros into src/lib/confmacros.h . I believe that is part of #30935.

src/app/config/statefile.c
nmathewson: Yes: also part of #30935.

src/lib/conf/conftypes.h
nmathewson: I think so, if I understand your question -- see the definition of struct_member_t.  The first extended type is routerset, which is first implemented in e16b90b88a76.

src/lib/confmgt/type_defs.c
nmathewson: I have opened a ticket as https://trac.torproject.org/projects/tor/ticket/31475#ticket

src/test/test_confparse.c
nmathewson: This also gets fixed when we combine everything into serc/lib/confmacros.h in c553750 as a part of #30935.

src/test/test_confparse.c
nmathewson: I think it should get fixed independently, after this code is stable.

src/test/test_confparse.c
nmathewson: I believe it is, yes.  The documentation says that if the intervals are bad it should return -1 IIC

teor2345: Ok so let's update the documentation for parse_interval() to document the special cases.

The doc change might belong in another ticket.

PR #1145, opened 2019-06-28
#31012: Make tor-print-ed-signing-cert output more machine readable

rl1987:
https://trac.torproject.org/projects/tor/ticket/31012

### DIFF
--- a/src/tools/tor-print-ed-signing-cert.c
+++ b/src/tools/tor-print-ed-signing-cert.c
@@ -75,7 +75,7 @@ main(int argc, char **argv)
 
   printf("Expires at: %s\n", ctime(&expiration));
   printf("RFC 822 timestamp: %s\n", rfc822_str);
-  printf("UNIX timestamp: %ld\n", expiration);
+  printf("UNIX timestamp: %ld\n", (long int)expiration);
 
   ed25519_cert_free(cert);
 
--- a/src/tools/tor-print-ed-signing-cert.c
+++ b/src/tools/tor-print-ed-signing-cert.c
@@ -5,6 +5,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <time.h>
+#include <locale.h>
 
 #include "trunnel/ed25519_cert.h"
 #include "lib/cc/torint.h"  /* TOR_PRIdSZ */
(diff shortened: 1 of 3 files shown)


src/tools/tor-print-ed-signing-cert.c
```diff
@@ -55,9 +56,26 @@ main(int argc, char **argv)
     return -5;
   }
 
-  time_t expires_at = (time_t)cert->exp_field * 60 * 60;
+  const time_t expiration = (const time_t)cert->exp_field * 60 * 60;
 
-  printf("Expires at: %s", ctime(&expires_at));
+  const struct tm *expires_at = localtime(&expiration);
+
+  char rfc822_str[16] = "";
```
teor2345: Looks like there's some kind of memory safety issue here, I suggest using 64 characters for RFC 822. (It's about 32, but let's not mess around.)

src/tools/tor-print-ed-signing-cert.c
```diff
@@ -55,9 +56,26 @@ main(int argc, char **argv)
...
+  setlocale(LC_TIME, "en_US_POSIX");
+
+/* Yes, we're fine with RFC822 being written in 1982 and not addressing Y2K. */
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-y2k"
+  strftime(rfc822_str, sizeof(rfc822_str), "%a, %d %b %y %T %z", expires_at);
+  // Format string taken from Linux strftime(3) manpage.
+#pragma GCC diagnostic pop
+
+  // XXX: On macOS, %z prints a single digit for timezone offset.
```
teor2345: Making the buffer longer should also fix your macOS truncation issue.

src/tools/tor-print-ed-signing-cert.c
```diff
@@ -55,9 +56,26 @@ main(int argc, char **argv)
...
+  const struct tm *expires_at = localtime(&expiration);
+
+  char rfc822_str[16] = "";
+
+  setlocale(LC_TIME, "en_US_POSIX");
+
+/* Yes, we're fine with RFC822 being written in 1982 and not addressing Y2K. */
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-y2k"
+  strftime(rfc822_str, sizeof(rfc822_str), "%a, %d %b %y %T %z", expires_at);
```
teor2345: strftime has a return value: checking it would have caught the truncation bug.
Let's check it anyway: it's good practice.

src/tools/tor-print-ed-signing-cert.c
```diff
@@ -55,9 +56,26 @@ main(int argc, char **argv)
...
+  // Format string taken from Linux strftime(3) manpage.
+#pragma GCC diagnostic pop
+
+  // XXX: On macOS, %z prints a single digit for timezone offset.
+
+  printf("Expires at: %s\n", ctime(&expiration));
```
teor2345: ctime() comes with a newline, you don't need an extra one.

PR #1154, opened 2019-06-28
Ticket31030

nmathewson:


src/test/test_introduce.c
```diff
@@ -383,8 +383,10 @@ make_intro_from_plaintext(
 
   /* Output the cell */
   *cell_out = cell;
+  cell = NULL;
 
  done:
+  tor_free(cell);
```
dgoulet-tor: This can't be right... We set `cell` in `cell_out` so freeing it will explode.

nmathewson: But immediately after, we set `cell = NULL`, so tor_free(cell) will do nothing... assuming that we reach the end of the function through this path.  If we reach it through tt_assert() failing, though, we need to free the cell that we allocated earlier.

PR #1161, opened 2019-07-01
[Tor Trac #25568]: hs: Lookup failure cache when introducing to an intro point

neelchauhan:
For the ticket [here](https://trac.torproject.org/projects/tor/ticket/25568).


changes/bug25568
```diff
@@ -0,0 +1,4 @@
+  o Minor bugfixes (onion services):
```
teor2345: This commit's message has a typo, it should say "Bug"

neelchauhan: Fixed it.

PR #1164, opened 2019-07-02
#5304: Support for OutboundBindAddressPT

ahf:
Add support for OutboundBindAddressPT in torrc and expose the value(s) to pluggable transports.

### DIFF
--- a/doc/tor.1.txt
+++ b/doc/tor.1.txt
@@ -735,6 +735,17 @@ GENERAL OPTIONS
     This setting will be ignored
     for connections to the loopback addresses (127.0.0.0/8 and ::1).
 
+[[OutboundBindAddressPT]] **OutboundBindAddressPT** __IP__::
+    Request that pluggable transports makes all outbound connections
+    originate from the IP address specified. Because outgoing connections
+    are handled by the pluggable transport itself, it is not possible for
+    Tor to enforce whether the pluggable transport honors this option. This
+    option overrides **OutboundBindAddress** for the same IP version. This
+    option may be used twice, once with an IPv4 address and once with an
+    IPv6 address. IPv6 addresses should be wrapped in square brackets. This
+    setting will be ignored for connections to the loopback addresses
+    (127.0.0.0/8 and ::1).
+
 [[PidFile]] **PidFile** __FILE__::
     On startup, write our PID to FILE. On clean shutdown, remove
     FILE. Can not be changed while tor is running.
--- a/scripts/maint/practracker/exceptions.txt
+++ b/scripts/maint/practracker/exceptions.txt
@@ -140,7 +140,7 @@ problem function-size /src/feature/client/entrynodes.c:entry_guards_upgrade_wait
 problem function-size /src/feature/client/entrynodes.c:entry_guard_parse_from_state() 246
 problem function-size /src/feature/client/transports.c:handle_proxy_line() 108
 problem function-size /src/feature/client/transports.c:parse_method_line_helper() 112
-problem function-size /src/feature/client/transports.c:create_managed_proxy_environment() 109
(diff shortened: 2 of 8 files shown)


doc/tor.1.txt
```diff
@@ -735,6 +735,17 @@ GENERAL OPTIONS
...
+[[OutboundBindAddressPT]] **OutboundBindAddressPT** __IP__::
+    Request that pluggable transports makes all outbound connections
+    originate from the IP address specified. Because outgoing connections
+    are handled by the pluggable transport itself, it is not possible for
+    Tor to enforce whether the pluggable transport honors this option. This
+    option overrides **OutboundBindAddress** for the same IP version. This
+    option may be used twice, once with an IPv4 address and once with an
+    IPv6 address. IPv6 addresses should be wrapped in square brackets. This
+    setting will be ignored for connections to the loopback addresses
+    (127.0.0.0/8 and ::1).
```
teor2345: Do we expect PTs to connect to the loopback addresses?
(I guess they do in chutney.)

Do we expect them to ignore outbound bind address for the loopback addresses?

scripts/maint/practracker/exceptions.txt
```diff
@@ -140,7 +140,7 @@ problem function-size /src/feature/client/entrynodes.c:entry_guards_upgrade_wait
 problem function-size /src/feature/client/entrynodes.c:entry_guard_parse_from_state() 246
 problem function-size /src/feature/client/transports.c:handle_proxy_line() 108
 problem function-size /src/feature/client/transports.c:parse_method_line_helper() 112
-problem function-size /src/feature/client/transports.c:create_managed_proxy_environment() 109
+problem function-size /src/feature/client/transports.c:create_managed_proxy_environment() 140
```
teor2345: It feels like there's a lot of code duplication in this patch.
I'm going to make some suggestions, feel free to do the ones that make sense.

src/feature/client/transports.c
```diff
@@ -1902,3 +1933,46 @@ managed_proxy_severity_parse(const char *severity)
...
+ * OutboundBindAddress or OutboundBindAddressPT. */
+STATIC const tor_addr_t *
+managed_proxy_outbound_address(const or_options_t *options, sa_family_t family)
+{
+  tor_assert(options);
+
+  const tor_addr_t *address = NULL;
+  int family_index;
+
+  switch (family) {
```
teor2345: There are now two copies of this code in tor, the other copy is in conn_get_outbound_address().
Let's refactor this code into a new function, maybe in config.c?

src/feature/client/transports.c
```diff
@@ -1902,3 +1933,46 @@ managed_proxy_severity_parse(const char *severity)
...
+    family_index = 1;
+    break;
+  default:
+    /* LCOV_EXCL_START */
+    tor_assert_unreached();
+    return NULL;
+    /* LCOV_EXCL_STOP */
+  }
+
+  /* We start by checking if the user specified an address in
```
teor2345: There are now three copies of this code in tor, the other two copies are in conn_get_outbound_address().
This implementation is different to the other copies.

Let's refactor this code into a new function, maybe in config.c?

A nice interface could look like:
STATIC const tor_addr_t *
config_get_outbound_address(const or_options_t *options, sa_family_t family, outbound_addr_t type)

And it would also solve the other code duplication issue.

src/feature/client/transports.c
```diff
@@ -1439,6 +1439,37 @@ create_managed_proxy_environment(const managed_proxy_t *mp)
    */
```
teor2345: If you'd like to remove the practracker exception for this function entirely, you could split out the address/port/proxy settings into another function. And leave the transport/state/other settings.

Just an idea. Not sure how serious we are about practracker.

PR #1170, opened 2019-07-08
remove specified target_hopnum from relay-side machines

pylls:
For circuit padding machines, the `target_hopnum`field in `struct circpad_machine_spec_t`is only for origin-side machines. This commit removes setting this field for the relay-side machines.


pylls:
https://trac.torproject.org/projects/tor/ticket/31112

PR #1171, opened 2019-07-08
Bug31080 041

nmathewson:


src/lib/err/backtrace.c
```diff
@@ -115,7 +115,7 @@ clean_backtrace(void **stack, size_t depth, const ucontext_t *ctx)
  * that with a backtrace log.  Send messages via the tor_log function at
  * logger". */
 void
-log_backtrace_impl(int severity, int domain, const char *msg,
+log_backtrace_impl(int severity, uint64_t domain, const char *msg,
```
teor2345: Why aren't we using log_domain_mask_t here?
Does it break our may include rules?
If so, can you add a comment about keeping them in sync?

src/lib/err/backtrace.c
```diff
@@ -240,7 +240,7 @@ remove_bt_handler(void)
 
 #ifdef NO_BACKTRACE_IMPL
 void
-log_backtrace_impl(int severity, int domain, const char *msg,
+log_backtrace_impl(int severity, uint64_t domain, const char *msg,
```
teor2345: Same comment as above.

src/lib/err/backtrace.c
teor2345: Or can we put the type in a low-level header?

src/lib/err/backtrace.h
```diff
@@ -12,11 +12,12 @@
 
 #include "orconfig.h"
 #include "lib/cc/compat_compiler.h"
+#include "lib/cc/torint.h"
 
-typedef void (*tor_log_fn)(int, unsigned, const char *fmt, ...)
+typedef void (*tor_log_fn)(int, uint64_t, const char *fmt, ...)
```
teor2345: Same comment as above.

src/lib/err/backtrace.h
```diff
@@ -12,11 +12,12 @@
...
   CHECK_PRINTF(3,4);
 
-void log_backtrace_impl(int severity, int domain, const char *msg,
+void log_backtrace_impl(int severity, uint64_t domain, const char *msg,
```
teor2345: Same comment as above.

PR #1177, opened 2019-07-18
Practracker fixes

nmathewson:


scripts/maint/practracker/practracker.py
```diff
@@ -164,6 +165,8 @@ def main(argv):
     parser = argparse.ArgumentParser(prog=progname)
     parser.add_argument("--regen", action="store_true",
                         help="Regenerate the exceptions file")
+    parser.add_argument("--list-overstrict", action="store_true",
```
teor2345: I think this option should be called "list overbroad" or something.

Double negatives are hard, but I don't know how to eliminate this one. Here are some possible options:
* list excess(ive)
* list unnecessary
* list better (that is, we are doing better than our exceptions)

Makefile.am
```diff
@@ -369,7 +369,9 @@ endif
 
 check-best-practices:
 if USEPYTHON
-	$(PYTHON) $(top_srcdir)/scripts/maint/practracker/practracker.py $(top_srcdir)
+	@if test "$$TOR_DISABLE_PRACTRACKER" = ""; then \
```
teor2345: Can we implement this check in the practracker script instead?

The git hooks call practracker directly, not through the Makefile.

nmathewson: Sure.  Fixed in 3221dc1b32f651

scripts/maint/practracker/practracker.py
nmathewson: 19536fd18d02a5 fixes this one.

PR #1187, opened 2019-07-23
Bug30942

asn-d6:


src/core/or/relay.c
```diff
@@ -1607,6 +1609,16 @@ handle_relay_command(cell_t *cell, circuit_t *circ,
       if (circpad_handle_padding_negotiated(circ, cell, layer_hint) == 0)
         circuit_read_valid_data(TO_ORIGIN_CIRCUIT(circ), rh->length);
       return 0;
+  }
+
+  /* If this is a padding circuit we don't need to parse any other commands
+   * than the padding ones. Just drop them to the floor. */
+  if (circ->purpose == CIRCUIT_PURPOSE_C_CIRCUIT_PADDING) {
+    return 0;
+  }
```
dgoulet-tor: I'm thinking that everything that has to do with circpad could be in its own function. It seems a bit spread out right now with `circpad_deliver_recognized_relay_cell_events()` in another function and then the 3 possible relay command and then this check and then we go on to process the cell.

One single callsite that would return a code to continue or the value to return?

asn-d6: Tried to address this in 04d0ce7 .

PR #1192, opened 2019-07-30
[Tor Trac #25568]: Check if intro points are marked as failed in v2 INTRO cells

neelchauhan:
For the bug [here](https://trac.torproject.org/projects/tor/ticket/25568).

### DIFF
--- a/src/feature/rend/rendcache.c
+++ b/src/feature/rend/rendcache.c
@@ -242,7 +242,15 @@ rend_cache_failure_check(rend_service_descriptor_t *desc)
     return 0;
   }
 
-  return strmap_get_lc(rend_cache_failure, service_id) != NULL;
+  SMARTLIST_FOREACH_BEGIN(desc->intro_nodes, rend_intro_point_t *, intro) {
+    rend_cache_failure_intro_t *entry;
+    if (cache_failure_intro_lookup((const uint8_t *)
+                                   intro->extend_info->identity_digest,
+                                   service_id, &entry)) {
+      return 1;
+    }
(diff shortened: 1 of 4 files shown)


src/feature/rend/rendcache.c
```diff
@@ -228,6 +228,23 @@ rend_cache_entry_free_void(void *p)
...
+  char service_id[REND_SERVICE_ID_LEN_BASE32 + 1];
+
+  if (desc == NULL) {
+    return 0;
+  }
+  if (rend_get_service_id(desc->pk, service_id) < 0) {
+    return 0;
+  }
+
+  return strmap_get_lc(rend_cache_failure, service_id) != NULL;
```
dgoulet-tor: Please use `cache_failure_intro_lookup()` here instead of duplicating the lookup.

Also, we'll soon have a code standard policy to use bool for this type of functions. So using bool now is great step forward! :)

src/feature/rend/rendclient.c
```diff
@@ -1112,7 +1112,7 @@ rend_client_any_intro_points_usable(const rend_cache_entry_t *entry)
   extend_info_t *extend_info =
     rend_client_get_random_intro_impl(entry, get_options()->StrictNodes, 0);
 
-  int rv = (extend_info != NULL);
+  int rv = (extend_info != NULL) && !(rend_cache_failure_check(entry->parsed));
```
dgoulet-tor: `rend_client_get_random_intro_impl()` is a bit more CPU intensive so slight optimization would be to check the failure cache _before_ and not continue. If we can avoid work, lets do it.

PR #1209, opened 2019-08-06
Avoid using labs() on time_t in channeltls.c

nmathewson:
On some windows builds, time_t is 64 bits but long is not.  This is
causing appveyor builds to fail.

Also, one of our uses of labs() on time_t was logically incorrect:
it was telling us to accept NETINFO cells up to three minutes
_before_ the message they were responding to, which doesn't make
sense.

This patch adds a time_abs() function that we should eventually move
to intmath.h or something.  For now, though, it will make merges
easier to have it file-local in channeltls.c.

Fixes bug 31343; bugfix on 0.2.4.4-alpha.

### DIFF
--- a/changes/bug31343
+++ b/changes/bug31343
@@ -1,0 +1,9 @@
+  o Minor bugfixes (compilation):
+    - Avoid using labs() on time_t, which can cause compilation warnings
+      on 64-bit Windows builds.  Fixes bug 31343; bugfix on 0.2.4.4-alpha.
+
+  o Minor bugfixes (clock skew detection):
+    - Don't believe clock skew results from NETINFO cells that appear to
+      arrive before the VERSIONS cells they are responding to were sent.
+      Previously, we would accept them up to 3 minutes "in the past".
+      Fixes bug 31343; bugfix on 0.2.4.4-alpha.
--- a/src/or/channeltls.c
+++ b/src/or/channeltls.c
@@ -1721,7 +1721,7 @@ channel_tls_process_netinfo_cell(cell_t *cell, channel_tls_t *chan)
   /* Act on apparent skew. */
   /** Warn when we get a netinfo skew with at least this value. */
 #define NETINFO_NOTICE_SKEW 3600
-  if (time_abs(apparent_skew) &&
+  if (time_abs(apparent_skew) > NETINFO_NOTICE_SKEW &&
       router_get_by_id_digest(chan->conn->identity_digest)) {
     int trusted = router_digest_is_trusted_dir(chan->conn->identity_digest);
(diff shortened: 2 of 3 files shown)


src/or/channeltls.c
```diff
@@ -1583,6 +1583,18 @@ channel_tls_process_versions_cell(var_cell_t *cell, channel_tls_t *chan)
...
+/**
+ * Helper: compute the absolute value of a time_t.
+ *
+ * (we need this because labs() doesn't always work for time_t, since
+ * long can be shorter than time_t.)
+ */
+static inline time_t
+time_abs(time_t val)
+{
+  return (val < 0) ? -val : val;
```
teor2345: -TIME_T_MIN will overflow, but that's an issue in the existing code. (And very unlikely.)
Let's open a separate ticket in Tor: unspecified?

src/or/channeltls.c
```diff
@@ -1705,7 +1721,7 @@ channel_tls_process_netinfo_cell(cell_t *cell, channel_tls_t *chan)
   /* Act on apparent skew. */
   /** Warn when we get a netinfo skew with at least this value. */
 #define NETINFO_NOTICE_SKEW 3600
-  if (labs(apparent_skew) > NETINFO_NOTICE_SKEW &&
+  if (time_abs(apparent_skew) &&
```
teor2345: Deleting "> NETINFO_NOTICE_SKEW" is not documented in the changes file: we'll now warn on any skew, not just skews over 1 hour.

changes/bug31343
```diff
@@ -0,0 +1,9 @@
+  o Minor bugfixes (compilation):
+    - Avoid using labs() on time_t, which can cause compilation warnings
+      on 64-bit Windows builds.  Fixes bug 31343; bugfix on 0.2.4.4-alpha.
+
+  o Minor bugfixes (clock skew detection):
+    - Don't believe clock skew results from NETINFO cells that appear to
+      arrive before the VERSIONS cells they are responding to were sent.
+      Previously, we would accept them up to 3 minutes "in the past".
```
teor2345: Previously, we would calculate the skew up to 3 minutes in the past, but never warn on it, because we only warned on skews over 1 hour.

src/or/channeltls.c
nmathewson: Okay w me.

src/or/channeltls.c
nmathewson: Ug, that's a bug.

changes/bug31343
nmathewson: That's not quite right either -- there are two times here.  One is the skew, and one is the allowable delay in the netinfo cell.  It's possible to see a 1 hour skew even if the netinfo cell delay is negative.

PR #1218, opened 2019-08-08
Print summary at the end of the configure script

dgoulet-tor:


configure.ac
```diff
@@ -2519,3 +2519,118 @@ https://trac.torproject.org/projects/tor/wiki/doc/TorFragileHardening
...
+PPRINT_SET_TS(38)
+
+AS_ECHO
+AS_ECHO("${PPRINT_COLOR_BLDBLU}Tor $PACKAGE_VERSION $PPRINT_COLOR_RST")
+AS_ECHO
+
+PPRINT_SUBTITLE([Build Features])
+
+test "x$enable_gcc_warnings" != "xno" && value=1 || value=0
+PPRINT_PROP_BOOL([GCC Warnings], $value)
```
nmathewson: We should call this "verbose warnings"; it applies to clang too.

nmathewson: Maybe it would be a good idea to also print, for this setting and all the others, how to override it?

configure.ac
```diff
@@ -2519,3 +2519,118 @@ https://trac.torproject.org/projects/tor/wiki/doc/TorFragileHardening
...
+
+test "x$enable_fatal_warnings" != "xno" && value=1 || value=0
+PPRINT_PROP_BOOL([GCC Fatal Warnings], $value)
```
nmathewson: Let's call this "Warnings are fatal"

configure.ac
```diff
@@ -2519,3 +2519,118 @@ https://trac.torproject.org/projects/tor/wiki/doc/TorFragileHardening
...
+
+test "x$enable_coverage" = "xyes" && value=1 || value=0
+PPRINT_PROP_BOOL([Code Coverage], $value)
```
nmathewson: This maybe belongs in a testing section.

configure.ac
```diff
@@ -2519,3 +2519,118 @@ https://trac.torproject.org/projects/tor/wiki/doc/TorFragileHardening
...
+fi
+
+if test "x$enable_static_zlib" = "xyes"; then
+  PPRINT_PROP_STRING([zlib1g], [$TOR_LIBDIR_zlib])
+else
+  PPRINT_PROP_BOOL([zlib1g], $value)
+fi
+
+AS_ECHO
+PPRINT_SUBTITLE([Security])
```
nmathewson: Maybe we should have an "optional libraries" section for NSS and libscrypt and libsystemd.  This section makes it sould like your system is less secure if it doesn't have NSS, which isn't really true.

configure.ac
```diff
@@ -2519,3 +2519,118 @@ https://trac.torproject.org/projects/tor/wiki/doc/TorFragileHardening
...
+PPRINT_PROP_BOOL([libnss Support], $value)
+
+test "x$enable_seccomp" = "xyes" && value=1 || value=0
+PPRINT_PROP_BOOL([libseccomp Support], $value)
+
+test "x$enable_libscrypt" = "xyes" && value=1 || value=0
+PPRINT_PROP_BOOL([libscrypt support], $value)
+
+test "x$enable_fragile_hardening" != "xno" && value=1 || value=0
+PPRINT_PROP_BOOL([libasan support], $value)
```
nmathewson: Let's call this "fragile hardening" to match the option.  Also let's mark it as "not for production use" somehow-- people should use this in production.

configure.ac
```diff
@@ -2519,3 +2519,118 @@ https://trac.torproject.org/projects/tor/wiki/doc/TorFragileHardening
...
+
+test "x$enable_gcc_hardening" = "xyes" && value=1 || value=0
+PPRINT_PROP_BOOL([GCC Hardening], $value)
```
nmathewson: Let's call this "compiler hardening"? 

This option is supposed to be on by default, but when I ran the configure script, it said that it was off.  I think there might be a  bug.

configure.ac
```diff
@@ -2519,3 +2519,118 @@ https://trac.torproject.org/projects/tor/wiki/doc/TorFragileHardening
...
+PPRINT_PROP_BOOL([HTML Manual], $value)
+
+AS_ECHO
+PPRINT_SUBTITLE([Tests])
+
+test "x$enable_unittests" != "xno" && value=1 || value=0
+PPRINT_PROP_BOOL([Unit tests], $value)
+
+test "x$enable_asserts_in_tests" = "xyes" && value=1 || value=0
+PPRINT_PROP_BOOL([Unit tests assert()], $value)
```
nmathewson: This option was mis-reported when I ran the script.  It is almost always on.

Its name is a little confusing, too. How about invert this, can call it "assert()s disabled for unit tests"

We should also mark this option as not safe to use --disable-asserts-in-tests in production.

configure.ac
```diff
@@ -2519,3 +2519,118 @@ https://trac.torproject.org/projects/tor/wiki/doc/TorFragileHardening
...
+test "x$enable_libfuzzer" = "xyes" && value=1 || value=0
+PPRINT_PROP_BOOL([libFuzzer support], $value)
+
+test "x$enable_oss_fuzz" = "xyes" && value=1 || value=0
+PPRINT_PROP_BOOL([OSS-Fuzz support], $value)
+
+AS_ECHO
+PPRINT_SUBTITLE([Install Directories])
+
+report_mandir="`eval eval echo $mandir`"
```
nmathewson: eval eval?  I don't understand this.

configure.ac
```diff
@@ -2519,3 +2519,118 @@ https://trac.torproject.org/projects/tor/wiki/doc/TorFragileHardening
...
+PPRINT_PROP_STRING([Binaries], [$BINDIR])
+PPRINT_PROP_STRING([Configuration], [$CONFDIR])
+PPRINT_PROP_STRING([Man Pages], [$report_mandir])
```
nmathewson: Is there any way that we can get the complete configure command line as it was provided to us?  For a lot of cases, that would be more useful for debugging than all the rest of this stuff.

configure.ac
dgoulet-tor: Because just one eval gives: `${prefix}/share/man` so the second one evaluates that prefix.

configure.ac
dgoulet-tor: Yup. There is one fixup commit that fixes it all. I recommend you rerun the configure script to see the end result. I changed many things from your review :).

configure.ac
dgoulet-tor: +1

configure.ac
dgoulet-tor: +1

configure.ac
dgoulet-tor: +1

configure.ac
dgoulet-tor: +1

configure.ac
dgoulet-tor: +1

I fixed the bug I believe.

configure.ac
dgoulet-tor: +1

configure.ac
dgoulet-tor: I added a last line that does that :).

PR #1230, opened 2019-08-12
Bug31356+logs rebased2

mikeperry-tor:


src/core/or/or.h
```diff
@@ -841,8 +841,8 @@ typedef struct protover_summary_flags_t {
   unsigned int supports_v3_rendezvous_point: 1;
 
   /** True iff this router has a protocol list that allows clients to
-   * negotiate link-level padding. Requires Padding>=1. */
-  unsigned int supports_padding : 1;
+   * negotiate hs circuit setup padding. Requires Padding>=2. */
```
teor2345: This comment is a bit confusing.
When we remove Padding=2, 0.4.1 will stop negotiating padding, even if we have added Padding=3.

I think this is a decent fix, but I'd be happy with any clarification:
```suggestion
   * negotiate hs circuit setup padding. Requires Padding=2-*
   * Padding=2 will always be used for HS circuit setup.
   * Future padding versions will introduce support for other
   * machines.
   */
```

PR #1232, opened 2019-08-13
Ticket30924 042 01

dgoulet-tor:


src/feature/hs/hs_intropoint.c
```diff
@@ -181,6 +181,108 @@ hs_intro_send_intro_established_cell,(or_circuit_t *circ))
...
+
+  tor_assert(parsed_cell);
+  tor_assert(circ);
+
+  extensions = trn_cell_establish_intro_getconst_extensions(parsed_cell);
+  if (extensions == NULL) {
+    goto end;
+  }
+
+  /* Go over all extensions. */
```
asn-d6: You mentioned that this `N_EXTENSIONS` `N_PARAMS` inception situation is confusing and I can see why this is that. Since we basically have:
```
     AUTH_KEY_TYPE         [1 byte]
     AUTH_KEY_LEN          [2 bytes]
     AUTH_KEY              [AUTH_KEY_LEN bytes]
     N_EXTENSIONS          [1 byte]
     N_EXTENSIONS times:
        EXT_FIELD_TYPE     [1 byte]
        EXT_FIELD_LEN      [1 byte]
        N_PARAMS    [1 byte]
        N_PARAMS times:
           PARAM_TYPE  [1 byte]
           PARAM_VALUE [8 byte]
     HANDSHAKE_AUTH        [MAC_LEN bytes]
     SIG_LEN               [2 bytes]
     SIG                   [SIG_LEN bytes]
```

Just for the sake of argument and sorry for suggesting it late in the process, but what would happen if we ditched the second layer of abstraction and encoded rate and burst directly as `N_EXTENSIONS` instead of `N_PARAMS`?

src/feature/hs/hs_intropoint.c
```diff
@@ -181,6 +181,108 @@ hs_intro_send_intro_established_cell,(or_circuit_t *circ))
...
+      break;
+    case TRUNNEL_DOS_PARAM_TYPE_INTRO2_BURST_PER_SEC:
+      intro2_burst_per_sec = trn_cell_extension_dos_param_get_value(param);
+      break;
+    default:
+      goto end;
+    }
+  }
+
+  /* Validation. A value of 0 on either of them means the defenses are
```
asn-d6: It would be really nice if this value validation was the same code that does the torrc validation, to minimize the desynch that can happen between torrc and intro point validation here (so that we only have possible desynch between tor versions). Do you think this is possible?

src/feature/hs/hs_intropoint.c
```diff
@@ -181,6 +181,108 @@ hs_intro_send_intro_established_cell,(or_circuit_t *circ))
...
+  if ((intro2_rate_per_sec > INT32_MAX || intro2_rate_per_sec == 0) ||
+      (intro2_burst_per_sec > INT32_MAX || intro2_burst_per_sec == 0) ||
+      (intro2_burst_per_sec < intro2_rate_per_sec)) {
+    circ->introduce2_dos_defense_enabled = 0;
+    log_info(LD_REND, "Intro point DoS defenses disabled due to bad values");
+  } else {
+    circ->introduce2_dos_defense_enabled = 1;
+
+    /* Initialize the INTRODUCE2 token bucket for the rate limiting. */
+    token_bucket_ctr_init(&circ->introduce2_bucket,
```
asn-d6: We are re-initing the bucket here. Is this OK? We should document this somewhere.

src/feature/hs/hs_cell.c
```diff
@@ -473,26 +473,125 @@ introduce1_set_legacy_id(trn_cell_introduce1_t *cell,
...
+ * extensions pointer is always set to a valid cell extension object. */
+static void
+build_establish_intro_extensions(const hs_service_config_t *service_config,
+                                 const hs_service_intro_point_t *ip,
+                                 trn_cell_extension_t **extensions)
+{
+  tor_assert(ip);
+  tor_assert(extensions);
+
+  *extensions = trn_cell_extension_new();
```
asn-d6: Instead of passing `**extensions` to this func, let's just create it and return it from this func. This will simplify things IMO.

src/feature/hs/hs_dos.c
```diff
@@ -156,8 +156,10 @@ hs_dos_can_send_intro2(or_circuit_t *s_intro_circ)
 {
   tor_assert(s_intro_circ);
 
-  /* Always allowed if the defense is disabled. */
```
asn-d6: Hmm this commit was confusing to me. I need to look into this more. But perhaps we can make `param_introduce_defense_enabled` private so that it's obvious that it should not be used by public functions like this one?

src/app/config/config.c
```diff
@@ -484,6 +484,11 @@ static config_var_t option_vars_[] = {
   VAR("HiddenServiceMaxStreamsCloseCircuit",LINELIST_S, RendConfigLines, NULL),
   VAR("HiddenServiceNumIntroductionPoints", LINELIST_S, RendConfigLines, NULL),
   VAR("HiddenServiceExportCircuitID", LINELIST_S,  RendConfigLines, NULL),
+  VAR("HiddenServiceEnableIntroDoSDefense", LINELIST_S, RendConfigLines, NULL),
```
asn-d6: We need man page entries for this,

src/feature/hs/hs_intropoint.c
dgoulet-tor: Not sure we can. The `N_EXTENSIONS` contains an extension itself. In this case, it contains the "INTRO2 DoS Extension" encoded data.

Which in turn has this `N_PARAMS` design to list the parameters. There could be other extensions in the future that have nothing to do with this DoS one.

src/feature/hs/hs_intropoint.c
dgoulet-tor: Hmm... possibly if I modularize the check in `hs_config.c` and expose the functions publicly so `hs_intropoint.c` could go for it.

Not a bad idea at all!

src/feature/hs/hs_intropoint.c
dgoulet-tor: Yes it is fine. It just resets all the counter but also changes the configuration to the new rate/burst values.

src/feature/hs/hs_cell.c
dgoulet-tor: Sounds good.

src/feature/hs/hs_dos.c
dgoulet-tor: `param_introduce_defense_enabled` needs to be public from a function. I believe one of the last commits makes it public so the intro point sets the service intro circuit with that value by default to know if it enables the defenses or not.

See `hs_dos_get_intro2_enabled_param()`

src/app/config/config.c
dgoulet-tor: Yesss and changes file!

src/feature/hs/hs_intropoint.c
asn-d6: Hm, let me give you a visual. Here is what prop305 is doing:
```
     AUTH_KEY_TYPE                          [1 byte]
     AUTH_KEY_LEN                           [2 bytes]
     AUTH_KEY                               [AUTH_KEY_LEN bytes]
     N_EXTENSIONS (1)                       [1 byte]
        EXT_FIELD_TYPE (DOS_PARAMETERS)     [1 byte]
        EXT_FIELD_LEN                       [1 byte]
        N_PARAMS (2)                        [1 byte]
        N_PARAMS times:
           PARAM_TYPE (DOS_INTRODUCE2_RATE_PER_SEC)    [1 byte]
           PARAM_VALUE (25)                      [8 byte]
           PARAM_TYPE (DOS_INTRODUCE2_BURST_PER_SEC)    [1 byte]
           PARAM_VALUE (250)                      [8 byte]
     HANDSHAKE_AUTH        [MAC_LEN bytes]
     SIG_LEN               [2 bytes]
     SIG                   [SIG_LEN bytes]
```

I'm wondering if we can do:
```
     AUTH_KEY_TYPE                          [1 byte]
     AUTH_KEY_LEN                           [2 bytes]
     AUTH_KEY                               [AUTH_KEY_LEN bytes]
     N_EXTENSIONS (2)                       [1 byte]
        EXT_FIELD_TYPE (DOS_INTRODUCE2_RATE_PER_SEC)    [1 byte]
        EXT_FIELD_VALUE (25)                      [8 byte]
        EXT_FIELD_TYPE (DOS_INTRODUCE2_BURST_PER_SEC)    [1 byte]
        EXT_FIELD_VALUE (250)                      [8 byte]
     HANDSHAKE_AUTH        [MAC_LEN bytes]
     SIG_LEN               [2 bytes]
     SIG                   [SIG_LEN bytes]
```

dgoulet-tor: Right. No we can not :).

Because the `EXT_FIELD_TYPE` is describing an *extension* type, can't be used to host the actual extension data.

Furthermore, we need to specify the `length` of the extension which is put in the `EXT_FIELD_LEN`. And also, considering that a ESTABLISH_INTRO cell could have many more different extensions, we need a way to specify which extension you are looking at when iterating over all extensions.

In our case, we are adding 1 new extension that is the "DoS defenses" so `N_EXTENSIONS` has to be 1.

This extension concept is spec out in all our HS cells so we can't move away from it.

src/feature/hs/hs_intropoint.c
dgoulet-tor: Oh wait now I remember why it is separated... It is because with the torrc options, we parse strings and use helper functions for that that do check for boundaries:

See `helper_parse_uint64()` ... so we pass the default values to that function and we get a boundary checks.

I'll at least use the `#define` that exists for the default values of the DoS defenses in `hs_intropoint.c`.

src/feature/hs/hs_intropoint.c
asn-d6: After discussion on IRC, this seems like a possible thing, but we should not do it because it breaks the semantics of extensions.

src/feature/hs/hs_cell.c
asn-d6: LGTM.

src/app/config/config.c
asn-d6: LGTM.

src/feature/hs/hs_intropoint.c
asn-d6: ACK!

src/feature/hs/hs_dos.c
asn-d6: Hmm, I'm still confused by this. Here are some reasons:

- I don't like the name of `hs_dos_get_intro2_rate_param()` and its two friends. The name I don't like because "param" is an ambiguous word in general, but even more so in this particular context given that the code means "consensus param" but the first thing that came to my mind were the new params from the ESTABLISH_INTRO cell. 

- I don't like that `hs_dos_get_intro2_enabled_param()` is public since it's not clear whether that function knows the ground truth, or I need to consult the `s_intro_circ->introduce2_dos_defense_enabled` to know the ground truth.

- It's a bit confusing to have `get_param_intro_dos_enabled()` and also `hs_dos_get_intro2_rate_param()` and neither of these two things gives me the ground truth in some cases (because if the values are specified in the establish_intro cell the ground truth is different).

Here is a suggestion. Please tell me if you like it:

- Dump the `get_param_intro_dos_enabled` etc. functions and do the consensus check inline in `set_consensus_parameters()`. This way we dont have two similarly named functions doing almost the same thing.

- Abstract the following code of `hs_intropoint.c` into its own `hs_dos_setup_default_introduce2_defences()` `hs_dos` function:


  ```
  circ->introduce2_dos_defense_enabled = hs_dos_get_intro2_enabled_param();
  token_bucket_ctr_init(&circ->introduce2_bucket,
                        hs_dos_get_intro2_rate_param(),
                        hs_dos_get_intro2_burst_param(),
                        (uint32_t) approx_time());
  ```

  This way we can then make the `hs_dos_get_intro2_` functions into private for the hs_dos module.

- Finally, rename the `hs_dos_get_intro2_rate_param()` functions into `get_intro2_rate_default_consensus_value()` because now they are private.

I can also try to do this if you want me , and then we can see if we like it.

doc/tor.1.txt
```diff
@@ -3025,6 +3025,38 @@ The following options are used to configure a hidden service.
...
+[[HiddenServiceEnableIntroDoSDefense]] **HiddenServiceEnableIntroDoSDefense** **0**|**1**::
+    Enable DoS defense at the intropoint level. When this is enabled, the
+    rate and burst parameter (see below) will be sent to the intro point which
+    will then use them to apply rate limiting for introduction request to this
+    service.
+  +
+    The introduction point honors the consensus parameters except if this is
+    specifically set by the service operator using this option. The service
+    never looks at the consensus parameters in order to enable or disable this
+    defense. (Default: 0)
```
asn-d6: What happens if `HiddenServiceEnableIntroDoSDefense` is enabled but no rate/burst values are set in the torrc? I was assuming that the default rate/burst values are picked (as documented below) but I think that's not the case, and the defense stays disabled. Is this true? Maybe this should be documented on the man page, or add a warn in the code?

doc/tor.1.txt
```diff
@@ -3025,6 +3025,38 @@ The following options are used to configure a hidden service.
...
+
+[[HiddenServiceEnableIntroDoSRatePerSec]] **HiddenServiceEnableIntroDoSRatePerSec** __NUM__::
+    The allowed client introduction rate per second at the introduction
+    point. If this option is 0, it is consider infinite and thus if
+    **HiddenServiceEnableIntroDoSDefense** is set, it then effectively
+    disables the defenses. (Default: 25)
```
asn-d6: s/consider/considered ?

src/feature/hs/hs_cell.c
```diff
@@ -473,26 +473,128 @@ introduce1_set_legacy_id(trn_cell_introduce1_t *cell,
...
+build_establish_intro_dos_param(trn_cell_extension_dos_t *dos_ext,
+                                uint8_t param_type, uint64_t param_value)
+{
+  trn_cell_extension_dos_param_t *dos_param =
+    trn_cell_extension_dos_param_new();
+
+  /* Extra safety. We should never send an unknown parameter type. */
+  tor_assert(param_type == TRUNNEL_DOS_PARAM_TYPE_INTRO2_RATE_PER_SEC ||
+             param_type == TRUNNEL_DOS_PARAM_TYPE_INTRO2_BURST_PER_SEC);
+
```
asn-d6: Let's add a `log_info` here with the values we chose, because there is no service-side log about this I think. We can have it `log_info` since building intro circuits is not so frequent.

src/feature/hs/hs_config.h
```diff
@@ -15,6 +15,14 @@
 #define HS_CONFIG_MAX_STREAMS_PER_RDV_CIRCUIT 65535
 /* Maximum number of intro points per version 3 services. */
 #define HS_CONFIG_V3_MAX_INTRO_POINTS 20
+/* Default value for the introduction DoS defenses. */
```
asn-d6: Let's document all of these and specify how each of them is used. For validation or as default values? By the service or the intro point?

src/feature/hs/hs_dos.c
```diff
@@ -46,11 +46,11 @@
 #define HS_DOS_INTRODUCE_ENABLED_DEFAULT 0
 
 /* Consensus parameters. */
```
asn-d6: Perhaps we should document that if there is a value in establish_intro cell these values get lower priority over the cell values?

src/feature/hs/hs_intropoint.c
```diff
@@ -181,6 +182,110 @@ hs_intro_send_intro_established_cell,(or_circuit_t *circ))
...
+    case TRUNNEL_DOS_PARAM_TYPE_INTRO2_BURST_PER_SEC:
+      intro2_burst_per_sec = trn_cell_extension_dos_param_get_value(param);
+      break;
+    default:
+      goto end;
+    }
+  }
+
+  /* Validation. A value of 0 on either of them means the defenses are
+   * disabled so we ignore. */
```
asn-d6: Let's split these three checks into individual `if` blocks and document each of them on what exactly they are doing, so that it's more clear what we are checking exactly.

src/feature/hs/hs_intropoint.c
```diff
@@ -181,6 +182,110 @@ hs_intro_send_intro_established_cell,(or_circuit_t *circ))
...
+  if ((intro2_rate_per_sec > HS_CONFIG_V3_DOS_DEFENSE_BURST_PER_SEC_MAX ||
+       intro2_rate_per_sec <= HS_CONFIG_V3_DOS_DEFENSE_RATE_PER_SEC_MIN) ||
+      (intro2_burst_per_sec > HS_CONFIG_V3_DOS_DEFENSE_RATE_PER_SEC_MAX ||
```
asn-d6: Why isn't `intro2_burst_per_sec > HS_CONFIG_V3_DOS_DEFENSE_RATE_PER_SEC_MAX` allowed? Is there an implicit limit to burst values based on the rate values? As suggested above some documentation on these checks would be useful.

src/feature/hs/hs_intropoint.c
```diff
@@ -181,6 +182,110 @@ hs_intro_send_intro_established_cell,(or_circuit_t *circ))
...
+       intro2_burst_per_sec <= HS_CONFIG_V3_DOS_DEFENSE_BURST_PER_SEC_MIN) ||
+      (intro2_burst_per_sec < intro2_rate_per_sec)) {
+    circ->introduce2_dos_defense_enabled = 0;
+    log_info(LD_REND, "Intro point DoS defenses disabled due to bad values");
```
asn-d6: Let's print the offending values in this case also.

src/feature/hs/hs_config.c
```diff
@@ -334,6 +348,48 @@ config_service_v3(const config_line_t *line_,
       export_circuit_id = true;
       continue;
     }
+    if (!strcasecmp(line->key, "HiddenServiceEnableIntroDoSDefense")) {
```
asn-d6: Perhaps we need some unittests for the torrc parsing and validation? And maybe cross-checking them with the intro-point validation checks?

src/feature/hs/hs_dos.c
dgoulet-tor: Ok I've pushed several new commits to address this. Not very complicated. I did not follow exactly what you said, made some changes, you'll see. Main thing is that I kept the consensus parameter variables (see start of `hs_dos.c`) but now I only use those in the hs_dos.c code.

Last commit is also a fix that I found while doing this refactoring.

doc/tor.1.txt
dgoulet-tor: Default rate and burst are used. See `set_service_default_config()`.

doc/tor.1.txt
dgoulet-tor: Fixup commit done.

src/feature/hs/hs_cell.c
dgoulet-tor: Commit 5e256d58c7]

src/feature/hs/hs_config.h
dgoulet-tor: Well hmmm... you sure that is necessary? We have man page explaining each of them. We have a full proposal explaining what they do.

And there is `_DEFAULT` for each param for the _exact_ purpose of indicating the default value? And MIN/MAX is obviously for validation.

Not sure what more would be useful here without duplicating quite a bit the same documentation?

src/feature/hs/hs_dos.c
dgoulet-tor: Commit 9eeb2a53a3

src/feature/hs/hs_intropoint.c
dgoulet-tor: I've majorly refactor this. And fixed a bug in there! There was a typo between RATE/BURST... And now even unit testing that ;)

Commit 24c49a51f2

src/feature/hs/hs_intropoint.c
dgoulet-tor: This is one of the bug I fixed in previous commit.

src/feature/hs/hs_intropoint.c
dgoulet-tor: Done in the refactored commit above.

src/feature/hs/hs_config.c
dgoulet-tor: See commit 667e6db249

PR #1236, opened 2019-08-19
Rust: require at least rustc version 1.28

shahn:
The #[global_allocator] attribute is not available in previous versions


shahn:
I am done playing with CI, think it is ready

PR #1243, opened 2019-08-20
Ticket30924 042 02

dgoulet-tor:


src/feature/hs/hs_intropoint.c
```diff
@@ -182,6 +182,60 @@ hs_intro_send_intro_established_cell,(or_circuit_t *circ))
...
+STATIC bool
+validate_cell_dos_extension_parameters(uint64_t intro2_rate_per_sec,
+                                       uint64_t intro2_burst_per_sec)
+{
+  bool ret = false;
+
+  /* Bound checking the received rate per second. As per proposal 305, a
+   * value of 0 disabled the defenses regardless of the other parameters in
+   * the extension which is why we stop right away. */
+  if (intro2_rate_per_sec > HS_CONFIG_V3_DOS_DEFENSE_RATE_PER_SEC_MAX ||
```
asn-d6: Hmm, I'm a bit confused with the min max inclusions here. Usually I think min is inclusive whereas max is exclusive, but I think the comparison above is doing the opposite.

Like if we say, "we accept values from 15 to 30", we usually mean that 15 is accepted but 30 is not. I think the above check does not agree with that. Is that true? Is that intended? Should we clarify that in the spec?

src/feature/hs/hs_intropoint.c
```diff
@@ -182,6 +182,60 @@ hs_intro_send_intro_established_cell,(or_circuit_t *circ))
...
+      intro2_rate_per_sec <= HS_CONFIG_V3_DOS_DEFENSE_RATE_PER_SEC_MIN) {
+    if (intro2_rate_per_sec != 0) {
```
asn-d6: What's the point of this explicit `intro2_rate_per_sec != 0` check? Can't we just do the log even if it's 0?

src/feature/hs/hs_intropoint.c
asn-d6: Since `#define HS_CONFIG_V3_DOS_DEFENSE_RATE_PER_SEC_MIN 0` doesn't this mean that 0 is not accepted  as a valid value? But isn't 0 the right way to disable the defence?

src/feature/hs/hs_intropoint.c
dgoulet-tor: 0 is valid value. It means "disable defense" per prop305. So instead of logging "bad value", we just skip it.

src/feature/hs/hs_intropoint.c
dgoulet-tor: Hmmm the `<=` with MIN was intended to catch 0 because in this case MIN is 0 but if it ever goes to let say 10, we want to catch 0 as something that disables the defenses.

I've pushed a fixup to make this whole thing a bit better: 057fa01718399e2d and bba9b54fb1b93e64

PR #1244, opened 2019-08-20
Ticket30914 merged

nmathewson:


src/feature/nodelist/routerset.c
```diff
@@ -461,3 +465,61 @@ routerset_free_(routerset_t *routerset)
...
+  routerset_t **p = (routerset_t**)value;
+  routerset_free(*p); // sets *p to NULL.
+}
+
+static int
+routerset_copy(void *dest, const void *src, const void *params)
+{
+  (void)params;
+  routerset_t **output = (routerset_t**)dest;
+  const routerset_t *input = *(routerset_t**)src;
```
teor2345: This type-casting is error-prone, but I think we fix that later in these branches.

src/lib/conf/conftypes.h
```diff
@@ -63,8 +63,48 @@ typedef enum config_type_t {
   CONFIG_TYPE_ROUTERSET,    /**< A list of router names, addrs, and fps,
                              * parsed into a routerset_t. */
   CONFIG_TYPE_OBSOLETE,     /**< Obsolete (ignored) option. */
+  CONFIG_TYPE_EXTENDED,     /**< Extended type; definition will appear in
+                             * pointer. */
```
teor2345: This documentation is unclear: which pointer, and what kind of pointer?

teor2345:
```suggestion
                             * the type_def pointer. */
```

PR #1265, opened 2019-08-26
Pr1182 squashed

nmathewson:


src/test/test_address.c
```diff
@@ -1269,12 +1269,13 @@ static void
 test_address_dirserv_router_addr_private(void *ignored)
 {
   (void)ignored;
+  routerinfo_t *ri = NULL; /* Not really a routerinfo. */
```
teor2345: This comment is confusing. The variable is a routerinfo_t.

Do you mean:
"A routerinfo that only has an address. All other fields are zeroed."

PR #1274, opened 2019-08-29
hs-v3: Make service pick the exact amount of intro points

dgoulet-tor:
When encoding introduction points, we were not checking if that intro points
had an established circuit.

When botting up, the service will pick, by default, 3 + 2 intro points and the
first 3 that establish, we use them and upload the descriptor.

However, the intro point is removed from the service descriptor list only when
the circuit has opened and we see that we have already enough intro points, it
is then removed.

But it is possible that the service establishes 3 intro points successfully
before the other(s) have even opened yet.

This lead to the service encoding extra intro points in the descriptor even
though the circuit is not opened or might never establish (#31561).

Fixes #31548

### DIFF
--- a/src/feature/hs/hs_service.c
+++ b/src/feature/hs/hs_service.c
@@ -1686,6 +1686,15 @@ build_desc_intro_points(const hs_service_t *service,
 
   DIGEST256MAP_FOREACH(desc->intro_points.map, key,
                        const hs_service_intro_point_t *, ip) {
+    if (!ip->circuit_established) {
+      /* Ignore un-established intro points. They can linger in that list
+       * because their circuit has not opened and they haven't been removed
+       * yet even though we have enough intro circuits.
+       *
+       * XXX: Due to #31561, it can stay in that list until rotation so this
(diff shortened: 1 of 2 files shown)


src/feature/hs/hs_service.c
```diff
@@ -1686,6 +1686,15 @@ build_desc_intro_points(const hs_service_t *service,
 
   DIGEST256MAP_FOREACH(desc->intro_points.map, key,
                        const hs_service_intro_point_t *, ip) {
+    if (!ip->circuit_established) {
+      /* Ignore un-established intro points. They can linger in that list
+       * because their circuit has not opened and they haven't been removed
+       * yet even though we have enough intro circuits.
+       *
+       * XXX: Due to #31561, it can stay in that list until rotation so this
```
asn-d6: I don't think it makes sense to have this alarming XXX in master where we will also merge #31561. Should we have two branches, one with the XXX for maint that don't include #31561 and one without for master? Or maybe just remove the XXX entirely? Up to you.

PR #1275, opened 2019-08-29
hs-v3: Remove intro point on circuit build time out

dgoulet-tor:
When establishing service intro points, if a circuit ever times out due to
CBT, callback into the HS subsystem in order to remove the intro point from
the service descriptor list.

Fixese #31561

### DIFF
--- a/src/core/or/circuituse.c
+++ b/src/core/or/circuituse.c
@@ -840,10 +840,15 @@ circuit_expire_building(void)
                  -1);
 
     circuit_log_path(LOG_INFO,LD_CIRC,TO_ORIGIN_CIRCUIT(victim));
-    if (victim->purpose == CIRCUIT_PURPOSE_C_MEASURE_TIMEOUT)
+    if (victim->purpose == CIRCUIT_PURPOSE_C_MEASURE_TIMEOUT) {
       circuit_mark_for_close(victim, END_CIRC_REASON_MEASUREMENT_EXPIRED);
-    else
+    } else {
       circuit_mark_for_close(victim, END_CIRC_REASON_TIMEOUT);
+      if (victim->purpose >= CIRCUIT_PURPOSE_S_HS_MIN_ &&
+          victim->purpose <= CIRCUIT_PURPOSE_S_HS_MAX_) {
+        hs_service_circuit_timed_out(TO_ORIGIN_CIRCUIT(victim));
+      }
+    }
 
     pathbias_count_timeout(TO_ORIGIN_CIRCUIT(victim));
   } SMARTLIST_FOREACH_END(victim);
--- a/src/core/or/circuituse.c
+++ b/src/core/or/circuituse.c
@@ -96,6 +96,10 @@ circuit_purpose_is_hs_service(uint8_t purpose)
 static int
 circuit_purpose_is_hs_client(uint8_t purpose)
 {
+  if (purpose == CIRCUIT_PURPOSE_HS_VANGUARDS) {
+    return 1;
+  }
+
   /* Client-side purpose */
   return (purpose >= CIRCUIT_PURPOSE_C_HS_MIN_ &&
           purpose <= CIRCUIT_PURPOSE_C_HS_MAX_);
--- a/src/core/or/circuituse.c
+++ b/src/core/or/circuituse.c
@@ -73,6 +73,34 @@
 STATIC void circuit_expire_old_circuits_clientside(void);
 static void circuit_increment_failure_count(void);
 
+/* Return true iff the circuit purpose is for an onion service excluding
+ * client, only service side. */
+static int
+circuit_purpose_is_hs_service(uint8_t purpose)
+{
+  if (purpose == CIRCUIT_PURPOSE_HS_VANGUARDS) {
+    return 1;
+  }
+
+  /* Service-side purpose */
+  if (purpose >= CIRCUIT_PURPOSE_S_HS_MIN_ &&
+      purpose <= CIRCUIT_PURPOSE_S_HS_MAX_) {
+    return 1;
+  }
+
+  return 0;
+}
+
+/* Return true iff the circuit purpose is for an onion service client
+ * excluding service side. */
+static int
+circuit_purpose_is_hs_client(uint8_t purpose)
+{
+  /* Client-side purpose */
+  return (purpose >= CIRCUIT_PURPOSE_C_HS_MIN_ &&
+          purpose <= CIRCUIT_PURPOSE_C_HS_MAX_);
+}
+
 /** Check whether the hidden service destination of the stream at
  *  <b>edge_conn</b> is the same as the destination of the circuit at
  *  <b>origin_circ</b>. */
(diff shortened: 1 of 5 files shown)


src/feature/hs/hs_service.c
```diff
@@ -3404,6 +3404,29 @@ service_encode_descriptor(const hs_service_t *service,
...
+ * established introduction cell. */
+static void
+service_intro_circuit_timed_out(const origin_circuit_t *circ)
+{
+  hs_service_t *service = NULL;
+  hs_service_intro_point_t *ip = NULL;
+
+  tor_assert(circ);
+  tor_assert(TO_CIRCUIT(circ)->purpose == CIRCUIT_PURPOSE_S_ESTABLISH_INTRO ||
+             TO_CIRCUIT(circ)->purpose == CIRCUIT_PURPOSE_S_INTRO);
```
asn-d6: Hmm, `CIRCUIT_PURPOSE_S_INTRO` should not be possible according to the caller. Is there a reason for having it here?

src/feature/hs/hs_service.c
```diff
@@ -3404,6 +3404,29 @@ service_encode_descriptor(const hs_service_t *service,
...
+/* An introduction circuit has timed out while building or waiting for an
+ * established introduction cell. */
+static void
+service_intro_circuit_timed_out(const origin_circuit_t *circ)
+{
+  hs_service_t *service = NULL;
+  hs_service_intro_point_t *ip = NULL;
+
+  tor_assert(circ);
+  tor_assert(TO_CIRCUIT(circ)->purpose == CIRCUIT_PURPOSE_S_ESTABLISH_INTRO ||
```
asn-d6: Let's ditch the fatal assert just in case something changes in the future?

src/core/or/circuituse.c
```diff
@@ -840,10 +840,15 @@ circuit_expire_building(void)
                  -1);
 
     circuit_log_path(LOG_INFO,LD_CIRC,TO_ORIGIN_CIRCUIT(victim));
-    if (victim->purpose == CIRCUIT_PURPOSE_C_MEASURE_TIMEOUT)
+    if (victim->purpose == CIRCUIT_PURPOSE_C_MEASURE_TIMEOUT) {
       circuit_mark_for_close(victim, END_CIRC_REASON_MEASUREMENT_EXPIRED);
-    else
+    } else {
       circuit_mark_for_close(victim, END_CIRC_REASON_TIMEOUT);
+      if (victim->purpose >= CIRCUIT_PURPOSE_S_HS_MIN_ &&
```
asn-d6: Given that `hs_service_circuit_timed_out()` is pseudo-generic I think we might want to include `CIRCUIT_PURPOSE_HS_VANGUARDS` here, like `circuit_purpose_is_hidden_service()` does. Perhaps we can make a new function for this check, or add an `include_client` argument to `circuit_purpose_is_hidden_service()`? No strong opinions here.

src/feature/hs/hs_service.c
```diff
@@ -3404,6 +3404,29 @@ service_encode_descriptor(const hs_service_t *service,
   return ret;
 }
 
+/* An introduction circuit has timed out while building or waiting for an
+ * established introduction cell. */
```
asn-d6: Can you please explain a bit more in this comment why we need to handle the timeout case specially? Don't we already have code that removes expired intro circuits from the circuitmap etc.? Why do we need more handling code for this?

src/core/or/circuituse.c
dgoulet-tor: Oh good idea. Forgot about the existence of `circuit_purpose_is_hidden_service()`

Fixup: 986e82279d

src/feature/hs/hs_service.c
dgoulet-tor: Hmmm well that is the whole purpose of that assert. If we change something later that sends something else than an introduction circuit, we can catch it _early_. Because of the switch case in `hs_service_circuit_timed_out()`, it should never happen so this assert would be hit instantly by our unit testing.

src/feature/hs/hs_service.c
dgoulet-tor: Removed in fixup: ce79052eab

src/feature/hs/hs_service.c
dgoulet-tor: Added in fixup commit: ce79052eab

src/feature/hs/hs_service.c
dgoulet-tor: I've also added a non fatal assert on `circuit_established` so we could catch more bugs if any.

src/feature/hs/hs_service.c
asn-d6: Hm, I don't see it in the fixup.

src/core/or/circuituse.c
asn-d6: I'm not sure if that's the right function to use since it also includes client-side onion service circuits. That's why I suggested the modifications above.

src/feature/hs/hs_service.c
asn-d6: LGTM!

src/core/or/circuituse.c
dgoulet-tor: Oh damn... good catch. No it shouldn't be for client in this case at all. This is for "service" only.

I've removed entirely that fixup ^ and added a new fresh commit that splits up the `is_hidden_service()` into client and service specific functions. WAY LESS confusing that way imo.

See commit: 7eedb90850

src/feature/hs/hs_service.c
dgoulet-tor: Oh my mistake. Fixup of the fixup: a05f3dd02d

src/test/test_hs_service.c
```diff
@@ -2172,6 +2172,50 @@ test_export_client_circuit_id(void *arg)
   tor_free(cp2);
 }
 
+static void
+test_circ_timeout(void *arg)
```
asn-d6: Hmm the unittest failed on clang.

src/core/or/circuituse.c
```diff
@@ -73,6 +73,34 @@
...
+
+  return 0;
+}
+
+/* Return true iff the circuit purpose is for an onion service client
+ * excluding service side. */
+static int
+circuit_purpose_is_hs_client(uint8_t purpose)
+{
+  /* Client-side purpose */
```
asn-d6: Hmm, the split is not entirely true because vanguards also can be used for clients. There is no error introduced by this commit, but there is a semantic issue. I don't have a strong opinion on how to resolve this.

dgoulet-tor: What should I do here hehe? I'm not sure. The `circuit_purpose_is_hs_service()` does look for vanguard also. If `is_hs_client()` can also return true for the Vanguard purpose, we have a problem since both would return true?...

Want me to add a comment saying <insert something> ?

src/test/test_hs_service.c
dgoulet-tor: I believe I've fixed that with fixup: 28ce878609730cc143a8ea3ef0f012f34cdbc0f4

src/core/or/circuituse.c
asn-d6: Hmm, how about we return True in `is_hs_clietn()` for vanguard circuits, so that in the future if someone uses it, we have taken it into account (and maybe also add it in the func docs)?

Do you think there will be a problem because they both would return true? Why?

dgoulet-tor: Unsure if I think there is a problem. I guess if the circuit is an HS `Vanguard` purpose, then it doesn't matter if both return True since all we want to know is if "it is HS" since these function are primarily used in `circuit_purpose_is_hidden_service()`.

Fixup commit has the fix: ec801f4af2ae14f1

PR #1276, opened 2019-08-29
Ticket31549

nmathewson:


src/feature/dirauth/process_descs.c
```diff
@@ -315,6 +315,41 @@ dirserv_would_reject_router(const routerstatus_t *rs)
...
+  if (!tor_version_as_new_as(platform,"0.2.9.5-alpha")) {
+    if (msg)
+      *msg = "Tor version is insecure or unsupported. Please upgrade!";
+    return true;
+  }
+
+  /* Series between Tor 0.3.0 and 0.3.4 inclusive are unsupported, and some
+   * have bug #27841, which makes them broken as intro points. Reject them. */
+  if (tor_version_as_new_as(platform,"0.3.0.0-alpha-dev") &&
+      !tor_version_as_new_as(platform,"0.3.5.1-alpha")) {
```
dgoulet-tor: Could we go up to the latest stable here? Here is the 035 count:
```
   4: 0.3.5.3-alpha     [0.10 %] 
   3: 0.3.5.6-rc        [0.07 %] 
 207: 0.3.5.7           [5.14 %] 
1442: 0.3.5.8           [20.98 %] 
```

Removing -alpha and -rc would remove 7 relays and 0.17% of weight.

For 0.3.5.7, it has TROVE-2019-001 but it represents 5.14% of the bw of the network so maybe enough to justify keeping it in for now.

src/feature/dirauth/process_descs.c
```diff
@@ -315,6 +315,41 @@ dirserv_would_reject_router(const routerstatus_t *rs)
...
+    if (msg) {
+      *msg = "Tor version is insecure or unsupported. Please upgrade!";
```
teor2345: This message is repeated, let's #define it once

src/feature/dirauth/process_descs.c
```diff
@@ -315,6 +315,41 @@ dirserv_would_reject_router(const routerstatus_t *rs)
...
+ */
+STATIC bool
+dirserv_rejects_tor_version(const char *platform,
+                            const char **msg)
+{
+  if (!platform)
+    return false;
+
+  /* Versions before Tor 0.2.9 are unsupported. Versions between 0.2.9.0 and
+   * 0.2.9.5 suffer from bug #20499, where relays don't keep their consensus
```
teor2345:
```suggestion
   * 0.2.9.4 suffer from bug #20499, where relays don't keep their consensus
```

src/feature/dirauth/process_descs.c
nmathewson: Sounds good to me.  0.3.5.7 should also be removed eventually, but IMO it's a separate thing.

nmathewson: 170c8b6ca0b607 changes the code's behavior; 4e54c1812776e2 updates the tests.

src/feature/dirauth/process_descs.c
nmathewson: Done in 064b9598d1baaf, preferring a const to a #define

src/feature/dirauth/process_descs.c
nmathewson: Good catch: fixed in 5548c12415b28d

PR #1277, opened 2019-08-30
Bug31571 035

teor2345:


src/lib/err/backtrace.c
```diff
@@ -274,7 +270,10 @@ configure_backtrace_handler(const char *tor_version)
     snprintf(version, sizeof(version), "Tor %s", tor_version);
   }
 
-  return install_bt_handler(version);
+  strncpy(bt_version, version, sizeof(bt_version) - 1);
```
nmathewson: In general our style is to prefer strlcpy unless there is a strong reason not to use it.

src/lib/err/torerr.c
```diff
@@ -146,13 +146,19 @@ tor_raw_assertion_failed_msg_(const char *file, int line, const char *expr,
...
     tor_log_err_sigsafe_write("\n");
   }
 
   dump_stack_symbols_to_error_fds();
+
+  /* Some platforms (macOS, maybe others?) can swallow the last write before an
+   * abort. This issue is probably caused by a race condition between write
+   * buffer cache flushing, and process termination. So we write an extra
+   * newline, to make sure that the message always gets through. */
+  tor_log_err_sigsafe_write("\n");
```
nmathewson: Do you think it would make sense, if we are about to crash, to also try to close all of the log fds right before we abort()? Doing so might also help make sure they get flushed.

If you think this is a good idea, I'll open a ticket: it isn't a blocker for this one IMO.

src/lib/err/backtrace.c
teor2345: I moved the previous implementation between functions.

I can change the whole module to use strlcpy() in another commit?

src/lib/err/torerr.c
teor2345: Yeah, let's try closing fds before abort.
But let's also use the extra newline, because race conditions.

teor2345: This is now:
https://trac.torproject.org/projects/tor/ticket/31594

src/lib/err/backtrace.c
teor2345: Fixed in ac929fb

PR #1289, opened 2019-09-04
WIP Bug31594 DO NOT MERGE

teor2345:


src/lib/err/torerr.c
```diff
@@ -118,8 +120,15 @@ tor_log_set_sigsafe_err_fds(const int *fds, int n)
     n = TOR_SIGSAFE_LOG_MAX_FDS;
   }
 
-  memcpy(sigsafe_log_fds, fds, n * sizeof(int));
-  n_sigsafe_log_fds = n;
+  /* Clear the entire array. There's a race condition here: err logs are
+   * disabled while the array is cleared. But that's better than accessing a
+   * partially written array. */
```
nmathewson: I agree about the race condition -- but this change does not solve it completely.  Another thread can still read n_sigsafe_log_fds before it is changed, then read the new values of sigsafe_log_fds after they are zeroed, or while they are partially written.

To solve this right, we would probably need to do something with atomics, which we are not currently exposing at this level.

I do not think problem should block merge of this branch: it is still an improvement over the status quo.

src/lib/log/log.c
```diff
@@ -834,6 +834,23 @@ logs_free_all(void)
    * that happened between here and the end of execution. */
 }
 
+/** Close signal-safe log files. */
```
nmathewson: Could we have a little more documentation about when this function is appropriate to call?

src/lib/err/torerr.c
teor2345: Ok, I will update the comment so that it's clear there's still some race here, and why.

src/lib/err/torerr.c
```diff
@@ -111,6 +111,12 @@ tor_log_get_sigsafe_err_fds(const int **out)
  * other emergency condition. Ignore any beyond the first
  * TOR_SIGSAFE_LOG_MAX_FDS.
  *
+ * These fds must remain open even after the log module has shut down. (And
+ * they should remain open even while logs are being reconfigured.) Therefore,
+ * any fds closed by the log module should be dup()ed, and the duplicate fd
+ * should be given to the err module in fds. In particular, the log module
+ * owns file log fds, but does not own stdio log fds.
```
teor2345: "closes" might be better than "owns" here

src/lib/log/log.c
```diff
@@ -834,6 +834,23 @@ logs_free_all(void)
...
+{
+  logfile_t *victim, *next;
+  /* We can't lock the logs in a signal handler, or deallocate memory */
+  next = logfiles;
+  logfiles = NULL;
+  while (next) {
+    victim = next;
+    next = next->next;
+    if (victim->needs_close) {
+      close_log(victim);
```
teor2345: It might be safer to split close_log() into close_log() and close_log_sigsafe().

src/lib/log/log.c
```diff
@@ -834,6 +834,23 @@ logs_free_all(void)
    * that happened between here and the end of execution. */
 }
 
+/** Close signal-safe log files. */
+void
+logs_close_minimal(void)
```
teor2345: Let's rename this logs_close_sigsafe().

src/lib/log/log.c
teor2345: Note to self: this function can be called from a signal handler. It should only be called when shutting down the log module. It's designed to be called by the err module, when terminating the process on an abnormal condition.

src/lib/err/torerr.c
teor2345: Fixed in 671ec51163ec69e8b632bb121ae60354b9bd5b1d

src/lib/log/log.c
teor2345: Fixed in 671ec51163ec69e8b632bb121ae60354b9bd5b1d

src/lib/err/torerr.c
teor2345: Fixed in 3d82b78afd1d7a8d1dd53a1ddb1dd2455c75fd41

src/lib/log/log.c
teor2345: Fixed in 671ec51163ec69e8b632bb121ae60354b9bd5b1d

src/lib/log/log.c
teor2345: Fixed in 671ec51163ec69e8b632bb121ae60354b9bd5b1d

teor2345:
I made a mistake with the fixup commits, I force-pushed the branch and updated the fixup commit hashes in my github comments.

PR #1294, opened 2019-09-04
Ticket31627 (missing config documentation)

nmathewson:


src/app/config/confparse.c
```diff
@@ -575,6 +578,10 @@ config_mark_lists_fragile(const config_mgr_t *mgr, void *options)
   } SMARTLIST_FOREACH_END(mv);
 }
 
+/**
+ * Log a warning that declaring that the option called <b>what</b>
+ * is deprecated because of the reason in <b>why</b>.
+ **/
 void
 warn_deprecated_option(const char *what, const char *why)
```
teor2345: This function will fail in log_warn() if what or why are NULL.

We should document that requirement and maybe add BUG() logs.
Or we could pass "" to log_warn() if what or why are NULL.

PR #1297, opened 2019-09-05
[Tor Trac #31652]: hs-v3: Service circuit retry limit should not close a valid circuit

neelchauhan:
For the bug [here](https://trac.torproject.org/projects/tor/ticket/31652).

### DIFF
--- a/src/feature/hs/hs_service.c
+++ b/src/feature/hs/hs_service.c
@@ -2612,10 +2612,14 @@ launch_intro_point_circuits(hs_service_t *service)
         ei = get_extend_info_from_intro_point(ip, 0);
       }
 
-      if (ei == NULL) {
-        /* This is possible if we can get a node_t but not the extend info out
-         * of it. In this case, we remove the intro point and a new one will
-         * be picked at the next main loop callback. */
+      /* We remove the intro point if:
+       * - We can get a node_t but not the extend info out of it.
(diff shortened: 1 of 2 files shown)


src/feature/hs/hs_service.c
```diff
@@ -2612,10 +2612,14 @@ launch_intro_point_circuits(hs_service_t *service)
...
-         * of it. In this case, we remove the intro point and a new one will
-         * be picked at the next main loop callback. */
+      /* We remove the intro point if:
+       * - We can get a node_t but not the extend info out of it.
+       * - The intro point has retried the circuit more than
+       *   MAX_INTRO_POINT_CIRCUIT_RETRIES times
+       * In this case, we remove the intro point and a new one will be picked
+       * at the next main loop callback. */
+      if (ei == NULL ||
+          ip->circuit_retries > MAX_INTRO_POINT_CIRCUIT_RETRIES) {
```
dgoulet-tor: So unfortunately, we can't do this here. Problem is that this IP needs to be remembered in our "failing cache", see `remember_failing_intro_point()`.

I think the IP should be cleaned up _before_ we get into the build circuit main loop event which calls this function.

The obvious candidate is `cleanup_intro_points()` which is where it is currently happening but I think the checks needs to be something around:

"if too many retries and _no_ existing circuits: cleanup()"

PR #1299, opened 2019-09-05
Ticket31354

nmathewson:


src/test/test_addr.c
```diff
@@ -17,6 +17,8 @@
 #include <sys/un.h>
 #endif
 
+extern int lambada;
```
tlyu: This seems extraneous?

nmathewson: Indeed. Wow. I have no memory where that came from.

PR #1302, opened 2019-09-09
Bug31614 035

teor2345:


src/lib/err/backtrace.c
```diff
@@ -190,13 +191,15 @@ dump_stack_symbols_to_error_fds(void)
     backtrace_symbols_fd(cb_buf, (int)depth, fds[i]);
 }
 
+/* The signals that we want our backtrace handler to trap */
+static int trap_signals[] = { SIGSEGV, SIGILL, SIGFPE, SIGBUS, SIGSYS,
+  SIGIO, -1 };
```
dgoulet-tor: This in theory should be indented properly to follow our coding standards. I'm not going to hold up on this though to get it merged but if you have a minute, great! :)

src/lib/err/backtrace.c
```diff
@@ -235,6 +238,24 @@ install_bt_handler(const char *software)
...
+  for (i = 0; trap_signals[i] >= 0; ++i) {
+    /* remove_bt_handler() is called on shutdown, from low-level code.
+     * It's not a fatal error, so we just ignore it. */
+    (void)sigaction(trap_signals[i], &sa, NULL);
+  }
+
+  /* We can't use tor_mutex_free() here, because lock uses err.
+   * This function is called while err is being shut down, so we can't
+   * log or take any useful action if pthread_mutex_destroy() fails. */
+  (void)pthread_mutex_destroy(&cb_buf_mutex);
```
dgoulet-tor: Is this comment accurate? It seems to me it is saying we can be in this handler while the mutex is locked? I would be a bit surprised but just in case, this can't be done. Man page states:

> It shall be safe to destroy an initialized mutex that is unlocked.  Attempting to destroy a locked mutex  or  a  mutex that  is  referenced  (for  example, while being used in a pthread_cond_timedwait() or pthread_cond_wait()) by another thread results in undefined behavior.

teor2345: You're right, there are two race conditions here:
1. log_backtrace_impl() can be called from other threads, and lock cb_buf_mutex, while the main thread is calling this function during shut down, which leads to undefined behaviour

2. cb_buf is used in a lot of functions, which could be called concurrently:
a) install_bt_handler
b) dump_stack_symbols_to_error_fds
c) crash_handler
d) log_backtrace_impl (protected by cb_buf_mutex)
Which could lead to corrupted stack traces.

I don't know if there is an easy way to fix this. Maybe I should just open a ticket.

@nmathewson do you have any ideas?

PR #1305, opened 2019-09-09
build: The <sys/sysctl.h> is now deprecated on Linux

dgoulet-tor:
Closes #31673

### DIFF
--- a/src/common/compat.c
+++ b/src/common/compat.c
@@ -27,9 +27,6 @@
 #ifdef HAVE_SYS_TYPES_H
 #include <sys/types.h>
 #endif
-#ifdef HAVE_SYS_SYSCTL_H
-#include <sys/sysctl.h>
-#endif
 #ifdef HAVE_SYS_STAT_H
 #include <sys/stat.h>
 #endif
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -1049,8 +1049,8 @@ AC_CHECK_HEADERS([assert.h \
                   sys/socket.h \
                   sys/statvfs.h \
                   sys/syscall.h \
-                  sys/sysctl.h \
                   sys/syslimits.h \
+                  sys/sysctl.h \
```
nmathewson: This reordering isn't needed.

src/common/compat.c
```diff
@@ -70,6 +67,11 @@
 #include <sys/capability.h>
 #endif
 
+/* Now deprecated in Linux GLIBC */
+#if HAVE_SYS_SYSCTL_H && !defined(_WIN32) && !defined(__linux__)
```
nmathewson: This should check "defined(HAVE_SYS_SYSCTL_H)", not "HAVE_SYS_SYSCTL_H" -- it causes appveyor failures otherwise.

PR #1307, opened 2019-09-09
build: The <sys/sysctl.h> is now deprecated on Linux

dgoulet-tor:
Closes #31673

### DIFF
--- a/src/lib/meminfo/meminfo.c
+++ b/src/lib/meminfo/meminfo.c
@@ -18,9 +18,6 @@
 #include "lib/log/log.h"
 #include "lib/malloc/malloc.h"
 
-#ifdef HAVE_SYS_SYSCTL_H
-#include <sys/sysctl.h>
-#endif
 #ifdef HAVE_FCNTL_H
 #include <fcntl.h>
 #endif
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -1473,8 +1473,8 @@ AC_CHECK_HEADERS([errno.h \
 		  sys/socket.h \
 		  sys/statvfs.h \
 		  sys/syscall.h \
-		  sys/sysctl.h \
 		  sys/syslimits.h \
+		  sys/sysctl.h \
```
teor2345: I don't think this change actually does anything?
So let's keep the headers in alphabetical order.

src/lib/meminfo/meminfo.c
```diff
@@ -36,6 +33,10 @@
 #endif
 #include <string.h>
 
+#if HAVE_SYS_SYSCTL_H && !defined(_WIN32) && !defined(__linux__)
```
nmathewson: We should check "defined(HAVE_SYS_SYSCTL_H)", not "SYS_SYSCTL_H": appveyor is failing with this error:

```
bash.exe : ../src/lib/meminfo/meminfo.c:36:5: error: "HAVE_SYS_SYSCTL_H" is not defined, evaluates to 0 [-Werror=undef]
```

configure.ac
nmathewson: +1

PR #1315, opened 2019-09-10
[Tor Trac #31652] hs-v3: Service circuit retry limit should not close a valid circuit 

neelchauhan:
For the ticket [here](https://trac.torproject.org/projects/tor/ticket/31652).

### DIFF
--- a/src/feature/hs/hs_service.c
+++ b/src/feature/hs/hs_service.c
@@ -2324,18 +2324,19 @@ intro_point_should_expire(const hs_service_intro_point_t *ip,
   return 1;
 }
 
-/* Return if we can retry the intro point for the service defined in <b>ip</b>.
+/* Return if we should remove the intro point for the service defined in
+ * <b>ip</b>.
  */
 static inline bool
-should_not_retry_intro_point(hs_service_intro_point_t *ip)
+should_remove_intro_point(hs_service_intro_point_t *ip)
 {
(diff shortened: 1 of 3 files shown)


src/feature/hs/hs_service.c
```diff
@@ -2364,9 +2373,7 @@ cleanup_intro_points(hs_service_t *service, time_t now)
        * the node_t anymore (removed from our latest consensus) or if we've
        * reached the maximum number of retry with a non existing circuit
        * and don't have any established circuits. */
-      if (has_expired || node == NULL ||
-          (ip->circuit_retries > MAX_INTRO_POINT_CIRCUIT_RETRIES &&
-           (!ip->circuit_established || hs_circ_service_get_intro_circ(ip)))) {
+      if (has_expired || node == NULL || can_retry_intro_point(ip)) {
```
asn-d6: I'm a bit confused on why we trash the intro point if we can retry it. Intuitively we should do the opposite right?

src/feature/hs/hs_service.c
```diff
@@ -2324,6 +2324,15 @@ intro_point_should_expire(const hs_service_intro_point_t *ip,
   return 1;
 }
 
+/* Return if we can retry the intro point for the service defined in <b>ip</b>.
+ */
+static inline int
```
asn-d6: This function could be bool too.

PR #1321, opened 2019-09-11
Ticket31705

nmathewson:


scripts/coccinelle/try_parse.sh
```diff
@@ -0,0 +1,18 @@
...
+top="$(dirname "$0")/../.."
+
+for fn in "$@"; do
+
+    if spatch -macro_file_builtins "$top"/scripts/coccinelle/tor-coccinelle.h \
+              -I "$top" -I "$top"/src -I "$top"/ext --parse-c "$fn" \
+              2>/dev/null | grep "perfect = 1" > /dev/null; then
+        : # it's perfect
+    else
+        echo "$fn"
```
teor2345: Can you make this script exit with a non-zero exit status if it finds any imperfect files?
That will let us call it from make and the git hooks in trac 31919.

scripts/coccinelle/tor-coccinelle.h
```diff
@@ -1,3 +1,21 @@
 #define MOCK_IMPL(a, b, c) a b c
```
teor2345: This header needs some documentation, and the text in the commit message is a good starting point:

Note that this header file behaves a bit strangely.  It is used by
coccinelle just for the purpose of knowing how to parse
difficult-to-parse stuff.  It doesn't need to produce good C -- just
grammatical C.

teor2345:
It would also be nice if test-operator-cleanup had some documentation. One short sentence should be enough.

teor2345:
It looks like you missed lots of `src/*/*/*.[ch]` in your cleanup.

scripts/coccinelle/try_parse.sh
teor2345: Can you also make try_parse.sh show the parsing errors, so they can be fixed?
If this verbose mode needs to be activated by an env var or argument, that's ok.

PR #1328, opened 2019-09-16
Ticket31684

ltbringer:
Add control port GETINFO support for dumping the local consensus

Refer: (ticket)[https://trac.torproject.org/projects/tor/ticket/31684]

### DIFF
--- a/changes/ticket31684
+++ b/changes/ticket31684
@@ -1,0 +1,2 @@
+  o Minor features (onion services):
+    - GETINFO dir/status-vote/current/consensus-microdesc dumps microdesc consensus. Closes ticket 31684.
--- a/changes/ticket31684
+++ b/changes/ticket31684
@@ -1,2 +1,6 @@
   o Minor features (onion services):
-    - GETINFO dir/status-vote/current/consensus-microdesc dumps microdesc consensus. Closes ticket 31684.
+    - Implement a new GETINFO command to fetch microdescriptor consensus. Closes ticket 31684.
+
+  o Code simplification and refactoring (onion services):
+    - Create a helper function that can fetch network status or microdesc consensuses.
+
--- a/src/feature/control/control_getinfo.c
+++ b/src/feature/control/control_getinfo.c
@@ -325,6 +325,51 @@ getinfo_helper_current_time(control_connection_t *control_conn,
   return 0;
 }
 
+/**
+ * Switch between microdesc vs networkstatus descriptor dumps
+ * Assumes that question would be either:
+ * - dir/status-vote/current/consensus-microdesc
+ * or
+ * - dir/status-vote/current/consensus
+ */
+STATIC int
+getinfo_helper_current_consensus(const char* question,
+                          char** answer,
+                          const char** errmsg)
+{
+  // Ensures question contains a request for
+  // ns or microdesc consensus
+  if (
+      strcmp(question, "dir/status-vote/current/consensus") &&
+      strcmp(question, "dir/status-vote/current/consensus-microdesc")) {
+    return 0;
+  }
+
+  const char* consensus_type = !strcmp(
+     question, "dir/status-vote/current/consensus-microdesc"
+  ) ? "microdesc" : "ns";
+
+  if (we_want_to_fetch_flavor(get_options(), FLAV_NS)) {
+    const cached_dir_t *consensus = dirserv_get_consensus(consensus_type);
+    if (consensus) {
+      *answer = tor_strdup(consensus->dir);
+    }
+  }
+  if (!*answer) { /* try loading it from disk */
+    tor_mmap_t *mapped = networkstatus_map_cached_consensus(consensus_type);
+    if (mapped) {
+      *answer = tor_memdup_nulterm(mapped->data, mapped->size);
+      tor_munmap_file(mapped);
+    }
+    if (!*answer) { /* generate an error */
+      *errmsg = "Could not open cached consensus. "
+        "Make sure FetchUselessDescriptors is set to 1.";
+       return -1;
+   }
+  }
+  return 0;
+}
+
 /** Implementation helper for GETINFO: knows the answers for questions about
  * directory information. */
 STATIC int
@@ -362,6 +407,7 @@ getinfo_helper_dir(control_connection_t *control_conn,
       if (body)
         *answer = tor_strndup(body, ri->cache_info.signed_descriptor_len);
     } else if (! we_fetch_router_descriptors(get_options())) {
+
       /* Descriptors won't be available, provide proper error */
       *errmsg = "We fetch microdescriptors, not router "
                 "descriptors. You'll need to use md/name/* "
@@ -576,24 +622,16 @@ getinfo_helper_dir(control_connection_t *control_conn,
     smartlist_free(descs);
   } else if (!strcmpstart(question, "dir/status/")) {
     *answer = tor_strdup("");
-  } else if (!strcmp(question, "dir/status-vote/current/consensus")) { /* v3 */
-    if (we_want_to_fetch_flavor(get_options(), FLAV_NS)) {
-      const cached_dir_t *consensus = dirserv_get_consensus("ns");
-      if (consensus)
-        *answer = tor_strdup(consensus->dir);
-    }
-    if (!*answer) { /* try loading it from disk */
-      tor_mmap_t *mapped = networkstatus_map_cached_consensus("ns");
-      if (mapped) {
-        *answer = tor_memdup_nulterm(mapped->data, mapped->size);
-        tor_munmap_file(mapped);
-      }
-      if (!*answer) { /* generate an error */
-        *errmsg = "Could not open cached consensus. "
-          "Make sure FetchUselessDescriptors is set to 1.";
-        return -1;
-      }
-    }
+  } else if (-1 == getinfo_helper_current_consensus(
+                                                    question,
+                                                    answer,
+                                                    errmsg))
+  {
+    /**
+     * answer is set by getinfo_helper_current_consensus
+     * if the question matches
+     */
+    return -1;
   } else if (!strcmp(question, "network-status")) { /* v1 */
     static int network_status_warned = 0;
     if (!network_status_warned) {
@@ -1513,6 +1551,8 @@ static const getinfo_item_t getinfo_items[] = {
          "v2 networkstatus docs as retrieved from a DirPort."),
   ITEM("dir/status-vote/current/consensus", dir,
        "v3 Networkstatus consensus as retrieved from a DirPort."),
+  ITEM("dir/status-vote/current/consensus-microdesc", dir,
+       "v3 Microdescriptors consensus as retrieved from a DirPort."),
   ITEM("exit-policy/default", policies,
        "The default value appended to the configured exit policy."),
   ITEM("exit-policy/reject-private/default", policies,
--- a/src/feature/control/control_getinfo.c
+++ b/src/feature/control/control_getinfo.c
@@ -328,12 +328,12 @@ getinfo_helper_current_time(control_connection_t *control_conn,
 /** GETINFO helper for dumping different consensus flavors
  * returns: 0 on success -1 on error. */
 STATIC int
-getinfo_helper_current_consensus(int flavor,
+getinfo_helper_current_consensus(consensus_flavor_t flavor,
                                  char** answer,
                                  const char** errmsg)
 {
   const char *flavor_name = networkstatus_get_flavor_name(flavor);
-  if (!strcmp(flavor_name, "")) {
+  if (!strcmp(flavor_name, "??")) {
     *errmsg = "Could not open cached consensus. "
       "Make sure FetchUselessDescriptors is set to 1.";
     return -1;
--- a/src/feature/control/control_getinfo.c
+++ b/src/feature/control/control_getinfo.c
@@ -325,38 +325,28 @@ getinfo_helper_current_time(control_connection_t *control_conn,
   return 0;
 }
 
-/**
- * Switch between microdesc vs networkstatus descriptor dumps
- * Assumes that question would be either:
- * - dir/status-vote/current/consensus-microdesc
- * or
- * - dir/status-vote/current/consensus
- */
+/** GETINFO helper for dumping different consensus flavors
+ * returns: 0 on success -1 on error. */
 STATIC int
-getinfo_helper_current_consensus(const char* question,
-                          char** answer,
-                          const char** errmsg)
+getinfo_helper_current_consensus(int flavor,
+                                 char** answer,
+                                 const char** errmsg)
 {
-  // Ensures question contains a request for
-  // ns or microdesc consensus
-  if (
-      strcmp(question, "dir/status-vote/current/consensus") &&
-      strcmp(question, "dir/status-vote/current/consensus-microdesc")) {
-    return 0;
+  const char *flavor_name = networkstatus_get_flavor_name(flavor);
+  if (!strcmp(flavor_name, "")) {
+    *errmsg = "Could not open cached consensus. "
+      "Make sure FetchUselessDescriptors is set to 1.";
+    return -1;
   }
-
-  const char* consensus_type = !strcmp(
-     question, "dir/status-vote/current/consensus-microdesc"
-  ) ? "microdesc" : "ns";
-
-  if (we_want_to_fetch_flavor(get_options(), FLAV_NS)) {
-    const cached_dir_t *consensus = dirserv_get_consensus(consensus_type);
+  if (we_want_to_fetch_flavor(get_options(), flavor)) {
+    /** Check from the cache */
+    const cached_dir_t *consensus = dirserv_get_consensus(flavor_name);
     if (consensus) {
       *answer = tor_strdup(consensus->dir);
     }
   }
   if (!*answer) { /* try loading it from disk */
-    tor_mmap_t *mapped = networkstatus_map_cached_consensus(consensus_type);
+    tor_mmap_t *mapped = networkstatus_map_cached_consensus(flavor_name);
     if (mapped) {
       *answer = tor_memdup_nulterm(mapped->data, mapped->size);
       tor_munmap_file(mapped);
@@ -364,8 +354,8 @@ getinfo_helper_current_consensus(const char* question,
     if (!*answer) { /* generate an error */
       *errmsg = "Could not open cached consensus. "
         "Make sure FetchUselessDescriptors is set to 1.";
-       return -1;
-   }
+      return -1;
+    }
   }
   return 0;
 }
@@ -622,16 +612,19 @@ getinfo_helper_dir(control_connection_t *control_conn,
     smartlist_free(descs);
   } else if (!strcmpstart(question, "dir/status/")) {
     *answer = tor_strdup("");
-  } else if (-1 == getinfo_helper_current_consensus(
-                                                    question,
-                                                    answer,
-                                                    errmsg))
-  {
-    /**
-     * answer is set by getinfo_helper_current_consensus
-     * if the question matches
-     */
-    return -1;
+  } else if (!strcmp(question, "dir/status-vote/current/consensus")) {
+    int consensus_result = getinfo_helper_current_consensus(FLAV_NS,
+                                                            answer, errmsg);
+    if (consensus_result == -1) {
+      return -1;
+    }
+  } else if (!strcmp(question,
+                     "dir/status-vote/current/consensus-microdesc")) {
+    int consensus_result = getinfo_helper_current_consensus(FLAV_MICRODESC,
+                                                            answer, errmsg);
+    if (consensus_result == -1) {
+        return -1;
+      }
   } else if (!strcmp(question, "network-status")) { /* v1 */
     static int network_status_warned = 0;
     if (!network_status_warned) {
--- a/src/feature/control/control_getinfo.h
+++ b/src/feature/control/control_getinfo.h
@@ -49,7 +49,7 @@ STATIC int getinfo_helper_downloads(
     const char *question, char **answer,
     const char **errmsg);
 STATIC int getinfo_helper_current_consensus(
-    int flavor,
+    consensus_flavor_t flavor,
     char **answer,
     const char **errmsg);
 STATIC int getinfo_helper_dir(
--- a/src/feature/control/control_getinfo.h
+++ b/src/feature/control/control_getinfo.h
@@ -48,6 +48,9 @@ STATIC int getinfo_helper_downloads(
     control_connection_t *control_conn,
     const char *question, char **answer,
     const char **errmsg);
+STATIC int getinfo_helper_current_consensus(
+    const char *question, char **answer,
+    const char **errmsg);
 STATIC int getinfo_helper_dir(
     control_connection_t *control_conn,
     const char *question, char **answer,
--- a/src/feature/control/control_getinfo.h
+++ b/src/feature/control/control_getinfo.h
@@ -49,7 +49,8 @@ STATIC int getinfo_helper_downloads(
     const char *question, char **answer,
     const char **errmsg);
 STATIC int getinfo_helper_current_consensus(
-    const char *question, char **answer,
+    int flavor,
+    char **answer,
     const char **errmsg);
 STATIC int getinfo_helper_dir(
     control_connection_t *control_conn,
--- a/src/test/test_controller.c
+++ b/src/test/test_controller.c
@@ -27,6 +27,7 @@
 #include "feature/dirclient/download_status_st.h"
 #include "feature/nodelist/microdesc_st.h"
 #include "feature/nodelist/node_st.h"
+#include "feature/dircache/dirserv.c"
 
 typedef struct {
   const char *input;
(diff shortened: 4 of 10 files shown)


changes/ticket31684
```diff
@@ -0,0 +1,2 @@
+  o Minor features (onion services):
+    - GETINFO dir/status-vote/current/consensus-microdesc dumps microdesc consensus. Closes ticket 31684.
```
asn-d6: Please check how other changes entries look like. They are usually human readable and look like "Implement a new GETINFO command ...".

src/feature/control/control_getinfo.c
```diff
@@ -325,6 +325,51 @@ getinfo_helper_current_time(control_connection_t *control_conn,
   return 0;
 }
 
+/**
+ * Switch between microdesc vs networkstatus descriptor dumps
```
asn-d6: Please improve the function documentation here by making it more readable. Please see how the other GETINFO helper functions are documented and take a similar human readable approach.

Also please document the return value of the function.

src/feature/control/control_getinfo.c
```diff
@@ -325,6 +325,51 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
+ * - dir/status-vote/current/consensus-microdesc
+ * or
+ * - dir/status-vote/current/consensus
+ */
+STATIC int
+getinfo_helper_current_consensus(const char* question,
+                          char** answer,
+                          const char** errmsg)
+{
+  // Ensures question contains a request for
```
asn-d6: Please use the `/*` `*/` comment style for comments.

src/feature/control/control_getinfo.c
```diff
@@ -325,6 +325,51 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
+      strcmp(question, "dir/status-vote/current/consensus") &&
+      strcmp(question, "dir/status-vote/current/consensus-microdesc")) {
+    return 0;
+  }
+
+  const char* consensus_type = !strcmp(
+     question, "dir/status-vote/current/consensus-microdesc"
+  ) ? "microdesc" : "ns";
+
+  if (we_want_to_fetch_flavor(get_options(), FLAV_NS)) {
```
asn-d6: Hmm, is this right? We call the `we_want_to_fetch_flavor()` with `FLAV_NS` but then we call `dirserv_get_consensus()` with arbitrary `consensus_type`?

Also, can you please document what's the purpose of this `we_want_to_fetch_flavor()` code block here?

src/feature/control/control_getinfo.c
```diff
@@ -362,6 +407,7 @@ getinfo_helper_dir(control_connection_t *control_conn,
       if (body)
         *answer = tor_strndup(body, ri->cache_info.signed_descriptor_len);
     } else if (! we_fetch_router_descriptors(get_options())) {
+
```
asn-d6: Newline added for no real reason.

src/feature/control/control_getinfo.c
```diff
@@ -576,24 +622,16 @@ getinfo_helper_dir(control_connection_t *control_conn,
...
-        *answer = tor_memdup_nulterm(mapped->data, mapped->size);
-        tor_munmap_file(mapped);
-      }
-      if (!*answer) { /* generate an error */
-        *errmsg = "Could not open cached consensus. "
-          "Make sure FetchUselessDescriptors is set to 1.";
-        return -1;
-      }
-    }
+  } else if (-1 == getinfo_helper_current_consensus(
```
asn-d6: I'm not sure if I like the interface here:
- If you see all the other checks in `getinfo_helper_dir()` you will notice that they only get in the `if` block if the GETINFO command matches, whereas in this patch we get in the block if there is an error. I think this is not the way it should be done, and we should copy the approach of the rest of the function. Also:

- I'd prefer if we explicitly checked the GETINFO string in this function like above, instead of hiding it into the helper function. I would probably go with:
  ```
  ...
  } else if (!strcmp(question, "dir/status-vote/current/consensus")) { /* ns consensus */
      get_info_helper_current_consensus("ns");
  } else if (!strcmp(question, "dir/status-vote/current/consensus-microdesc)) { /* md consensus */
      get_info_helper_current_consensus("nicrodesc");
  ...
  ```

- We don't usually use Yoda conditions in Tor.

src/feature/control/control_getinfo.c
```diff
@@ -576,24 +622,16 @@ getinfo_helper_dir(control_connection_t *control_conn,
...
+                                                    question,
```
asn-d6: Why intend the arguments in 4 different lines? If they fit, just put them in the same line except if you have something in mind.

src/feature/control/control_getinfo.c
```diff
@@ -325,6 +325,51 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
+ * - dir/status-vote/current/consensus
+ */
+STATIC int
+getinfo_helper_current_consensus(const char* question,
+                          char** answer,
+                          const char** errmsg)
+{
+  // Ensures question contains a request for
+  // ns or microdesc consensus
+  if (
```
asn-d6: Why is there a gap here? You could put the `strcmp` in the `if` line, except if you have something specific in mind.

src/feature/control/control_getinfo.c
```diff
@@ -1513,6 +1551,8 @@ static const getinfo_item_t getinfo_items[] = {
          "v2 networkstatus docs as retrieved from a DirPort."),
   ITEM("dir/status-vote/current/consensus", dir,
        "v3 Networkstatus consensus as retrieved from a DirPort."),
+  ITEM("dir/status-vote/current/consensus-microdesc", dir,
```
asn-d6: Please consider adding a unittest for the new functionality. See how `test_download_status_consensus()` does it for example.

src/feature/control/control_getinfo.c
ltbringer: There is a lint error that restricts lines to some 78 chars. That's why I thought of breaking them into different lines.

src/feature/control/control_getinfo.c
ltbringer: I understand, I got stuck thinking about how to move things out from the conditionals. I'll fix it.

src/feature/control/control_getinfo.c
```diff
@@ -326,46 +326,33 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
-
-  const char* consensus_type = !strcmp(
-     question, "dir/status-vote/current/consensus-microdesc"
-  ) ? "microdesc" : "ns";
 
-  if (we_want_to_fetch_flavor(get_options(), FLAV_NS)) {
-    const cached_dir_t *consensus = dirserv_get_consensus(consensus_type);
-    if (consensus) {
-      *answer = tor_strdup(consensus->dir);
-    }
```
asn-d6: Are we sure we don't need the `we_want_to_fetch_flavor()` call here anymore? What purpose was it serving in the past?

src/test/test_controller.c
```diff
@@ -1689,6 +1689,42 @@ test_download_status_bridge(void *arg)
...
+  control_connection_t dummy;
+  /* Get results out */
+  char *answer = NULL;
+  const char *errmsg = NULL;
+
+  (void)arg;
+
+  setup_bridge_mocks();
+
+  getinfo_helper_dir(&dummy,
```
asn-d6: Nice start! However this unittest does not actually test that the right consensus is returned. Ideally the unittest here should make sure that the right consensus string will be returned. (Hint: This might require some mocking)

src/feature/control/control_getinfo.c
ltbringer: I checked `we_want_to_fetch_flavor()` function and I saw that it returns a list of consensus flavors. since we are passing that ourselves, I thought about removing it.

teor2345: I think the we_want_to_fetch_flavor() check is used to make sure that the flavour is in RAM or on disk.

src/feature/control/control_getinfo.c
teor2345: I think the we_want_to_fetch_flavor() check is used to make sure that the flavour is in RAM or on disk.

src/feature/control/control_getinfo.c
```diff
@@ -326,46 +326,37 @@ getinfo_helper_current_time(control_connection_t *control_conn,
 }
 
 /**
- * Switch between microdesc vs networkstatus descriptor dumps
- * Assumes that question would be either:
- * - dir/status-vote/current/consensus-microdesc
- * or
- * - dir/status-vote/current/consensus
+ * Helps to access directory information for a given
```
asn-d6: This still doesn't look like our usual documentation format. Please describe the return value in a proper sentence and not in a list.

src/feature/control/control_getinfo.c
```diff
@@ -326,46 +326,37 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
       *answer = tor_strdup(consensus->dir);
     }
   }
-  if (!*answer) { /* try loading it from disk */
+  if (!*answer) { /* else, try loading consensus from disk */
     tor_mmap_t *mapped = networkstatus_map_cached_consensus(consensus_type);
     if (mapped) {
       *answer = tor_memdup_nulterm(mapped->data, mapped->size);
       tor_munmap_file(mapped);
+      tor_free(mapped);
```
asn-d6: Why are you freeing this here? Are you sure this is right? Please double check this because I think you might be freeing the actual global consensus.

src/feature/control/control_getinfo.c
```diff
@@ -622,16 +613,17 @@ getinfo_helper_dir(control_connection_t *control_conn,
...
-                                                    answer,
-                                                    errmsg))
-  {
-    /**
-     * answer is set by getinfo_helper_current_consensus
-     * if the question matches
-     */
-    return -1;
+  } else if (!strcmp(question, "dir/status-vote/current/consensus")) {
+    if (getinfo_helper_current_consensus(
```
asn-d6: Please don't wrap the code like this. Check the rest of the code, we usually wrap it between the arguments and not after the parentheses.

We also usually do `< 0` when checking for errors and not `== -1`.

src/test/test_controller.c
```diff
@@ -1689,6 +1695,62 @@ test_download_status_bridge(void *arg)
...
+}
+
+static void
+test_getinfo_helper_dir(void *arg)
+{
+  /* We just need one of these to pass, it doesn't matter what's in it */
+  control_connection_t dummy;
+  /* Get results out */
+  char *answer = NULL;
+  const char *errmsg = NULL;
```
asn-d6: This unittest is failing in Appveyor for Windows. Any idea why? It might be that the extra free is causing trouble, or something else.

src/feature/control/control_getinfo.c
```diff
@@ -326,46 +326,37 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
-  }
-
-  const char* consensus_type = !strcmp(
-     question, "dir/status-vote/current/consensus-microdesc"
-  ) ? "microdesc" : "ns";
-
-  if (we_want_to_fetch_flavor(get_options(), FLAV_NS)) {
+  /** Check if the consensus flavor is cached or in the ram. */
+  if (we_want_to_fetch_flavor(get_options(), flavor)) {
     const cached_dir_t *consensus = dirserv_get_consensus(consensus_type);
```
asn-d6: Can you also test this branch in the unittest perhaps?

ltbringer: Yes I can try.

src/feature/control/control_getinfo.c
ltbringer: The CI was complaining about memory leaks and that's why I tried to do this. I found that the problem was in the test itself where I wasn't freeing `answer` I'll undo this and check.

src/test/test_controller.c
ltbringer: It is something else :( I haven't been able to find what exactly. Without the extra free here the CI was failing on both Travis and Appveyor due to memory leaks.

src/feature/control/control_getinfo.c
```diff
@@ -325,6 +325,33 @@ getinfo_helper_current_time(control_connection_t *control_conn,
   return 0;
 }
 
+/**
+ * Helps to access directory information for a given
+ * type of consensus.
+ *
+ * returns: int
```
teor2345: `returns: int` is redundant, because the code tells us the type that the function returns.

changes/ticket31684
```diff
@@ -0,0 +1,2 @@
+  o Minor features (onion services):
+    - Implement a new GETINFO command to fetch consensus. Closes ticket 31684.
```
teor2345:
```suggestion
    - Implement a new GETINFO command to fetch the microdesc consensus. Closes ticket 31684.
```

src/feature/control/control_getinfo.h
```diff
@@ -48,6 +48,10 @@ STATIC int getinfo_helper_downloads(
     control_connection_t *control_conn,
     const char *question, char **answer,
     const char **errmsg);
+STATIC int getinfo_helper_current_consensus(
+    const char* consensus_type,
```
teor2345: Other code uses flavor_name
```suggestion
    const char* flavor_name,
```

src/feature/control/control_getinfo.c
```diff
@@ -325,6 +325,33 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
+ *   -1 in cases of an error
+ *   0 otherwise
+ */
+STATIC int
+getinfo_helper_current_consensus(const char* consensus_type,
```
teor2345: Other code uses flavor_name, please be consistent

src/feature/control/control_getinfo.c
```diff
@@ -325,6 +325,33 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
+                                 char** answer,
+                                 const char** errmsg)
+{
+  tor_mmap_t *mapped = networkstatus_map_cached_consensus(consensus_type);
+  if (mapped) {
+    *answer = tor_memdup_nulterm(mapped->data, mapped->size);
+    tor_munmap_file(mapped);
+    tor_free(mapped);
```
teor2345: Double-free, this line is not in the original code at:
https://github.com/torproject/tor/pull/1328/files#diff-ac2399586b806edb39db242c0a5664bbL589

src/feature/control/control_getinfo.c
```diff
@@ -325,6 +325,33 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
+ *
+ * returns: int
+ *   -1 in cases of an error
+ *   0 otherwise
+ */
+STATIC int
+getinfo_helper_current_consensus(const char* consensus_type,
+                                 char** answer,
+                                 const char** errmsg)
+{
```
teor2345: Why did you delete the code that checks for a consensus in memory in this refactor?
```
    if (we_want_to_fetch_flavor(get_options(), FLAV_NS)) {
      const cached_dir_t *consensus = dirserv_get_consensus("ns");
      if (consensus)
        *answer = tor_strdup(consensus->dir);
    }
    if (!*answer) { /* try loading it from disk */
```

ltbringer: I am checking for memory leaks in the code, I am sorry these were not meant for review. I shall have that code back once I understand what is causing the seg fault and memory leak on mingw and clang.

teor2345:
It seems like you're pushing a lot of commits here.
Can we help you get set up to run these tests locally?

Maybe this task isn't a good fir for you.
Would you like us to take over this refactor?

ltbringer:
I'll take it off on a separate branch and use the CI on my fork. I assumed I'd get this done faster but I should have done this before. Sincere apologies.

changes/ticket31684
```diff
@@ -1,2 +1,5 @@
   o Minor features (onion services):
-    - GETINFO dir/status-vote/current/consensus-microdesc dumps microdesc consensus. Closes ticket 31684.
+    - Implement a new GETINFO command to fetch microdesc consensus. Closes ticket 31684.
```
asn-d6: "fetch the microdescriptor consensus"

src/feature/control/control_getinfo.c
```diff
@@ -325,47 +325,32 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
 }
 
-/**
- * Switch between microdesc vs networkstatus descriptor dumps
- * Assumes that question would be either:
- * - dir/status-vote/current/consensus-microdesc
- * or
- * - dir/status-vote/current/consensus
- */
+/** Fetch network-status or micro-descriptor
```
asn-d6: "GETINFO helper for dumping different consensus flavors".

src/feature/control/control_getinfo.c
```diff
@@ -325,47 +325,32 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
+ * returns: 0 on success -1 on error. */
 STATIC int
-getinfo_helper_current_consensus(const char* question,
-                          char** answer,
-                          const char** errmsg)
+getinfo_helper_current_consensus(int flavor,
```
asn-d6: No need to pass both `flavor` and `flavor_name`. That's redundant.

Please pass just one, and derive the other in the function.

src/test/test_controller.c
```diff
@@ -1689,6 +1691,187 @@ test_download_status_bridge(void *arg)
   return;
 }
 
+/** Mock cached consensus */
+static cached_dir_t *mock_ns_consensus_cache;
+static cached_dir_t *mock_microdesc_consensus_cache;
+
+int
```
asn-d6: Why do you have the function prototype and the actual function separate? We don't do that. Please dump the function prototype and just document the function directly. Same for the ones below.

src/test/test_controller.c
```diff
@@ -1689,6 +1691,187 @@ test_download_status_bridge(void *arg)
...
+mock_networkstatus_get_cache_fname(int flav,
+                                   const char *flavor_name,
+                                   int unverified_consensus);
+
+char *
+mock_networkstatus_get_cache_fname(int flav,
+                                   const char *flavor_name,
+                                   int unverified_consensus)
+{
+  /** A function needs to do something with
```
asn-d6: You don't need to do useless stuff with variables. Instead just do `(void) flav` if you have no need for the argument.

src/test/test_controller.c
```diff
@@ -1689,6 +1691,187 @@ test_download_status_bridge(void *arg)
...
+mock_we_want_to_fetch_flavor(const or_options_t *options,
+                             int flavor) {
+  if (options->FetchUselessDescriptors ||
+      flavor == FLAV_NS || flavor == FLAV_MICRODESC) {
+    return 1;
+  }
+  return 0;
+}
+
+/**  Mock the function that retrieves consensus from
```
asn-d6: Seems like you are wrapping text at like 60 columns, whereas we usually wrap at 80. It seems even more weird than our already weird wrapping logic, so please adjust all your code to fit the norm.

src/test/test_controller.c
```diff
@@ -1689,6 +1691,187 @@ test_download_status_bridge(void *arg)
...
+  /* Get results out */
+  char *answer = NULL;
+  const char *errmsg = NULL;
+
+  (void)arg;
+
+  setup_bridge_mocks();
+  MOCK(tor_mmap_file, mock_tor_mmap_file);
+  MOCK(tor_munmap_file, mock_tor_munmap_file);
+  MOCK(networkstatus_get_cache_fname,
```
asn-d6: I don't think you need to mock `networkstatus_get_cache_fname()`. Just use the `get_options_mutable()` to change the options so that the function does not crash. See how other tests are doing it. e.g. `test_router_dump_router_to_string_no_bridge_distribution_method()`.

src/test/test_controller.c
```diff
@@ -1689,6 +1691,187 @@ test_download_status_bridge(void *arg)
...
+  /* We just need one of these to pass, it doesn't matter what's in it */
+  control_connection_t dummy;
+  /* Get results out */
+  char *answer = NULL;
+  const char *errmsg = NULL;
+
+  (void)arg;
+
+  setup_bridge_mocks();
+  MOCK(we_want_to_fetch_flavor, mock_we_want_to_fetch_flavor);
```
asn-d6: Why do you mock `we_want_to_fetch_flavor()`? This seems like an interesting function that we should know how to navigate, instead of black-boxing it. Please try to navigate it by using `or_options_t` as suggested above.

src/test/test_controller.c
```diff
@@ -1838,6 +2020,10 @@ struct testcase_t controller_tests[] = {
     NULL },
   { "download_status_consensus", test_download_status_consensus, 0, NULL,
     NULL },
+  {"getinfo_helper_dir_from_cache", test_getinfo_helper_dir_from_cache,
```
asn-d6: Please rename these tests to clarify that they test the consensus fetching function and not the whole `getinfo_helper_dir()`.

src/test/test_controller.c
ltbringer: These are not functions, these are variables that I create outside the test case so that I can free them. The original code:

```
  if (we_want_to_fetch_flavor(get_options(), flavor)) {
    /** Check from the cache */
    const cached_dir_t *consensus = dirserv_get_consensus(flavor_name);
    if (consensus) {
      *answer = tor_strdup(consensus->dir);
    }
  }
```
... doesn't free the consensus as it points to a global object. I keep these mock variables for the sake of avoiding memory leaks from within the test by freeing them after use.

src/test/test_controller.c
ltbringer: Thanks, I didn't know I could do that.

src/test/test_controller.c
ltbringer: I am trying to mock the parts of [this function](https://github.com/torproject/tor/blob/35cfe2e77679e1df5c3ae4b71112fa252acafdca/src/feature/nodelist/networkstatus.c#L245)

I don't see it using an `or_options_t` nor its callee `networkstatus_map_cached_consensus` ?

src/feature/control/control_getinfo.c
```diff
@@ -325,47 +325,37 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
- * or
- * - dir/status-vote/current/consensus
- */
+/** GETINFO helper for dumping different consensus flavors
+ * returns: 0 on success -1 on error. */
 STATIC int
-getinfo_helper_current_consensus(const char* question,
-                          char** answer,
-                          const char** errmsg)
+getinfo_helper_current_consensus(int flavor,
```
asn-d6: Instead of `int flavor` you can use the `consensus_flavor_t` enum.

src/feature/control/control_getinfo.c
```diff
@@ -325,47 +325,37 @@ getinfo_helper_current_time(control_connection_t *control_conn,
...
+                                 const char** errmsg)
 {
-  // Ensures question contains a request for
-  // ns or microdesc consensus
-  if (
-      strcmp(question, "dir/status-vote/current/consensus") &&
-      strcmp(question, "dir/status-vote/current/consensus-microdesc")) {
-    return 0;
+  const char *flavor_name = networkstatus_get_flavor_name(flavor);
+  if (!strcmp(flavor_name, "")) {
```
asn-d6: This strcmp with "" is no longer useful after using `networkstatus_get_flavor_name`.  At least change it to check for "??" or change the general check logic.

src/test/test_controller.c
```diff
@@ -1689,6 +1691,143 @@ test_download_status_bridge(void *arg)
...
+test_getinfo_helper_current_consensus_from_cache(void *arg)
+{
+  /* We just need one of these to pass, it doesn't matter what's in it */
+  control_connection_t dummy;
+  /* Get results out */
+  char *answer = NULL;
+  const char *errmsg = NULL;
+
+  (void)arg;
+  setup_bridge_mocks();
```
teor2345: Not needed

src/test/test_controller.c
```diff
@@ -1689,6 +1691,143 @@ test_download_status_bridge(void *arg)
...
+  getinfo_helper_dir(&dummy,
+                     "dir/status-vote/current/consensus-microdesc",
+                     &answer,
+                     &errmsg);
+  tt_str_op(answer, OP_EQ, "mock_microdesc_consensus");
+  tt_ptr_op(errmsg, OP_EQ, NULL);
+  tor_free(mock_microdesc_consensus_cache->dir);
+  tor_free(answer);
+  errmsg = NULL;
+  options->FetchUselessDescriptors = previous_fetch_value;
```
teor2345: Should be after done:

src/test/test_controller.c
```diff
@@ -1689,6 +1691,143 @@ test_download_status_bridge(void *arg)
...
+  /* We just need one of these to pass, it doesn't matter what's in it */
+  control_connection_t dummy;
+  /* Get results out */
+  char *answer = NULL;
+  const char *errmsg = NULL;
+
+  (void)arg;
+  setup_bridge_mocks();
+  or_options_t *options = get_options_mutable();
+  int previous_fetch_value = options->FetchUselessDescriptors;
```
teor2345: Not needed

src/test/test_controller.c
```diff
@@ -1689,6 +1691,143 @@ test_download_status_bridge(void *arg)
...
+                     &errmsg);
+  tt_str_op(answer, OP_EQ, "mock_microdesc_consensus");
+  tt_ptr_op(errmsg, OP_EQ, NULL);
+  tor_free(mock_microdesc_consensus_cache->dir);
+  tor_free(answer);
+  errmsg = NULL;
+  options->FetchUselessDescriptors = previous_fetch_value;
+
+ done:
+  clear_bridge_mocks();
```
teor2345: Not needed

teor2345:
Latest code in #1434.

PR #1338, opened 2019-09-18
Ticket30920

JMGuisadoG:
[Trac link](https://trac.torproject.org/projects/tor/ticket/30920)

Nowrap multiplication function has its own commit. That commit has no change file since I was not sure if it needed a changelog for users. If needed I can amend that commit.

### DIFF
--- a/src/lib/confmgt/unitparse.c
+++ b/src/lib/confmgt/unitparse.c
@@ -15,6 +15,7 @@
 #include "lib/log/util_bug.h"
 #include "lib/string/parse_int.h"
 #include "lib/string/util_string.h"
+#include "lib/intmath/muldiv.h"
 
 #include <string.h>
 
@@ -109,6 +110,7 @@ const struct unit_table_t time_msec_units[] = {
  * table <b>u</b>, then multiply the number by the unit multiplier.
  * On success, set *<b>ok</b> to 1 and return this product.
  * Otherwise, set *<b>ok</b> to 0.
+ * Warns user when overflow or a negative value is detected.
  */
 uint64_t
 config_parse_units(const char *val, const unit_table_t *u, int *ok)
@@ -142,10 +144,35 @@ config_parse_units(const char *val, const unit_table_t *u, int *ok)
 
   for ( ;u->unit;++u) {
     if (!strcasecmp(u->unit, cp)) {
-      if (use_float)
-        v = (uint64_t)(u->multiplier * d);
-      else
-        v *= u->multiplier;
+      if (use_float) {
+        d = u->multiplier * d;
+
+        if (d < 0) {
+          log_warn(LD_CONFIG, "Negative value arised when parsing %s %s",
+                   val, u->unit);
+          *ok = 0;
+          goto done;
+        }
+
+        // Some compilers may warn about casting a double to an unsigned type
+        // because they don't know if d is >= 0
+        if (d >= 0 && (d > (double)INT64_MAX || (uint64_t)d > INT64_MAX)) {
+          log_warn(LD_CONFIG, "Overflow detected parsing %s %s", val, u->unit);
+          *ok = 0;
+          goto done;
+        }
+
+        v = (uint64_t) d;
+      } else {
+        v = tor_mul_u64_nowrap(v, u->multiplier);
+
+        if (v > INT64_MAX) {
+          log_warn(LD_CONFIG, "Overflow detected parsing %s %s", val, u->unit);
+          *ok = 0;
+          goto done;
+        }
+      }
+
       *ok = 1;
       goto done;
     }
--- a/src/lib/intmath/muldiv.c
+++ b/src/lib/intmath/muldiv.c
@@ -69,6 +69,20 @@ gcd64(uint64_t a, uint64_t b)
   return a;
 }
 
+/** Return the unsigned integer product of <b>a</b> and <b>b</b>, if overflow
+ * is detected return UINT64_MAX instead. */
+uint64_t
+tor_mul_u64_nowrap(uint64_t a, uint64_t b)
+{
+  if (a == 0 || b == 0) {
+    return 0;
+  } else if (PREDICT_UNLIKELY(UINT64_MAX / a < b)) {
+    return UINT64_MAX;
+  } else {
+    return a*b;
+  }
+}
+
 /* Given a fraction *<b>numer</b> / *<b>denom</b>, simplify it.
  * Requires that the denominator is greater than 0. */
 void
--- a/src/test/test_confparse.c
+++ b/src/test/test_confparse.c
@@ -906,11 +906,22 @@ test_confparse_unitparse(void *args)
   tt_assert(ok);
 
   /* u64 overflow */
-  /* XXXX our implementation does not currently detect this. See bug 30920. */
-  /*
   tt_u64_op(config_parse_memunit("20000000 TB", &ok), OP_EQ, 0);
   tt_assert(!ok);
-  */
+  // This test fails the double check as the float representing 15000000.5 TB
+  // is greater than (double) INT64_MAX
+  tt_u64_op(config_parse_memunit("15000000.5 TB", &ok), OP_EQ, 0);
+  tt_assert(!ok);
+  // 8388608.1 TB passes double check because it falls in the same float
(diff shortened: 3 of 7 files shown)


src/lib/confmgt/unitparse.c
```diff
@@ -145,7 +146,14 @@ config_parse_units(const char *val, const unit_table_t *u, int *ok)
       if (use_float)
         v = (uint64_t)(u->multiplier * d);
       else
-        v *= u->multiplier;
+        v = tor_mul_u64_nowrap(v, u->multiplier);
+
+      if ( v > SSIZE_MAX ) {
```
JMGuisadoG: Would it be okay to check for `v > UINT64_MAX` instead of `SSIZE_MAX` ?

[Trac comment](https://trac.torproject.org/projects/tor/ticket/30920#comment:13) regarding using SSIZE_MAX instead of UINT64_MAX. Probably the difference between SSIZE_MAX in 64-bit machines and 32-bit machines are causing some tests to fail.

src/lib/confmgt/unitparse.c
```diff
@@ -142,10 +143,26 @@ config_parse_units(const char *val, const unit_table_t *u, int *ok)
...
   for ( ;u->unit;++u) {
     if (!strcasecmp(u->unit, cp)) {
-      if (use_float)
-        v = (uint64_t)(u->multiplier * d);
-      else
-        v *= u->multiplier;
+      if (use_float) {
+        d = u->multiplier * d;
+
+        if ( d > INT64_MAX ) {
```
teor2345: Let's make sure we're doing a floating point comparison here, using:
```suggestion
        if ( d > (double)INT64_MAX ) {
```

teor2345: See my updated suggestion, where I do both the integer and floating point checks.

src/test/test_confparse.c
```diff
@@ -906,11 +906,8 @@ test_confparse_unitparse(void *args)
   tt_assert(ok);
 
   /* u64 overflow */
-  /* XXXX our implementation does not currently detect this. See bug 30920. */
-  /*
   tt_u64_op(config_parse_memunit("20000000 TB", &ok), OP_EQ, 0);
```
teor2345: We'll also need tests for the floating-point case.
It would be great to have a test that fails the float comparison, and then one that passes the float comparison, but fails the int comparison.

JMGuisadoG: I've never dealt with float operations in such a context so I had to read about it a bit before. I'm pasting here what I found (by testing in a separate c file) to be a case for each failing comparison before submitting.

In this case the double comparison will fail
```
 // Fails double comparison
 tt_u64_op(config_parse_memunit("15000000.5 TB", &ok), OP_EQ, 0);
 tt_assert(!ok);
```

This case is a valid float number, (`INT64_MAX` - 0.5) but its integer representation is greater than INT64_MAX.
```
 // Fails int comparison
 tt_u64_op(config_parse_memunit("9223372036854775806.5 B", &ok), OP_EQ, 0);
 tt_assert(!ok);
```

src/lib/confmgt/unitparse.c
JMGuisadoG: Thank you! This helped learn a bit more about safety comparison.

I'll update with your suggestion when the tests are ok.

src/test/test_confparse.c
```diff
@@ -906,11 +906,14 @@ test_confparse_unitparse(void *args)
...
   tt_assert(!ok);
-  */
+  // Fails double comparison
+  tt_u64_op(config_parse_memunit("15000000.5 TB", &ok), OP_EQ, 0);
```
JMGuisadoG: This test fails the double check as the float representing 15000000.5 TB is greater than `(double) INT64_MAX`

src/lib/confmgt/unitparse.c
```diff
@@ -142,10 +143,26 @@ config_parse_units(const char *val, const unit_table_t *u, int *ok)
...
   for ( ;u->unit;++u) {
     if (!strcasecmp(u->unit, cp)) {
-      if (use_float)
-        v = (uint64_t)(u->multiplier * d);
-      else
-        v *= u->multiplier;
+      if (use_float) {
+        d = u->multiplier * d;
+
+        if ( d > (double)INT64_MAX || (uint64_t)d > INT64_MAX ) {
```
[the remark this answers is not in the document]
JMGuisadoG: Thank you for noting this Nick. Do you think a comment stating this would be necessary or would just bloat the code?

src/lib/confmgt/unitparse.c
```diff
@@ -142,10 +143,26 @@ config_parse_units(const char *val, const unit_table_t *u, int *ok)
 
   for ( ;u->unit;++u) {
     if (!strcasecmp(u->unit, cp)) {
-      if (use_float)
-        v = (uint64_t)(u->multiplier * d);
-      else
-        v *= u->multiplier;
+      if (use_float) {
```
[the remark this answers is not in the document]
JMGuisadoG: > We should update the documentation of this function to note its behavior on error.

I'm understanding you mean `config_parse_units`, right?

By updating documentation would a comment above the function suffice,? (Like the one you've commented before about the new `tor_mul_u64_nowrap`)

src/lib/intmath/muldiv.c
```diff
@@ -69,6 +69,18 @@ gcd64(uint64_t a, uint64_t b)
   return a;
 }
 
+uint64_t
```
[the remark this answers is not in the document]
JMGuisadoG: I could try giving it some documentation and accept any suggestions afterwards, which are likely given that I'm new to Tor contributions.

I can rewrite the commit for this function; or maybe, squashing this contribution into a single one is preferred?

src/test/test_confparse.c
teor2345: Can you put this information in the code?

src/test/test_confparse.c
```diff
@@ -906,11 +906,14 @@ test_confparse_unitparse(void *args)
...
+  tt_assert(!ok);
+  // Fails int comparison
+  tt_u64_op(config_parse_memunit("8388608.1 TB", &ok), OP_EQ, 0);
```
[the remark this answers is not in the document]
teor2345: Can you put this information in the code as a comment?

src/lib/confmgt/unitparse.c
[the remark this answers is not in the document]
teor2345: We want another test
```suggestion
        if ( d > (double)INT64_MAX || d < 0 || (uint64_t)d > INT64_MAX ) {
```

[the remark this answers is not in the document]
teor2345: But the error message is different, so it should be a separate if block.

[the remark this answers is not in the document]
JMGuisadoG: > But the error message is different, so it should be a separate if block.

Would it be ok to warn about "Negative or too big" in the same message so this way there's no need to add another if block?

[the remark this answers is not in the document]
teor2345: if blocks are cheap. But people's time is expensive, and dealing with confusing error messages takes time.

So just add a different message. The users will thank you :-)

JMGuisadoG:
I've updated the commits with your suggestions.

- Added documentation to the new nowrap mul function
- Added comments to the unitparse tests explaining each float representation case
- Added `d < 0` if-block and warn message in unitparse.c
- Added `d > 0` test inside float comparison if-block to avoid some compiler warnings in unitparse.c

src/lib/confmgt/unitparse.c
```diff
@@ -142,10 +144,33 @@ config_parse_units(const char *val, const unit_table_t *u, int *ok)
...
+        d = u->multiplier * d;
+
+        if (d < 0) {
+          log_warn(LD_CONFIG, "Negative value arised when parsing %s %s",
+                   val, u->unit);
```
teor2345: We also need to fail on negative values:
```suggestion
                   val, u->unit);
          *ok = 0;
           goto done;  
```

JMGuisadoG: Ouch! Sorry for missing this.

I will also add a negative double test so everything is covered.

PR #1339, opened 2019-09-18
Ticket31759: improvements to annotate_ifdef_directives

nmathewson:


scripts/maint/annotate_ifdef_directives
```diff
@@ -26,14 +26,33 @@ import re
...
     """
-    Return fmt%argument, for use as a commented line.
+    Return fmt%argument, for use as a commented line.  If the line would
+    be longer than maxwidth, truncate argument.
+
+    Requires that fmt%"..." will fit into maxwidth characters.
     """
-    return fmt % argument
+    result = fmt % argument
+    if len(result) <= maxwidth:
```
tlyu: I think this will allow a 80-character line. doc/HACKING/Codingstyle.md says 79 characters per line, and checkSpaces.pl seems to enforce 79.

teor2345: Should we have a test case for 79 and 80 character lines?
Or is this script too small to have test cases?

tlyu: > Should we have a test case for 79 and 80 character lines?
> Or is this script too small to have test cases?

I think that's a great idea to have a test case for 79 vs 80 character lines.

scripts/maint/annotate_ifdef_directives.py
```diff
@@ -0,0 +1,317 @@
...
+    'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa */\n'
+
+
+    """
+    assert fmt.endswith("\n")
+    result = fmt % argument
+    if len(result) <= maxwidth:
+        return result
+    else:
+        # How long can we let the agument be?  Try filling in the
```
tlyu: Typo: "agument" should be "argument"

PR #1346, opened 2019-09-20
lock: Don't call pthread_mutex_destroy() on a global mutex

teor2345:
Closes bug 31736.


src/app/config/config.c
```diff
@@ -1187,7 +1187,10 @@ init_protocol_warning_severity_level(void)
 static void
 cleanup_protocol_warning_severity_level(void)
 {
-   atomic_counter_destroy(&protocol_warning_severity_level);
+  /* Destroying a locked mutex is undefined behaviour. Therefore, global
+   * mutexes can not be destroyed, because multiple threads may still be able
+   * to access them. See #31735. */
+   //atomic_counter_destroy(&protocol_warning_severity_level);
```
nmathewson: Should we restore this to its original level of LOG_WARN?

PR #1349, opened 2019-09-23
Bug31408 master

teor2345:


changes/bug31408
```diff
@@ -0,0 +1,4 @@
+  o Major bugfixes (torrc):
+    - Stop ignoring torrc options after an %include directive, when the
+      included directory ends with an empty file. (Or a file only containing
```
Jigsaw52: I believe empty (0 byte) files will not cause this problem. Empty files are skipped when building the list of files for a %include directive (see config_get_file_list in conffile.c) and they do not go through the buggy code. Also, the file src/test/conf_examples/include_bug_31408/included/02_empty.inc added in this commit should have some content.

PR #1359, opened 2019-09-24
[#31810] Handle errors from execve() in the Unix process backend more gracefully.

ahf:
This patch removes a call to tor_assert_unreached() after execve()
failed. This assertion leads to the child process emitting a stack trace
on its output, which makes the error harder for the user to demystify,
since they think it is an internal error in Tor instead of "just" being
a "no such file or directory" error.

See: https://bugs.torproject.org/31810

### DIFF
--- a/src/lib/process/process_unix.c
+++ b/src/lib/process/process_unix.c
@@ -255,18 +255,13 @@ process_unix_exec(process_t *process)
     /* Call the requested program. */
     retval = execve(argv[0], argv, env->unixoid_environment_block);
 
-    /* If we made it here it is because execve failed :-( */
-    if (-1 == retval)
-      fprintf(stderr, "Call to execve() failed: %s", strerror(errno));
-
+    /* LCOV_EXCL_START */
     tor_free(argv);
     process_environment_free(env);
 
-    tor_assert_unreached();
-
  error:
-    /* LCOV_EXCL_START */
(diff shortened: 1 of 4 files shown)


src/lib/process/process_unix.c
```diff
@@ -255,18 +255,13 @@ process_unix_exec(process_t *process)
     /* Call the requested program. */
     retval = execve(argv[0], argv, env->unixoid_environment_block);
 
-    /* If we made it here it is because execve failed :-( */
-    if (-1 == retval)
-      fprintf(stderr, "Call to execve() failed: %s", strerror(errno));
-
+    /* LCOV_EXCL_START */
```
nmathewson: I think it isn't appropriate to have LCOV_EXCL_START here if we believe that this code can sometimes be reached. We should only use LCOV_EXCL for things that are untestable because they are truly unreachable.

src/lib/process/process_unix.c
```diff
@@ -255,18 +255,13 @@ process_unix_exec(process_t *process)
...
     tor_free(argv);
     process_environment_free(env);
 
-    tor_assert_unreached();
-
  error:
-    /* LCOV_EXCL_START */
-    fprintf(stderr, "Error from child process: %s", strerror(errno));
+    fprintf(stderr, "Error from child process: %s\n", strerror(errno));
```
nmathewson: This eventually gets logged, right?

PR #1380, opened 2019-09-30
Stop libevent from reading data from closed connections.

asn-d6:
Code adapted from Rob's proposed patch in #30344.

Also add a comment in connection_mark_for_close_internal_() on why we should
not be adding extra code there without a very good reason.

### DIFF
--- a/src/core/mainloop/connection.c
+++ b/src/core/mainloop/connection.c
@@ -907,8 +907,8 @@ connection_mark_for_close_(connection_t *conn, int line, const char *file)
  * controlling channel has been notified (e.g. with
  * connection_or_notify_error()), or you actually are the
  * connection_or_close_for_error() or connection_or_close_normally() function.
- * For all other cases, use connection_mark_and_flush() instead, which checks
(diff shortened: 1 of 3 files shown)


src/core/mainloop/connection.c
```diff
@@ -900,13 +900,19 @@ connection_mark_for_close_(connection_t *conn, int line, const char *file)
...
+ *
+ * This _internal version bypasses the CONN_TYPE_OR checks; this should be
+ * called when you either are sure that if this is an or_connection_t the
+ * controlling channel has been notified (e.g. with
+ * connection_or_notify_error()), or you actually are the
  * connection_or_close_for_error() or connection_or_close_normally() function.
- * For all other cases, use connection_mark_and_flush() instead, which
- * checks for or_connection_t properly, instead.  See below.
+ * For all other cases, use connection_mark_and_flush() instead, which checks
+ * for or_connection_t properly, instead.  See below.
```
teor2345: Repeated "instead"

PR #1389, opened 2019-10-02
feat(control): add details to circuit-info

aveuiller:


teor2345:
We had some CI issues on Appveyor over the weekend, I fixed them in master. Your next build shouldn't have those failures.

aveuiller:
Thanks! I rebased on master. However, I don't really understand why `make check` or `make distcheck` seems to fail on some travis environments.

If I understand correctly the test logs are only available in `test-suite.log`, do we have a way to take a look at them?

src/feature/control/control_fmt.c
```diff
@@ -165,6 +171,74 @@ circuit_describe_status_for_controller(origin_circuit_t *circ)
...
+    SMARTLIST_FOREACH(circ->prepend_policy, addr_policy_t *, policy, {
+      smartlist_add(prepend_policy, addr_policy_to_controller_string(policy));
+    });
+  }
+  if (smartlist_len(prepend_policy)) {
+    prepend_policy_joined = smartlist_join_strings(prepend_policy, ",",
+                                                   0, NULL);
+    smartlist_add_asprintf(descparts, "PREPEND_POLICY=%s",
+                           prepend_policy_joined);
+  }
```
ahf: You leak the `prepend_policy` smartlist here I think?

src/feature/control/control_fmt.c
```diff
@@ -165,6 +171,74 @@ circuit_describe_status_for_controller(origin_circuit_t *circ)
...
+                         circ->remaining_relay_early_cells);
+  smartlist_add_asprintf(descparts, "PADDING_NEGOTIATION_FAILED=%u",
+                         circ->padding_negotiation_failed);
+  smartlist_add_asprintf(descparts, "PATH_STATE=%s",
+                         pathbias_state_to_controller_string(circ->path_state)
+                         );
+  smartlist_add_asprintf(descparts, "HAS_RELAXED_TIMEOUT=%u",
+                         circ->relaxed_timeout);
+  smartlist_add_asprintf(descparts, "IS_ANCIENT=%u", circ->is_ancient);
+  smartlist_add_asprintf(descparts, "HAS_OPENED=%u", circ->has_opened);
```
ahf: Should %u really be used as a format string here for these boolean types? It might be we do that elsewhere and it's fine, but it just caught my eyes as being odd. %u can have a lot of different values other than 0 or 1, but the spec also doesn't allow us to use the "true" and "false" literal.

src/feature/control/control_fmt.c
```diff
@@ -165,6 +171,74 @@ circuit_describe_status_for_controller(origin_circuit_t *circ)
...
+  smartlist_add_asprintf(descparts, "HS_HAS_TIMED_OUT=%u",
+                         circ->hs_circ_has_timed_out);
+
+  // Prepend policy data
+  const char *prepend_policy_joined;
```
ahf: `NULL` initialize?

src/feature/control/control_fmt.c
aveuiller: Indeed, I added a free

src/feature/control/control_fmt.c
aveuiller: The variable is now initialized every time, thanks for the comment.

src/feature/control/control_fmt.c
aveuiller: I have to admit that I do not really know how to print those values otherwise. They are defined as unsigned int, so I printed them as such, even if they are mostly defined on 1 bit.

I tried to look in the project, but I could not see another boolean output definition. Do you have something in mind to use instead of `%u`?

aveuiller: Hello,

I'm coming back to you on this part. I personally don't feel like it's an issue to keep `%u` since the fields are effectively `unsigned int` limited to 1 bit (Cf. https://github.com/torproject/tor/blob/master/src/core/or/origin_circuit_st.h#L151-L156).

However, if you prefer using `true` and `false` literals, I can change the specs on this part since it hasn't been merged already.

What would you prefer? :)

PR #1393, opened 2019-10-03
Ticket31841

nmathewson:


src/test/resolve_test_helpers.c
```diff
@@ -0,0 +1,87 @@
...
+      } else {
+        int s = tor_addr_parse(addr, entries[i].ipv4);
+        tt_int_op(s, OP_EQ, AF_INET);
+      }
+      r = 0;
+      break;
+    }
+  }
+
+#if 0
```
teor2345: Can you delete this block, or turn it into a debug log?

src/test/test_addr.c
```diff
@@ -1169,6 +1171,8 @@ test_addr_parse(void *arg)
 
   (void)arg;
 
+  mock_hostname_resolver();
```
teor2345: I think the changes in the tests are meant to be in the final commit. But they are actually in the second commit.

PR #1398, opened 2019-10-07
[0.4.0] #31810 execve() error handling

ahf:
These patches makes execve() error handling slightly better on Unix and adds a test case that made us notice that we did not run the exit callback when CreateProcessA() fails on Windows.


ahf:
Wrong target branch...

PR #1402, opened 2019-10-09
Ticket29698 042 01

dgoulet-tor:


src/core/or/circuitmux.c
```diff
@@ -1335,3 +1335,20 @@ circuitmux_compare_muxes, (circuitmux_t *cmux_1, circuitmux_t *cmux_2))
   }
 }
 
+/**
+ * Touch the circuitmux so its policy can be updated at this point in time.
+ */
+void
+circuitmux_touch(circuitmux_t *cmux)
```
nmathewson: Would a better name for "touch" be "update policy", "recalculate priority", "calculate_initial_priority", or something similar?

src/core/or/circuitmux.c
```diff
@@ -1335,3 +1335,20 @@ circuitmux_compare_muxes, (circuitmux_t *cmux_1, circuitmux_t *cmux_2))
   }
 }
 
+/**
+ * Touch the circuitmux so its policy can be updated at this point in time.
```
nmathewson: This comment should have more detail to explain what "update" and "touch" mean.

src/core/or/circuitmux_ewma.c
```diff
@@ -216,6 +216,9 @@ static int
 ewma_cmp_cmux(circuitmux_t *cmux_1, circuitmux_policy_data_t *pol_data_1,
               circuitmux_t *cmux_2, circuitmux_policy_data_t *pol_data_2);
 
+static void
+ewma_touch(circuitmux_policy_data_t *pol_data);
```
nmathewson: We should not have a newline between the return type and the function name when we are declaring the function: only when we are defining it.

src/test/test_scheduler.c
```diff
@@ -1265,17 +1265,44 @@ test_scheduler_kist_pending_list(void *arg)
   UNMOCK(channel_should_write_to_kernel);
 }
 
+static void *
```
nmathewson: Let's use a comment to explain why this is necessary.

PR #1408, opened 2019-10-15
Ticket31882  move Android build config into core tor 

eighthave:
Guardian Project has maintained a wrapper build system for building tor for Android since the beginning. Since many more people are now building tor for Android, I'm working to upstream as much of that as possible. The goal is that the requirements for building on Android can be fulfilled by tor's own configure.ac.

full discussion here:
https://trac.torproject.org/projects/tor/ticket/31882

The GitLab-CI jobs are run here to show that `./configure` still works everywhere:
https://gitlab.com/eighthave/tor/pipelines/88974591

The extended _.gitlab-ci.yml_ additions can easily be handled separately from this pull request, as needed.

@n8fr8 @grote @akwizgran @yaronyg

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -97,6 +97,20 @@ else
             [Defined if we're building with OpenSSL or LibreSSL])
 fi
 
+dnl Enable minimal, standard Android build
+AC_ARG_ENABLE(android,
+     AS_HELP_STRING(--enable-android, [setup build for Android exeutable/shared library]))
+
+if test "x$enable_android" = "xyes"; then
+  asciidoc=false
+  enable_html_manual=no
+  enable_manpage=no
+  enable_pic=yes
+  enable_system_torrc=no
+  enable_tool_name_check=no
+  have_systemd=no
+fi
+
 if test "$enable_static_tor" = "yes"; then
   enable_static_libevent="yes";
   enable_static_openssl="yes";
@@ -126,7 +140,7 @@ AC_ARG_ENABLE(asciidoc,
         "yes") asciidoc=true ;;
         "no")  asciidoc=false ;;
         *) AC_MSG_ERROR(bad value for --disable-asciidoc) ;;
-      esac], [asciidoc=true])
+      esac], [test "x$asciidoc" = "xtrue" && asciidoc=true])
 
 # systemd notify support
 AC_ARG_ENABLE(systemd,
@@ -236,19 +250,12 @@ if test x$enable_event_tracing_debug = xyes; then
   AC_DEFINE([TOR_EVENT_TRACING_ENABLED], [1], [Compile the event tracing instrumentation])
 fi
 
-dnl Enable Android only features.
-AC_ARG_ENABLE(android,
-     AS_HELP_STRING(--enable-android, [build with Android features enabled]))
-AM_CONDITIONAL([USE_ANDROID], [test "x$enable_android" = "xyes"])
-
-if test "x$enable_android" = "xyes"; then
-  AC_DEFINE([USE_ANDROID], [1], [Compile with Android specific features enabled])
-
-  dnl Check if the Android log library is available.
-  AC_CHECK_HEADERS([android/log.h])
-  AC_SEARCH_LIBS(__android_log_write, [log])
-
-fi
+dnl Enable Android only standard libs
+case "$host" in
+    *android*)
+	AC_CHECK_LIB([log], [__android_log_write], [LIBS="-llog $LIBS"])
+	;;
+esac
 
 dnl ---
 dnl Tor modules options. These options are namespaced with --disable-module-XXX
--- a/src/app/config/config.c
+++ b/src/app/config/config.c
@@ -5805,14 +5805,14 @@ options_init_logs(const or_options_t *old_options, or_options_t *options,
       }
 
       if (!strcasecmp(smartlist_get(elts, 0), "android")) {
-#ifdef HAVE_ANDROID_LOG_H
+#ifdef __ANDROID__
         if (!validate_only) {
           add_android_log(severity, options->AndroidIdentityTag);
         }
 #else
         log_warn(LD_CONFIG, "Android logging is not supported"
                             " on this system. Sorry.");
-#endif /* defined(HAVE_ANDROID_LOG_H) */
+#endif /* defined(__ANDROID__) */
         goto cleanup;
       }
     }
--- a/changes/ticket31882
+++ b/changes/ticket31882
@@ -1,2 +1,7 @@
   o Minor feature (configure, build system):
     - The standard __ANDROID__ macro is now used for platform detection.
+    - --enable-android provides the standard, minimal build configuration
+      for creating Tor binaries to run either as a shared library or a daemon.
+      It enables PIC, disables system torrc, systemd, and all doc generation.
+      This flag is not required to build for Android.
+      Closes ticket 31882.
--- a/changes/ticket31882
+++ b/changes/ticket31882
@@ -1,0 +1,2 @@
+  o Minor feature (configure, build system):
+    - The standard __ANDROID__ macro is now used for platform detection.
--- a/doc/HACKING/android/Simpleperf.md
+++ b/doc/HACKING/android/Simpleperf.md
@@ -85,8 +85,7 @@ was spend on the call.
   found in the `app_data` directory. The `torrc` can be found in the `app_bin/`
   directory.
 
-- You can enable logging in Tor via the syslog (or android) log
-  mechanism with:
+- You can enable logging to Android's logcat using the syslog mechanism:
 
       $ adb shell
       (device):/ $ run-as org.torproject.android
--- a/doc/tor.1.txt
+++ b/doc/tor.1.txt
@@ -661,9 +661,10 @@ GENERAL OPTIONS
     every NUM seconds on open connections that are in use. (Default: 5 minutes)
 
 [[Log]] **Log** __minSeverity__[-__maxSeverity__] **stderr**|**stdout**|**syslog**::
+
     Send all messages between __minSeverity__ and __maxSeverity__ to the standard
-    output stream, the standard error stream, or to the system log. (The
-    "syslog" value is only supported on Unix.) Recognized severity levels are
+    output stream, the standard error stream, or to the system log. (The "syslog"
+    value is not supported Windows or iOS). Recognized severity levels are
     debug, info, notice, warn, and err. We advise using "notice" in most cases,
     since anything more verbose may provide sensitive information to an
     attacker who obtains the logs. If only one severity level is given, all
@@ -675,8 +676,13 @@ GENERAL OPTIONS
     Signal-safe logs are always sent to stderr or stdout. They are also sent to
     a limited number of log files that are configured to log messages at error
     severity from the bug or general domains. They are never sent as syslogs,
-    android logs, control port log events, or to any API-based log
-    destinations.
+    Android logcat messages, control port log events, or to any API-based log
+    destinations. +
+ +
+    If tor is running as a shared library on Android, only "syslog" is available.
+    Native code have "stderr" and "stdout" piped to /dev/null by the Android OS.
+    Early errors like port conflicts with **ControlPort**, **SocksPort**, etc. will
+    not be logged, since the logging is added later.
 
 [[Log2]] **Log** __minSeverity__[-__maxSeverity__] **file** __FILENAME__::
     As above, but send log messages to the listed filename. The
--- a/src/lib/log/log.c
+++ b/src/lib/log/log.c
@@ -51,9 +51,9 @@
 #include "lib/fdio/fdio.h"
 #include "lib/cc/ctassert.h"
 
-#ifdef HAVE_ANDROID_LOG_H
+#ifdef __ANDROID__
 #include <android/log.h>
-#endif // HAVE_ANDROID_LOG_H.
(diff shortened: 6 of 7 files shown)


eighthave:
At @sisbell and GeKo's request, I added a gitlab-ci job to run this build with the Android NDK r17b, which is what Tor Browser currently uses.  Here's that full run:
https://gitlab.com/eighthave/tor/-/jobs/322526878

configure.ac
```diff
@@ -95,6 +95,22 @@ else
             [Defined if we're building with OpenSSL or LibreSSL])
 fi
 
+dnl Enable Android only features.
+AC_ARG_ENABLE(android,
+     AS_HELP_STRING(--enable-android, [setup build for Android exeutable/shared library]))
+
+if test "x$enable_android" = "xyes"; then
+  asciidoc=false
```
nmathewson: I'm okay with overriding these defaults, but we need to test to make sure that the user can override them if they want. So I think we need to make sure (and document) that this comes before any of the other options that it adjusts.  Does it?

src/app/config/config.c
```diff
@@ -5734,7 +5734,7 @@ options_init_logs(const or_options_t *old_options, or_options_t *options,
         goto cleanup;
       }
 
-      if (!strcasecmp(smartlist_get(elts, 0), "android")) {
+      if (!strcasecmp(smartlist_get(elts, 0), "logcat")) {
```
nmathewson: I'm happy to accept that "logcat" is a better name here, but it isn't okay to break backward compatibility with everybody who was using the old name.  If we want to make this change, we should keep "android" as a deprecated synonym.

Additionally this should probably be documented someplace.

configure.ac
```diff
@@ -95,6 +95,22 @@ else
...
+  enable_html_manual=no
+  enable_manpage=no
+  enable_pic=yes
+  enable_system_torrc=no
+  enable_tool_name_check=no
+  have_systemd=no
+
+  LIBS="-llog $LIBS"
```
nmathewson: Should the user have to build with "--enable-android" to get `-llog` ?  Everywhere else, we are using `__ANDROID__` to detect that library, so if the user forgets to say `--enable-android` they won't get it.

IMO we should either automatically include the logging library if building on android, or we should omit it when not using --enable-android.

src/app/config/config.c
```diff
@@ -5735,14 +5735,14 @@ options_init_logs(const or_options_t *old_options, or_options_t *options,
       }
 
       if (!strcasecmp(smartlist_get(elts, 0), "android")) {
-#ifdef HAVE_ANDROID_LOG_H
+#ifdef __ANDROID__
```
nmathewson: Since we're no longer using HAVE_ANDROID_LOG_H, let's remove the code to generate it from configure.ac.

configure.ac
eighthave: If someone wants to customize all those things, they can omit `--enable-android`, then specify them one at a time.

After talking with Orbot, Briar, Thali, Tor Browser, and iOS OnionBrowser developers on the build config topic, and looking at the relevant bits in the code of those projects, I can see no real use case for tweaking the `--enable-android` flag.  This isn't like UNIX, where there are different packages and libraries installed.  Tor on Android and iOS it built by the native wrappers using `--enable-android`. 

There is never going to be systemd or system-torrc on Android or iOS, nor a reason to install that documentation for those platforms.  Android requires position-independent code to even run, and will soon only allow shared libraries embedded in APKs.

So long story short, I didn't try to make --enable-android customizable, or test whether it is.

src/app/config/config.c
eighthave: I could find no evidence that the "android" name was in use in a shipped app.  I looked in Orbot, Briar, and Tor Browser.  Seemed cleaner to just drop the old name and move on.  Do you know somewhere where it is used?

src/app/config/config.c
eighthave: This pull request does remove all traces that I could find.  What did I miss?

configure.ac
eighthave: Makes sense, how do you want to do that?  I guess the `android/log.h` header check could be repurposed to do that.  My autotools is very rusty these days.

eighthave: I guess something like this would make more sense:
```
AC_CHECK_LIB([log], [__android_log_write], [LIBS="-llog $LIBS"])
```

configure.ac
teor2345: Why is it called --enable-android if it is used for iOS as well?
We should use a more generic name, or create an alias for iOS.

src/app/config/config.c
eighthave: I started writing the docs for this, and then did some tests.  I've changed my mind.  I think the "android" option should just be removed.  syslog works well on Android.
https://trac.torproject.org/projects/tor/ticket/32181

Since the `android` log option was never documented, and it wasn't used by the well known Tor apps.  It always seemed to me to be a private/alpha API, so better to just remove the cruft, perhaps there could be an error if people use it?  I'm going to update the docs in this pull request, and the changes file to point Android users to syslog.

configure.ac
eighthave: Sorry for the confusion there, this patch is purely about Android.  But I've been looking at iOS at the same time, and working to combine the Android and iOS setups as much as it makes sense.  I think in the end, we might end up with really similar setups with only a few quirks for the differences.

The Android and iOS builds currently are using a whole collection of custom flags.  That means that widely used build setups are unknown to the core Tor developers (unless you go digging for them).   My goal here is to get the main Android and iOS setups to build using a collective flag like --enable-android, so that the core Tor developers can add configurations there when they are ready to be used by the downstream projects (Tor Browser, Orbot, iCepa, OnionBrowser, etc).  For example, when code is split out into a module, like dirauth, and it is ready for mobile configurations, then that configure variable can be included in the Android and iOS build setups.

For example, iCepa is the mostly commonly used iOS Tor framework, here is how it calls ./configure: https://github.com/iCepa/Tor.framework/blob/30480101b3bd2166baced913347fc8bb7392b3da/Tor/tor.sh#L139

_tor-android-binary_ is the only Tor build that I know of that is currently included in shipping apps, here's how it calls configure:
https://github.com/guardianproject/tor-android/blob/0830c87a29ea197ecd9a0b5aee7d9a4fb88dce25/external/Makefile#L271

After this patch, it can be reduced to:
{{{
			./configure \
				--host=$(ALTHOST) \
				--enable-android \
				--enable-lzma \
				--enable-restart-debugging \
				--enable-static-libevent --with-libevent-dir=$(EXTERNAL_ROOT) \
				--enable-static-openssl --with-openssl-dir=$(EXTERNAL_ROOT) \
				--enable-zstd \
				--prefix=$(EXTERNAL_ROOT)
}}}

src/app/config/config.c
eighthave: I'll remove this change from this pull request.

src/app/config/config.c
nmathewson: Never mind, you're right. I missed the part where you removed `AC_CHECK_HEADERS`

configure.ac
nmathewson: That would be great.

configure.ac
nmathewson: > So long story short, I didn't try to make --enable-android customizable, or test whether it is.

I still can't agree with this.  This is not just about a change for android; this is about a change to the semantics of the configure.ac file, which will break the semantics of several other options unless we fix it.  It's never a great idea to break options silently, even if we're fairly sure we don't want to use them.

If you're sure that the options will never need change, then perhaps instead of --enable-android, it would make sense to have a "configure_for_android.sh" script in contrib/android/ or something?

Otherwise the fix is to move the "enable_android" check so that it happens before any of the other options that it adjusts.

src/app/config/config.c
nmathewson: sounds good.

configure.ac
eighthave: If you can point to something this breaks on non-Android platforms, I'll fix that.  Tor's ./configure has lots of issues when compiling for mobile, see some of the issues I've filed. And I've been told that the [pkg-config bug won't be fixed](https://trac.torproject.org/projects/tor/ticket/32191#comment:5), that affects the recommended Tor Android build config that is what all the apps are using.  Yet we are now talking about fixing hypothetical configurations.

The point of --enable-android is to provide a build that is as familiar to an Android developer as possible.  Android developers basically never use autotools, the majority are ignorant to the existence of autotools.  `ndk-build` and `cmake` are what's used, and these include automatic configurations like what's provided by --enable-android.   PIC/PIE is a [hard requirement](https://stackoverflow.com/questions/33659673/android-ndk-are-fpic-and-pie-mututally-exclusive), the code will not run without it. Android builds don't build HTML documentation, I've never seen an Android app build and include man pages, so failing on asciidoc will only confuse people.  "system torrc" is a UNIX concept, it does not belong in anyway on Android.   I'm also working on building Tor into Calyx's Android ROM, so I'm intimately familiar with what Android with tor provided by the system looks like.

See the configure lines I referenced that are currently used in the binaries shipped in Tor Browser, Orbot, Briar, Thali, Onion Browser, iCepa, etc.  And autotools have always been a nightmare for Android builds.  autotools is so centered on UNIX, and it really falls down when things are different.  (I've been working with autotools for 20 years).  So yes, this is weird when thinking in terms of UNIX and autotools, but it is an big improvement on the status quo.

Making Android builds is currently only possible with a brittle and elaborate wrapper.  I've gotten real Android builds working without the wrapper script in this patch for the first time.  I have been working to get Guardian Project out of the business of providing the official build system for Tor on Android.  Tor Browser uses our build system.  Integrations like this have to happen to achieve this.  And that will only happen if the tor maintainers are willing to accept that cross-compiling for mobile is really different than building on UNIX and Windows.

teor2345: **Other Tickets**

> And I've been told that the pkg-config bug won't be fixed, that affects the recommended Tor Android build config that is what all the apps are using.

I was asking questions on that ticket to try and work out how to fix the bug. Now that we know how to fix the issue, I have suggested a way forward on that ticket.

If there are issues with my suggestion, please reply on that ticket.

**This Pull Request**

> > Otherwise the fix is to move the "enable_android" check so that it happens before any of the other options that it adjusts.

@eighthave, what's wrong with this suggested fix?

As far as I understand, moving the check to the start will still set all the defaults that are required for Android. But if one of those defaults is wrong for a particular platform or build, then the developers will be able to change it.

Can you give a specific example where this solution will not work?

**Next Steps**

Can we please try to focus on the specific changes that we need to move forward on this pull request?

Tickets and pull requests aren't a great place for long-form discussion. If you want to talk about the broader context, and how tor can support mobile better, please write an email to the tor-dev list.

We're trying to help. But we do have engineering and compatibility constraints. If you want to change those constraints, let's talk about that more on the mailing list.

eighthave: I'm happy to focus on specifics. If it is enough to just move this block:
```diff
+dnl Enable minimal, standard Android build
+AC_ARG_ENABLE(android,
+     AS_HELP_STRING(--enable-android, [setup build for Android exeutable/shared library]))
+
+if test "x$enable_android" = "xyes"; then
+  asciidoc=false
+  enable_html_manual=no
+  enable_manpage=no
+  enable_pic=yes
+  enable_system_torrc=no
+  enable_tool_name_check=no
+  have_systemd=no
+fi
+
```

Right after `AC_CANONICAL_HOST`, then it could also enable a hard error when pkg-config is missing.  Otherwise, right before `AC_ARG_ENABLE(openbsd-malloc,` should work in theory.  It would need to be tested.

eighthave: Sorry if I'm grumpy in all this, all this fighting with cross-compiling, Android emulators, autotools vs NDK, etc. is driving me a little crazy.  But I can see the light at the end of the tunnel, though not always clearly...

teor2345: > it could also enable a hard error when pkg-config is missing

That's not how pkg-config works. It's an optional dependency that automatically sets some variables from *.pc files. But the user can always set those variables manually.

eighthave: pkg-config is currently not optional for the standard tor build on Android, e.g. with `--enable-lzma`, and in the case where pkg-config is missing, `./configure` will not exit with an error.  And for big wrapper scripts like are required for Android and iOS, that means the giant build process will complete but the resulting tor binary will not have lzma.  I fine with pkg-config being optional.  I still think it makes sense to skip pkg-config entirely for Android builds, and perhaps iOS, and just use --prefix to find dependencies.

teor2345: I can see that this lzma issue really concerns you. But I'm struggling to understand what you need us to change. Can you open another ticket with your request? I'm a bit confused by all the different topics in this one comment thread.

In any case, if you can get the right CFLAGS and LDFLAGS without pkg-config, then go for it! You can avoid installing pkg-config on android, or set all of its env vars to empty directories. And then you should be fine. No need to modify configure at all :-)

eighthave: so back to --enable-android, it needs to be something that stands out from the regular ./configure flags because how else can there be checks for requirements like `--host-` and cross-compiling?  The only other option I can see is basing all this stuff on:
```bash
   case "$host" in
       *-linux-android*) 
       # do checks here
       ;;
  esac
```

eighthave:
We're going to maintain the Android and iOS build systems in this fork:
* https://github.com/guardianproject/tor
* https://gitlab.com/guardianproject/tor (gitlab mirror)

Please take anything from there as you see fit, under the license you require.

PR #1418, opened 2019-10-17
Subsystem thread initialization and cleanup in all threads

stevenengler:
This fixes [ticket \#32103](https://trac.torproject.org/projects/tor/ticket/32103). It extends the threadpool to accept a custom thread spawn function. This allows us to use our own spawn function which calls `subsystems_thread_cleanup` and the new `subsystems_thread_init`. It also calls `spawn_exit`, which negates the need for #1412. Finally the above two functions are called in the main thread before `subsystems_shutdown` and after `subsystems_init` respectively.

### DIFF
--- a/src/app/main/subsysmgr.c
+++ b/src/app/main/subsysmgr.c
@@ -231,6 +231,27 @@ subsystems_postfork(void)
 }
 
 /**
+ * Run thread-init code on all subsystems that declare any
+ **/
+void
+subsystems_thread_init(void)
+{
+  check_and_setup();
+
+  for (int i = (int)n_tor_subsystems - 1; i >= 0; --i) {
+    const subsys_fns_t *sys = tor_subsystems[i];
+    if (!sys->supported)
+      continue;
+    if (! sys_initialized[i])
+      continue;
+    if (sys->thread_init) {
+      log_debug(LD_GENERAL, "Thread init: %s", sys->name);
+      sys->thread_init();
+    }
+  }
+}
+
+/**
  * Run thread-cleanup code on all subsystems that declare any
  **/
 void
--- a/src/app/main/subsysmgr.c
+++ b/src/app/main/subsysmgr.c
@@ -238,7 +238,7 @@ subsystems_thread_init(void)
 {
   check_and_setup();
 
-  for (int i = (int)n_tor_subsystems - 1; i >= 0; --i) {
+  for (unsigned i = 0; i < n_tor_subsystems; ++i) {
     const subsys_fns_t *sys = tor_subsystems[i];
     if (!sys->supported)
       continue;
--- a/src/app/main/tor_threads.c
+++ b/src/app/main/tor_threads.c
@@ -1,0 +1,43 @@
+/* Copyright (c) 2003-2004, Roger Dingledine
+ * Copyright (c) 2004-2006, Roger Dingledine, Nick Mathewson.
+ * Copyright (c) 2007-2019, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+#include "app/main/tor_threads.h"
+
+#include "lib/thread/threads.h"
+#include "app/main/subsysmgr.h"
+
+/** Wraps a void (*)(void*) function and its argument so we can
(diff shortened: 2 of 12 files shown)


src/app/main/subsysmgr.c
```diff
@@ -230,6 +230,27 @@ subsystems_postfork(void)
...
 
+/**
+ * Run thread-init code on all subsystems that declare any
+ **/
+void
+subsystems_thread_init(void)
+{
+  check_and_setup();
+
+  for (int i = (int)n_tor_subsystems - 1; i >= 0; --i) {
```
teor2345: Do we want to start at the lowest subsystem first?

stevenengler: Oh no... Good catch, that's definitely a copy-paste error :( I'll wait and see if any other mistakes are found, and I'll fix this to follow `subsystems_init_upto`, so `for (unsigned i = 0; i < n_tor_subsystems; ++i)`. Thanks!

src/lib/evloop/workqueue.c
```diff
@@ -85,6 +85,9 @@ struct threadpool_s {
   void *(*new_thread_state_fn)(void*);
   void (*free_thread_state_fn)(void*);
   void *new_thread_state_arg;
+
+  /** Function to start a thread. Should return a negative number on error. */
+  int (*thread_spawn_fn)(void (*func)(void *), void *data);
```
nmathewson: I think it might be convenient to have a typedef for this function pointer type; the documentation should also explain what the arguments mean.

src/lib/subsys/subsys.h
```diff
@@ -70,6 +70,12 @@ typedef struct subsys_fns_t {
    */
   void (*postfork)(void);
 
+  /**
+   * Initialize any thread-local resources held by this subsystem. Called
+   * after the subsystem's global components are initialized.
```
nmathewson: Is the idea that this should be invoked on every new thread that we spawn? If so we should document it.

src/app/main/tor_threads.h
```diff
@@ -0,0 +1,11 @@
+/* Copyright (c) 2003-2004, Roger Dingledine
+ * Copyright (c) 2004-2006, Roger Dingledine, Nick Mathewson.
+ * Copyright (c) 2007-2019, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
```
nmathewson: Every file needs a doxygen block with at `@file` and `@brief` declaration.

src/app/main/tor_threads.c
```diff
@@ -0,0 +1,43 @@
+/* Copyright (c) 2003-2004, Roger Dingledine
+ * Copyright (c) 2004-2006, Roger Dingledine, Nick Mathewson.
+ * Copyright (c) 2007-2019, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
```
nmathewson: Every file needs a doxygen block with at `@file` and `@brief` declaration.

src/app/main/tor_threads.c
```diff
@@ -0,0 +1,43 @@
...
+ * invoke them with proper subsystem thread management.
+ */
+typedef struct tor_thread_helper_info_t {
+  void (*func)(void *);
+  void *data;
+} tor_thread_helper_info_t;
+
+static void tor_thread_helper(void *_info) ATTR_NORETURN;
+
+static void
```
nmathewson: We need to have documentation for this function, and for the next one as well.

src/app/main/tor_threads.c
```diff
@@ -0,0 +1,43 @@
...
+tor_thread_helper(void *_info)
+{
+  tor_thread_helper_info_t *info = _info;
+  void (*func)(void*) = info->func;
+  void *data = info->data;
+  tor_free(_info);
+
+  subsystems_thread_init();
+  func(data);
```
nmathewson: Do we need to document somewhere that "func" must exit by returning, and that it must not call `spawn_exit()`?

Alternatively, should we fix `spawn_exit()` so that it is guaranteed to call subsystems_thread_cleanup?

PR #1421, opened 2019-10-17
Ticket32123

teor2345:


src/feature/relay/routermode.h
```diff
@@ -20,4 +22,20 @@ MOCK_DECL(int, advertised_server_mode, (void));
 
 void set_server_advertised(int s);
 
+#define have_module_relay() (1)
```
nmathewson: This needs a doxygen comment, since it isn't advertised elsewhere.

src/feature/relay/router.h
```diff
@@ -28,7 +28,13 @@ struct ed25519_keypair_t;
 MOCK_DECL(crypto_pk_t *,get_onion_key,(void));
 time_t get_onion_key_set_at(void);
 void set_server_identity_key(crypto_pk_t *k);
+/* Some compilers are clever enough to know that when relay mode is disabled,
+ * this function always asserts. */
+#ifdef HAVE_MODULE_RELAY
 MOCK_DECL(crypto_pk_t *,get_server_identity_key,(void));
+#else
+#define get_server_identity_key() (tor_assert_unreached(),NULL)
```
nmathewson: This isn't valid; tor_assert_unreached is a statement, not an expression.

src/feature/relay/router.c
```diff
@@ -372,6 +372,8 @@ assert_identity_keys_ok(void)
   }
 }
 
+#ifdef HAVE_MODULE_RELAY
```
nmathewson: This is okay for now, though eventually we should make this whole file relay-only.

src/test/test_parseconf.sh
```diff
@@ -125,12 +128,30 @@ for dir in "${EXAMPLEDIR}"/*; do
...
-    if test -f "./expected"; then
-        if test -f "./error"; then
-            echo "FAIL: Found both ${dir}/expected and ${dir}/error."
-            echo "(Only one of these files should exist.)"
-            exit $EXITCODE
+    # If tor has some modules disabled, search for a custom result file for
+    # the disabled modules
+    for prefix in "$TOR_MODULES_DISABLED" ""; do
+
+        if test -f "./${prefix}expected"; then
```
nmathewson: I am a little concerned about having relay_expected  mean "expected when NOT built a asrelay."  I would have expected something more like "norelay_expected" or expected_norelay.

src/feature/relay/routermode.h
teor2345: Fixed in 25a509478e0a57b698fe2785ce2de3549c2702f4
Fixed a missing comment for have_module_dirauth() in e6056fa6492821b67bf749a1011bff48382d05d4

src/feature/relay/router.h
teor2345: Replaced with tor_abort_() in b67c6dad9de3a9b89a367475000c675d7ef7e9b0

Is there a compiler warning that catches issues like this?
clang and gcc both seem very accepting.

src/feature/relay/router.c
teor2345: I opened tickets to:
* fix router.c - https://trac.torproject.org/projects/tor/ticket/32162
* fix similar issues with the dirauth module - https://trac.torproject.org/projects/tor/ticket/32163

src/test/test_parseconf.sh
teor2345: Changed to expected_no_dirauth in 3d38436d94e32cdf7e8157c3447bc70ee8328357 and d3f798096f725b1ebfe495af6e9c6c4c9a0460ce

Also added documentation for the module-based results files.

teor2345:
Rebased in #1429 and added fixups.
Squashed fixups in #1430.

PR #1423, opened 2019-10-17
Ticket30382 043 01

dgoulet-tor:


src/core/or/connection_edge.c
```diff
@@ -1553,6 +1553,95 @@ consider_plaintext_ports(entry_connection_t *conn, uint16_t port)
   return 0;
 }
 
+/** If address is of the form "y.onion" with a well-formed handle y:
```
asn-d6: Hm.I think this func doc can be improved.
 
- Before going into all the various cases let's document what this function is doing in general. Something simple like "Parse an address and set 'type_out' and return blabla"
- An undocumented behavior of this function is that it always sets 'type_out' regardless of success/fail. This is a behavior that other code seems to rely on and we should document it.

src/core/or/connection_edge.c
```diff
@@ -2020,16 +2109,15 @@ connection_ap_handshake_rewrite_and_attach(entry_connection_t *conn,
...
-
   /* Now see whether the hostname is bogus.  This could happen because of an
    * onion hostname whose format we don't recognize. */
-  if (addresstype == BAD_HOSTNAME) {
+  hostname_type_t addresstype;
+  if (!parse_extended_hostname(socks->address, &addresstype)) {
     control_event_client_status(LOG_WARN, "SOCKS_BAD_HOSTNAME HOSTNAME=%s",
                                 escaped(socks->address));
+    if (addresstype == ONION_V3_HOSTNAME) {
+      conn->socks_request->socks_extended_error_code = SOCKS5_HS_IS_INVALID;
```
asn-d6: Hmm, seems to me this is the feature from #30022 but we actually return that the descriptor is invalid. We should make a new error code for when the address is invalid, and this way we also satisfy #30022.

src/feature/hs/hs_cache.c
```diff
@@ -395,19 +395,29 @@ lookup_v3_desc_as_client(const uint8_t *key)
...
+      ret != HS_DESC_DECODE_BAD_CLIENT_AUTH) {
+    /* In the case of a missing or bad client authorization, we'll keep the
+     * descriptor in the cache because those credentials can arrive later. */
     goto end;
   }
-  tor_assert(desc);
+  /* Make sure we do have a descriptor if decoding was successful. */
+  if (ret == HS_DESC_DECODE_OK) {
+    tor_assert(desc);
+  }
```
asn-d6: Let's also check that desc is NULL if `ret != HS_DESC_DECODE_OK`.

src/feature/hs/hs_cache.c
```diff
@@ -761,27 +785,41 @@ hs_cache_lookup_as_client(const ed25519_public_key_t *key)
...
+ *                                  was not usable but the descriptor was
+ *                                  still stored.
+ *
+ *  Any other codes means indicate where the error occured and the descriptor
+ *  was not stored. */
+hs_desc_decode_status_t
 hs_cache_store_as_client(const char *desc_str,
                          const ed25519_public_key_t *identity_pk)
 {
+  hs_desc_decode_status_t ret;
```
asn-d6: Let's initialize this to some value for future proofing.

src/feature/hs/hs_client.c
```diff
@@ -1261,12 +1471,12 @@ hs_client_note_connection_attempt_succeeded(const edge_connection_t *conn)
...
  *
  * Return 0 on success else a negative value and desc is set to NULL. */
-int
+hs_desc_decode_status_t
 hs_client_decode_descriptor(const char *desc_str,
                             const ed25519_public_key_t *service_identity_pk,
                             hs_descriptor_t **desc)
 {
-  int ret;
+  hs_desc_decode_status_t ret;
```
asn-d6: Please initialize to something for future proofing.

src/feature/dirclient/dirclient.c
```diff
@@ -2728,62 +2728,7 @@ handle_response_fetch_hsdesc_v3(dir_connection_t *conn,
...
-    /* Fire control port FAILED event. */
-    hs_control_desc_event_failed(conn->hs_ident, conn->identity_digest,
-                                 "UNEXPECTED");
-    hs_control_desc_event_content(conn->hs_ident, conn->identity_digest,
-                                  NULL);
-    break;
-  }
-
+  hs_client_dir_fetch_done(conn, reason, body, status_code);
   return 0;
```
asn-d6: We can turn this function to void.

src/feature/hs/hs_cache.c
```diff
@@ -761,27 +785,41 @@ hs_cache_lookup_as_client(const ed25519_public_key_t *key)
   tor_assert(key);
 
   cached_desc = lookup_v3_desc_as_client(key->pubkey);
-  if (cached_desc) {
-    tor_assert(cached_desc->desc);
+  if (cached_desc && cached_desc->desc) {
```
asn-d6: Perhaps let's explicilty document that this function will never return a non-decrypted descriptor.

src/feature/hs/hs_client.c
```diff
@@ -1261,12 +1471,12 @@ hs_client_note_connection_attempt_succeeded(const edge_connection_t *conn)
  * newly allocated descriptor object.
  *
  * Return 0 on success else a negative value and desc is set to NULL. */
```
asn-d6: Let's update the function doc here, since while it's technically true, we now return a very detailed enum. The return value of this function is important since it floats up to the surface.

src/feature/hs/hs_client.c
```diff
@@ -1681,62 +1892,39 @@ hs_config_client_authorization(const or_options_t *options,
   return ret;
 }
 
-/** This is called when a descriptor has arrived following a fetch request and
- * has been stored in the client cache. Every entry connection that matches
- * the service identity key in the ident will get attached to the hidden
- * service circuit. */
+/** Called when a descriptor directory fetch is done. The ident is the
```
asn-d6: Let's improve  func doc here by saying what this function does and not only when it's called (or what it's arguments are).

For example, its calls are tasked with setting the SOCKS extended error code but this is not documented neither here or on its calls function documentation.

src/core/or/connection_edge.c
dgoulet-tor: Yes! I need to extend prop304 for that and then add that code. I recommend we do that part of #30022 but for sure before we release.

src/feature/hs/hs_cache.c
dgoulet-tor: I'm less incline here and the reason is because we loose the compiler warning if we fail to set it.

A default value makes me worry that we'll miss issues in the future actually.

src/feature/hs/hs_client.c
dgoulet-tor: See my reasoning above.

src/feature/dirclient/dirclient.c
dgoulet-tor: So hmmm that is a bit complicated because every `handle_response_*` return a value, even if they can't fail. Returning void means we need to handle that in the parent function that is a giant switch case expecting a returned value :S

src/feature/hs/hs_client.c
dgoulet-tor: Fixup commit: 3aae107e8e

src/feature/hs/hs_client.c
dgoulet-tor: Fixup commit 4194e6af10

src/feature/hs/hs_cache.c
dgoulet-tor: Fixup commit: e47973e15d

src/feature/hs/hs_cache.c
dgoulet-tor: Fixup commit: a4917825db

src/core/or/connection_edge.c
dgoulet-tor: Very good catch.

Fixup commit: 6a730b0118

src/feature/hs/hs_cache.c
asn-d6: I'm actually a bit anxious about the fatal asserts here, given that to see the whole logic we need to also check the correctness of `hs_client_decode_descriptor()`. Can we perhaps turn this into a BUG-and-exit thing?

src/core/or/connection_edge.c
asn-d6: ACK. How can we make sure that we don't release in the meanwhile? Because if we release, we will have to deal with weird "if tor_version == XXX then {} else {}" things.

dgoulet-tor: 043 is not in feature freeze nor close to a stable?

src/feature/hs/hs_cache.c
dgoulet-tor: Fixup commit: 8b694bd4b1

PR #1436, opened 2019-10-21
Ticket32172: port test suite to Android to run in emulator

eighthave:
* there is no /tmp on Android, there is /data/local/tmp for root and the shell user
* no use in testing the user ID stuff, Android apps cannot ever change users, and it was failing

https://trac.torproject.org/projects/tor/ticket/32172

Once the GitLab-CI setup from  !1408 is merged, then I'll submit the final work for running these tests in the Android emulator in GitLab-CI.  That should be portable to Travis-CI as well.

@n8fr8

### DIFF
--- a/src/test/testing_common.c
+++ b/src/test/testing_common.c
@@ -89,6 +89,22 @@ setup_directory(void)
                  (int)getpid(), rnd32);
     r = mkdir(temp_dir);
   }
+#elif defined(__ANDROID__)
+  /* tor might not like the default perms, so create a subdir */
+  tor_snprintf(temp_dir, sizeof(temp_dir),
+               "/data/local/tmp/tor_%d_%d_%s",
+               (int) getuid(), (int) getpid(), rnd32);
+  r = mkdir(temp_dir, 6770);
+  if (r) {
+    fprintf(stderr, "Can't create directory %s:", temp_dir);
+    perror("");
+    exit(1);
+  }
+  tor_snprintf(temp_dir, sizeof(temp_dir),
+               "/data/local/tmp/tor_%d_%d_%s/test_%d_%d_%s",
+               (int) getuid(), (int) getpid(), rnd32,
+               (int) getuid(), (int) getpid(), rnd32);
+  r = mkdir(temp_dir, 6770);
 #else /* !defined(_WIN32) */
   tor_snprintf(temp_dir, sizeof(temp_dir), "/tmp/tor_test_%d_%s",
                (int) getpid(), rnd32);


src/test/test_util.c
```diff
@@ -6286,6 +6286,10 @@ test_util_map_anon_nofork(void *arg)
 #define UTIL_TEST_NO_WIN(n, f) { #n, NULL, TT_SKIP, NULL, NULL }
 #define UTIL_TEST_WIN_ONLY(n, f) UTIL_TEST(n, (f))
 #define UTIL_LEGACY_NO_WIN(n) UTIL_TEST_NO_WIN(n, 0)
+#elif defined(__ANDROID__)
+#define UTIL_TEST_NO_WIN(n, f) { #n, NULL, TT_SKIP, NULL, NULL }
```
nmathewson: WIN should only mean "Windows", not "Android".  If we're going to change the meaning of these tests, we should also change their name.

src/test/testing_common.c
```diff
@@ -89,6 +89,11 @@ setup_directory(void)
                  (int)getpid(), rnd32);
     r = mkdir(temp_dir);
   }
+#elif defined(__ANDROID__)
+  r = mkdir("/data/local/tmp/tor", 0700);
```
nmathewson: We should check this return value.

nmathewson: (If different users all run this test, they seem to get subdirectories of the same directory, owned by an earlier user. That seems dangerous.)

src/test/test_util.c
eighthave: Ok, what do you want to call it?  I didn't rename it because I don't really understand what its filtering out.

src/test/testing_common.c
eighthave: There are no "different users" in Android in terms of UIDs for file permissions.  Each app gets assigned its own UID.  Apps cannot write to _/data/local/tmp_, only root or `adb shell` connections from a laptop.  So if the same device is connected to a different laptop or user account on that laptop, `adb shell` will still log into the Android emulator as the "shell" user.

Seems pedantic to check that return value, since if it fails, the next `mkdir()` would also fail.  But I'll do it if you want.

src/test/test_util.c
eighthave: I removed this part from this pull request.  I'll leave that to someone who knows that stuff.  For now, the pwdb test fails on Android.

eighthave:
Since Android has no /tmp/ or `mkdtemp()`, I changed this to fake it, and check the results of `mkdir()` in the process.

That took too long, ./configure with cross bulids is always painful, and tor is unfortunately no exception.  Then the tests have to run in an emulator.  I spent at least 3 hours updating this.

src/test/testing_common.c
```diff
@@ -89,6 +89,22 @@ setup_directory(void)
                  (int)getpid(), rnd32);
     r = mkdir(temp_dir);
   }
+#elif defined(__ANDROID__)
+  /* tor might not like the default perms, so create a subdir */
+  tor_snprintf(temp_dir, sizeof(temp_dir),
+               "/data/local/tmp/tor_%d_%d_%s",
+               (int) getuid(), (int) getpid(), rnd32);
+  r = mkdir(temp_dir, 6770);
```
nmathewson: Can we make this 700?

src/test/testing_common.c
```diff
@@ -89,6 +89,22 @@ setup_directory(void)
...
+  if (r) {
+    fprintf(stderr, "Can't create directory %s:", temp_dir);
+    perror("");
+    exit(1);
+  }
+  tor_snprintf(temp_dir, sizeof(temp_dir),
```
nmathewson: I'm not sure why this secondary directory is needed.  You mention permissions in the commit message, but this is created with the same permissions as the parent directory.

src/test/testing_common.c
```diff
@@ -89,6 +89,22 @@ setup_directory(void)
...
+               "/data/local/tmp/tor_%d_%d_%s/test_%d_%d_%s",
+               (int) getuid(), (int) getpid(), rnd32,
+               (int) getuid(), (int) getpid(), rnd32);
+  r = mkdir(temp_dir, 6770);
```
nmathewson: Have a look at the code that cleans up temp_dir when the tests exit, in "remove_directory()".  If we replace the value of temp_dir in this function, then that function will only remove the inner directory , not the outer one.

src/test/testing_common.c
eighthave: sure.

src/test/testing_common.c
eighthave: * _/data/local/tmp_ is not _/tmp_, it works differently, including different perms
* _/data/local/tmp_ does not get automatically wiped, only manually
* having the first dir with test subdirs keeps things a bit neater after multiple runs

eighthave:
These fixes are quite simple, but there seems to be many requirements that are unclear to me.  So I'll leave this code as an example for someone who does know the requirements to use how they see fit under any license they want.

Our upstreaming budget is done, so we're maintaining the required Android changes in our fork:
https://github.com/guardianproject/tor

PR #1450, opened 2019-10-22
Ticket31549 035

teor2345:


teor2345:
We decided not to backport this past 0.4.1.

PR #1456, opened 2019-10-23
hs-v3: Remove the circuit_established intro flag

dgoulet-tor:
Only use the HS circuit map to know if an introduction circuit is established
or not. No need for a flag to keep state of something we already have in the
circuit map. Furthermore, the circuit map gets cleaned up properly so it will
always have the "latest truth".

This commit also removes a unit test that was testing specifically that flag
but now we rely solely on the HS circuit map which is also tested few lines
below the removed test.

Fixes #32094

### DIFF
--- a/src/test/test_hs_service.c
+++ b/src/test/test_hs_service.c
@@ -1013,7 +1013,6 @@ test_intro_established(void *arg)
   /* Send an empty payload. INTRO_ESTABLISHED cells are basically zeroes. */
   ret = hs_service_receive_intro_established(circ, payload, sizeof(payload));
   tt_int_op(ret, OP_EQ, 0);
-  tt_u64_op(ip->circuit_established, OP_EQ, 1);
   tt_int_op(TO_CIRCUIT(circ)->purpose, OP_EQ, CIRCUIT_PURPOSE_S_INTRO);
 
  done:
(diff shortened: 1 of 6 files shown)


src/test/test_hs_service.c
```diff
@@ -1296,18 +1295,11 @@ test_service_event(void *arg)
      * descriptor map so we can retry it. */
     ip = helper_create_service_ip();
     service_intro_point_add(service->desc_current->intro_points.map, ip);
-    ip->circuit_established = 1;  /* We'll test that, it MUST be 0 after. */
-    run_housekeeping_event(now);
-    tt_int_op(digest256map_size(service->desc_current->intro_points.map),
-              OP_EQ, 1);
-    /* No removal if we have an established circuit after retries. */
```
asn-d6: Hmm, these two lines try to test the #31652 mechanism. Did the patch remove the "No removal if we have an established circuit after retries" part of the test?

dgoulet-tor: I don't think so.  From the commit message:

> This commit also removes a unit test that was testing specifically that flag
> but now we rely solely on the HS circuit map which is also tested few lines
> below the removed test.

I meant by that is this is removed because just after that removed test, we test the same thing but with a registered circuit. In other words, we bumped the retry value with a registered circuit and we still have it after housekeeping.

PR #1457, opened 2019-10-23
Add a script to rename C identifiers, and optionally add a commit mes…

nmathewson:
…sage

This script should "pay for itself" if it prevents one bug caused by
careless search-replace-commit loops.


scripts/maint/rename_c_identifier.py
```diff
@@ -0,0 +1,214 @@
...
+       >>> R.get_count()
+       3
+    """
+    def __init__(self, replacements):
+        """Make a new Rewriter. Takes a sequence of pairs of
+           (from_id, to_id), where from_id is an identifier to replace,
+           and to_id is its replacement.
+        """
+        self._patterns = [ ]
+        for id1,id2 in replacements:
```
pastly: nit: `id1, id2` unless you are really okay with missing a space

scripts/maint/rename_c_identifier.py
```diff
@@ -0,0 +1,214 @@
...
+            pat = re.compile(r"\b{}\b".format(re.escape(id1)))
+            self._patterns.append((pat, id2))
+
+        self._count = 0
+
+    def apply(self, line):
+        """Return `line` as transformed by this rewriter."""
+        for pat, ident in self._patterns:
+            line,count = pat.subn(ident, line)
```
pastly: nit: whitespace

scripts/maint/rename_c_identifier.py
```diff
@@ -0,0 +1,214 @@
...
+        print("Uncommitted changes found. Not running.", file=sys.stderr)
+        return 1
+
+    pairs = []
+    print("renaming {} to {}".format(args.from_id, args.to_id))
+    pairs.append((args.from_id, args.to_id))
+    for idx in range(0,len(args.more),2):
+        id1 = args.more[idx]
+        id2 = args.more[idx+1]
+        print("renaming {} to {}".format(id1, id2))
```
pastly: Not sys.stderr like other prints

scripts/maint/rename_c_identifier.py
```diff
@@ -0,0 +1,214 @@
...
+    """Try to commit the current git state, generating the commit message as
+       appropriate.  If `no_verify` is True, pass the --no-verify argument to
+       git commit.
+    """
+    args = []
+    if no_verify:
+        args.append("--no-verify")
+
+    fname = None
+    try:
```
pastly: I feel like this `try` is unneeded and that putting everything that uses `fname` and `f` inside the `with` block is enough. Is the extra `try` block and the `delete=false` flag addressing these docs? If so, I think that explains the extra wrapping.

> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later)

https://docs.python.org/3.5/library/tempfile.html#tempfile.NamedTemporaryFile

scripts/maint/rename_c_identifier.py
```diff
@@ -0,0 +1,214 @@
...
+       >>> is_c_file("a/b/.module.c")
+       False
+       >>> is_c_file("a/b/module.cpp")
+       False
+    """
+    fn = os.path.split(fn)[1]
+    if fn.startswith("."):
+        return False
+    ext = os.path.splitext(fn)[1]
+    return ext in {".c",".h",".i","inc"}
```
teor2345:
```suggestion
    return ext in {".c",".h",".i",".inc"}
```

scripts/maint/rename_c_identifier.py
```diff
@@ -0,0 +1,214 @@
...
+
+def list_c_files(topdir=TOPDIR):
+    """
+       Use git to list all the C files under version control.
+
+       >>> lst = list(list_c_files())
+       >>> "src/core/mainloop/mainloop.c" in lst
+       True
+       >>> "src/core/mainloop/twiddledeedoo.c" in lst
+       False
```
teor2345: Can you add an example for a generated C file?
We generate the git version include file.

scripts/maint/rename_c_identifier.py
```diff
@@ -0,0 +1,214 @@
...
+        encoding="utf-8")
+    for line in proc.stdout.readlines():
+        line = line.strip()
+        if is_c_file(line):
+            yield line
+
+class Rewriter:
+    """
+       A rewriter applies a series of word-by-word replacements, in
+       sequence.  Replacements only happen at word boundaries.
```
teor2345: What is a word?
Can you be clearer about which symbols and numbers are parts of words?
Can you show examples?

teor2345: "A word is defined as a sequence of alphanumeric or underscore characters"
"the precise set of characters deemed to be alphanumeric depends on the values of the UNICODE and LOCALE flags"
https://docs.python.org/2/library/re.html#regular-expression-syntax

teor2345: Strictly, this doesn't match C identifier syntax, which is `[a-zA-Z_][a-zA-Z_0-9]*`. But that's probably ok.

teor2345: c9a484361d2b452098cc8bdbfa71fbce22b9ce90

scripts/maint/rename_c_identifier.py
teor2345: 2be5f50bf9c8f1c7e8dfeeac1f5ea605e7fed3e8

scripts/maint/rename_c_identifier.py
teor2345: 2be5f50bf9c8f1c7e8dfeeac1f5ea605e7fed3e8

scripts/maint/rename_c_identifier.py
teor2345: 2be5f50bf9c8f1c7e8dfeeac1f5ea605e7fed3e8

scripts/maint/rename_c_identifier.py
teor2345: c69f17ae28fdc626e3019468a62c9dce4915b680

scripts/maint/rename_c_identifier.py
teor2345: 7ad0f4b8f9c922d7685fba87c038a63a250a153c

scripts/maint/rename_c_identifier.py
teor2345: 7272a4c1d74da7a3f11b404eb4b3f531c23394e9

scripts/maint/rename_c_identifier.py
teor2345: f7cde07da5a0b554406ba0c08c6ebbe7957515db

scripts/maint/rename_c_identifier.py
teor2345: 21c9c85da45fa2b65567e39cd8dc586a1643f93a

PR #1465, opened 2019-10-24
configure: Add --disable-module-relay to summary

dgoulet-tor:
Closes #32266

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2624,6 +2624,9 @@ PPRINT_SUBTITLE([Modules])
 test "x$enable_module_dirauth" != "xno" && value=1 || value=0
 PPRINT_PROP_BOOL([Directory Authority (--disable-module-dirauth)], $value)
 
+test "x$enable_module_relay" != "xno" && value=1 || value=0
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -2624,6 +2624,9 @@ PPRINT_SUBTITLE([Modules])
 test "x$enable_module_dirauth" != "xno" && value=1 || value=0
 PPRINT_PROP_BOOL([Directory Authority (--disable-module-dirauth)], $value)
 
+test "x$enable_module_relay" != "xno" && value=1 || value=0
+PPRINT_PROP_BOOL([Relay (--disable-module-relay)], $value)
```
teor2345: This logic is the wrong way around: if the relay module is enabled, we should print "--enable-module-relay yes", but we print "--disable-module-relay yes".

A similar issue affects lots of different options in the configure summary.

dgoulet-tor: Yeah, the string is not suppose to reflect what was enabled. Maybe we should do that instead but that is much more configure.ac trickery.

teor2345: Not really, it's just s/disable/enable/g

PR #1468, opened 2019-10-24
Ticket32277 tor manpage: reword command-line options and description; improve formatting

tlyu:
Contains wording changes and formatting improvements to DESCRIPTION and COMMAND-LINE OPTIONS sections. Based on a patch by Swati Thacker.

### DIFF
--- a/doc/tor.1.txt
+++ b/doc/tor.1.txt
@@ -18,22 +18,25 @@ SYNOPSIS
 
 DESCRIPTION
 -----------
-Tor is a connection-oriented anonymizing communication
-service. Users choose a source-routed path through a set of nodes, and
-negotiate a "virtual circuit" through the network, in which each node
-knows its predecessor and successor, but no others. Traffic flowing down
-the circuit is unwrapped by a symmetric key at each node, which reveals
-the downstream node. +
 
-Basically, Tor provides a distributed network of servers or relays ("onion routers").
-Users bounce their TCP streams -- web traffic, ftp, ssh, etc. -- around the
-network, and recipients, observers, and even the relays themselves have
-difficulty tracking the source of the stream.
+Tor is a connection-oriented anonymizing communication service. Users
+choose a source-routed path through a set of nodes, and negotiate a
+"virtual circuit" through the network. Each node in a virtual circuit
+knows its predecessor and successor nodes, but no other nodes. Traffic
+flowing down the circuit is unwrapped by a symmetric key at each node,
+which reveals the downstream node. +
 
-By default, **tor** will act as a client only.  To help the network
-by providing bandwidth as a relay, change the **ORPort** configuration
-option -- see below.  Please also consult the documentation on the Tor
-Project's website.
+Basically, Tor provides a distributed network of servers or relays
+("onion routers").  Users bounce their TCP streams, including web
+traffic, ftp, ssh, etc., around the network, so that recipients,
+observers, and even the relays themselves have difficulty tracking the
+source of the stream.
+
+[NOTE]
+By default, **tor** acts as a client only.  To help the network by
+providing bandwidth as a relay, change the **ORPort** configuration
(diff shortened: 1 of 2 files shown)


doc/tor.1.txt
```diff
@@ -37,6 +37,25 @@ Project's website.
...
+Configuration options can be specified on the command-line in the
+format **`--`**_OptionName_ _OptionValue_, or in a configuration file.
+For instance, you can tell Tor to start listening for SOCKS
+connections on port 9999 by passing either **`--SocksPort 9999`** or
+**`SocksPort 9999`** on the command line, or by specifying
+**`SocksPort 9999`** in the configuration file.  Quote option values
+that contain spaces. For instance, if you want Tor to log all
+debugging messages to **`debug.log`**, you must specify **`--Log
+"debug file debug.log"`**.
+
```
nmathewson: It might be good to have a sentence here saying "The following options are only recognized on the command line"

doc/tor.1.txt
```diff
@@ -18,22 +18,25 @@ SYNOPSIS
...
+Tor is a connection-oriented anonymizing communication service. Users
+choose a source-routed path through a set of nodes, and negotiate a
+"virtual circuit" through the network. Each node in a virtual circuit
+knows its predecessor and successor nodes, but no other nodes. Traffic
+flowing down the circuit is unwrapped by a symmetric key at each node,
+which reveals the downstream node. +
+
+Basically, Tor provides a distributed network of servers or relays
+("onion routers").  Users bounce their TCP streams, including web
+traffic, ftp, ssh, etc., around the network, while recipients,
```
nmathewson: Maybe replace "while" with "so that"?

doc/tor.1.txt
```diff
@@ -111,50 +111,55 @@ debugging messages to **`debug.log`**, you must specify **`--Log
...
+    certificate, if you already have a master key.  Optionally, you
+    can encrypt the master identity key with a passphrase.  When Tor
+    asks you for a passphrase and you don't want to encrypt the master
+    key, just don't enter any passphrase when asked. +
+ +
+    Use the **--newpass** option with --keygen only when you need to
+    add, change, or remove a passphrase on an existing ed25519 master
+    identity key. You will be prompted for the old passphase (if any),
+    and the new passphrase (if any).
++
```
nmathewson: I notice this isn't indented any more. Is that intentional? My asciidoc skills are rusty.

tlyu: Yes. The `+` on a line by itself is a list continuation, unlike the `+` at the end of a line, which was a line break. The list continuation makes the following paragraph part of the preceding list item. If I put the `+` on a line by itself as a list continuation, the indented paragraph gets rendered as a literal paragraph (which is more visible in the HTML output compared to the manpage output) instead of as a normal paragraph.

On the other hand, it seems like by explicitly putting paragraph styles like `[NOTE]` and `[normal]` in front of the continuation paragraphs, the indentation no longer results in the paragraph rendering as a literal paragraph. I can make that change if you think it helps to keep the paragraphs indented in the Asciidoc source.

doc/tor.1.txt
tlyu: Which paragraph is best for that sentence, do you think? Also maybe this wording would be a little more clear? "The options described in this section are only recognized on the `tor` command line, and not in the configuration file."

doc/tor.1.txt
tlyu: It still seems a little awkward as a single sentence. I think your suggestion of "so that" is an improvement, though.

doc/tor.1.txt
nmathewson: I think that reading is fine too; I think it should immediately precede the list of command-line options.

doc/tor.1.txt
nmathewson: I think it's fine; I trust your asciidoc.

PR #1475, opened 2019-10-26
Ticket32178 Removes trailing spaces in control event log

amadeuszpawlik:
Removes all trailing spaces in control_event_logmsg

### DIFF
--- a/src/feature/control/control_events.c
+++ b/src/feature/control/control_events.c
@@ -1318,6 +1318,16 @@ control_event_logmsg(int severity, log_domain_mask_t domain, const char *msg)
       for (cp = b; *cp; ++cp)
         if (*cp == '\r' || *cp == '\n')
           *cp = ' ';
+
+      /* Remove trailing spaces */
+      for (--cp; *cp == ' ' && cp >= b; --cp)
+        *cp = '\0';
+
+      if ( cp == b ){
+        ++disable_log_messages;
+        tor_assert_nonfatal(*b);
+        --disable_log_messages;
+      }
     }
     switch (severity) {
       case LOG_DEBUG: s = "DEBUG"; break;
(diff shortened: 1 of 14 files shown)


src/feature/control/control_events.c
```diff
@@ -1318,6 +1318,10 @@ control_event_logmsg(int severity, log_domain_mask_t domain, const char *msg)
       for (cp = b; *cp; ++cp)
         if (*cp == '\r' || *cp == '\n')
           *cp = ' ';
+
+      /* Remove trailing spaces */
+      for (--cp; *cp == ' '; --cp)
```
teor2345: Please don't write past the beginning of a string that only has spaces in it:
```suggestion
      for (--cp; *cp == ' ' && cp >= b; --cp)
```

amadeuszpawlik: Agreed;
In such case, should an empty string be sent (send_control_event(...)) at all?

teor2345: That's a good question.

If we're sending a log message that is just spaces, that's a bug. But we can't log from inside a logging function, unless we disable controller logs:
`
    ++disable_log_messages;
    tor_assert_nonfatal(*b);
    --disable_log_messages;
`

src/feature/control/control_events.c
```diff
@@ -1318,6 +1318,16 @@ control_event_logmsg(int severity, log_domain_mask_t domain, const char *msg)
       for (cp = b; *cp; ++cp)
         if (*cp == '\r' || *cp == '\n')
           *cp = ' ';
+
+      /* Remove trailing spaces */
+      for (--cp; *cp == ' ' && cp >= b; --cp)
+        *cp = '\0';
+
+      if ( cp == b ){
```
teor2345: The `cp == b` check isn't necessary, `tor_assert_nonfatal(*b);` makes sure that b is not an empty string.

PR #1478, opened 2019-10-28
Confdecl

nmathewson:


src/lib/conf/confdecl.h
```diff
@@ -0,0 +1,166 @@
...
+ *     BEGIN_CONF_STRUCT(module_options_t)
+ *     CONF_VAR(ModuleIsActive, BOOLEAN, 0, "1")
+ *     END_CONF_STRUCT(module_options_t)
+ *
+ * Once you've done that, you can use that header to define a configuration
+ * structure by saying:
+ *
+ *     typedef struct module_options_t module_options_t;
+ *     #define CONF_CONTEXT STRUCT
+ *     #include "MOD_options.inc"
```
teor2345: Inconsistent names:
```suggestion
 *     #include "MOD_opt.inc"
```

src/lib/conf/confdecl.h
```diff
@@ -0,0 +1,166 @@
...
+ *     #undef CONF_CONTEXT
+ *
+ * And you can define your field definition table by saying:
+ *
+ *     #define CONF_CONTEXT TABLE
+ *     #include "MODULE_options.inc"
```
teor2345: Inconsistent names:
```suggestion
 *     #include "MOD_opt.inc"
```

src/lib/conf/confdecl.h
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * @defgroup TABLE_MACROS Internal macros: table definitions.
+ * Implementation helpers: the regular confdecl macros expand to these
+ * when CONF_CONTEXT is defined to TABLE.  Don't use them directly.
+ * @{*/
+#define BEGIN_CONF_STRUCT__TABLE(structname)                            \
+  /* We use this typedef so we can refer to the config type */          \
+  /* without having its name as a macro argument to CONF_VAR. */        \
+  typedef struct structname config_var_reference__obj;  \
```
teor2345: Does this definition mean that you can't have two BEGIN_CONF_STRUCT__TABLE in the same file?
Would we ever want to?

src/lib/crypt_ops/crypto_conf.c
```diff
@@ -0,0 +1,27 @@
...
+#include "lib/crypt_ops/crypto_conf.h"
+#include "lib/crypt_ops/crypto_options_st.h"
+
+#include "lib/conf/confdecl.h"
+
+#define CONF_CONTEXT TABLE
+#include "lib/crypt_ops/crypto_options.inc"
+#undef CONF_CONTEXT
+
+const config_format_t crypto_fmt = {
```
teor2345: Why don't we have this happen as part of the macro?
Do we need to document this somewhere?

nmathewson: There are several other fields in the config_format_t that need to be set by hand.

src/lib/conf/confdecl.h
nmathewson: I don't think we would want to, but if we did, we could say `#define config_var_reference__obj dummy_object_1` before the first, and so on.

src/lib/conf/confdecl.h
teor2345: This name is still inconsistent after the fixup: MOD vs MODULE

PR #1481, opened 2019-10-29
force pkg-config to only use --prefix when cross-compiling

eighthave:
The current pkg-config setup has no sense of whether it is cross-compiling, so it will detect things on the build system that are not present or are wrong for the host system.  This forces the cross-compiling build to only look for pkg-config .pc files in --prefix.

A version of this has been the setup for many years with the Android builds.

Fixes #32191

Signed-off-by: Hans-Christoph Steiner <hans@eds.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -39,6 +39,16 @@ else
     pkg_config_user_action="check the PKG_CONFIG_PATH environment variable"
 fi
 
+if test "x$PKG_CONFIG_PATH" = "x" && test "x$prefix" != "xNONE" && test "$host" != "$build"; then
+   export PKG_CONFIG_PATH=$prefix/lib/pkgconfig
+   AC_MSG_NOTICE([set PKG_CONFIG_PATH=$PKG_CONFIG_PATH to support cross-compiling])
+   if test -f "$PKG_CONFIG_PATH/libevent.pc"; then
+       echo "checking for $PKG_CONFIG_PATH/libevent.pc... yes"
+   else
+       AC_MSG_ERROR([$PKG_CONFIG_PATH/libevent.pc not found!])
+   fi
+fi
+
 AC_ARG_ENABLE(openbsd-malloc,
    AS_HELP_STRING(--enable-openbsd-malloc, [use malloc code from OpenBSD.  Linux only. Deprecated: see --with-malloc]))
 AC_ARG_ENABLE(static-openssl,
--- a/changes/ticket32191
+++ b/changes/ticket32191
@@ -1,0 +1,3 @@
+  o Minor features (build system):
+    - force pkg-config to only use --prefix when cross-compiling.
+      Closes ticket 32191.


configure.ac
```diff
@@ -39,6 +39,16 @@ else
     pkg_config_user_action="check the PKG_CONFIG_PATH environment variable"
 fi
 
+if test "x$PKG_CONFIG_PATH" = "x" && test "x$prefix" != "xNONE" && test "$host" != "$build"; then
```
teor2345: Is prefix really set to a literal "NONE"?

configure.ac
```diff
@@ -39,6 +39,16 @@ else
...
+   export PKG_CONFIG_PATH=$prefix/lib/pkgconfig
+   AC_MSG_NOTICE([set PKG_CONFIG_PATH=$PKG_CONFIG_PATH to support cross-compiling])
+   if test -f "$PKG_CONFIG_PATH/libevent.pc"; then
```
teor2345: This test is incorrect: we don't require any particular *.pc file, the user can always set the relevant LIBS and CFLAGS env vars instead.

eighthave:
yes, that's a copy-paste from further below. it works.

configure.ac
eighthave: yes, that's a copy-paste from further below. it works.

configure.ac
eighthave: libevent is a hard requirement for building, and the idea here is that this is testing that the user ran `make install` into _prefix_ for libevent.  In that case, _libevent.pc_ will be present.  Supporting partial installs i.e. not `make install` into _prefix_, will waste a lot of time and only achieve flaky results.  Also, If there are no .pc files, then there is no point in setting `PKG_CONFIG_PATH`.  And remember, if pkg-config is not properly setup, then --enable-lzma fails without an error on Android builds, and probably other cases too.

eighthave: Or in other words, the current tor ./configure setup is a buggy almost unusable mess when it comes to cross-compiling for Android and iOS.  There are brittle wrapper build setups that all the related devs are scared to touch at all, so that hinders usage of new functionality like the new modularity.  There needs to be one recommended way to build that works reliably, e.g. `make install` and `--prefix` on each lib and tor.  This is a carefully isolated change that will be easy to avoid for someone who wants to do all sorts of build tricks.

teor2345: Tor does not use pkg-config to find libevent:
https://github.com/torproject/tor/blob/master/configure.ac#L751

For libraries where tor does use pkg-config, the pkg-config autotools macros allow the user to manually override the (MODULE)_CLFAGS and (MODULE)_LIBS, as documented in configure's failure messages:
https://github.com/torproject/tor/blob/master/configure.ac#L885

Here is the reference documentation for PKG_CHECK_MODULES():
https://autotools.io/pkgconfig/pkg_check_modules.html

PR #1485, opened 2019-10-30
circ: Add an handle to circuit_t

dgoulet-tor:
Closes #32343

### DIFF
--- a/src/core/or/circuit_st.h
+++ b/src/core/or/circuit_st.h
@@ -14,6 +14,8 @@
 
 #include "core/or/or.h"
 
+#include "lib/container/handles.h"
+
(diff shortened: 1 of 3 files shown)


src/core/or/circuit_st.h
```diff
@@ -234,4 +239,9 @@ struct circuit_t {
   struct circpad_machine_runtime_t *padding_info[CIRCPAD_MAX_MACHINES];
 };
 
+/* Declare the handle helpers */
+HANDLE_DECL(circuit, circuit_t, )
```
nmathewson: I think that these belong in one of the circuit*.h files -- we only put structures in _st.h files.

PR #1490, opened 2019-10-31
Ticket32020 043 01

dgoulet-tor:


src/core/or/circuituse.c
```diff
@@ -1984,6 +1984,45 @@ circuit_purpose_is_hidden_service(uint8_t purpose)
    return 0;
 }
 
+/** Retrun true iff the given circuit is an HS client circuit. */
```
asn-d6: Hmm, `circuit_purpose_is_hidden_service()` right above these new functions seems very related. How can we reuse these functions there? So that we don't have duplicate purpose-checking code?

src/feature/hs/hs_circuit.c
```diff
@@ -1205,6 +1206,15 @@ hs_circ_cleanup(circuit_t *circ)
 {
   tor_assert(circ);
 
+  /* v2 specific circuits. */
```
asn-d6: I'm fully onboard with this refactoring, however I'd like to make sure that it's indeed a no-behavior-change refactoring as the commit message claims.

In particular, this moves a whole bunch of code that used to reside in `circuit_about_to_free()` to `hs_circ_cleanup()`. What's the relationship of those two functions? Do they always get called at the same time?

Can this be documented in the commit message to clarify that no behavior change is done indeed?

src/feature/rend/rendclient.c
```diff
@@ -1253,47 +1253,61 @@ rend_parse_service_authorization(const or_options_t *options,
...
-    int timed_out = (reason == END_CIRC_REASON_TIMEOUT);
+  reason = circ->marked_for_close_reason;
+  orig_reason = circ->marked_for_close_orig_reason;
+  ocirc = CONST_TO_ORIGIN_CIRCUIT(circ);
+
+  has_timed_out = (reason == END_CIRC_REASON_TIMEOUT);
+  ip_is_redundant = (orig_reason == END_CIRC_REASON_IP_NOW_REDUNDANT);
+
+  switch (circ->purpose) {
+  case CIRCUIT_PURPOSE_C_INTRODUCE_ACK_WAIT:
```
asn-d6: Fully on board with this refactoring as well.

However, one behavior-change I can see is that the old code used to check for `ocirc->rend_data` here whereas the new code takes it for granted in both cases. Let's make sure that this does not cause any crashes.

src/feature/hs/hs_client.c
```diff
@@ -1239,6 +1239,55 @@ find_client_auth(const ed25519_public_key_t *service_identity_pk)
...
+  case CIRCUIT_PURPOSE_C_INTRODUCING:
+    if (has_timed_out || !orig_circ->build_state) {
+      break;
+    }
+    failure = INTRO_POINT_FAILURE_UNREACHABLE;
+    log_info(LD_REND, "Failed v3 intro circ for service %s to intro point %s "
+                      "(while building circuit). Marking as unreachable.",
+       safe_str_client(ed25519_fmt(&orig_circ->hs_ident->identity_pk)),
+       safe_str_client(build_state_get_exit_nickname(orig_circ->build_state)));
+    hs_cache_client_intro_state_note(&orig_circ->hs_ident->identity_pk,
```
asn-d6: Seems like that in both cases we call `hs_cache_client_intro_state_note()` in exactly the same fashion. Perhaps we can just do a single call to `hs_cache_client_intro_state_note()` in the bottom of the function and just use the switch clause to do the nice logging?

src/feature/rend/rendclient.c
```diff
@@ -1253,7 +1253,7 @@ rend_parse_service_authorization(const or_options_t *options,
 /** The given circuit is being freed. Take appropriate action if it is of
  * interest to the client subsystem. */
 void
-rend_client_circuit_cleanup(const circuit_t *circ)
+rend_client_circuit_cleanup(const circuit_t *circ, hs_circ_cleanup_type_t type)
```
asn-d6: Hm. I have mixed feelings about this cleanup type thing. On one hand it's a reasonable design, on the other hand I feel like it's a backwards step from the original nice refactoring of `hs_circ_cleanup()`. So like we managed to centralize all the cleanup code into a single function (yay!), but at the same time we introduced this cleanup type which changes the behavior of the function based on when it's called. So is this really one function, or is it actually three different functions (one for each cleanup type)?

It's hard to tell to be honest, because the type gets carried down to the deepest children of the function, and IMO that's a bit confusing.

I'd appreciate some changes to this effect:
a) Improve documentation to make it clear how each different cleanup type is used in the long-term. Right now it's not really clear to me. I see many functions only working for the `cleanup_on_free` type but I don't see the other types being used (?)
b) Improve function naming to make it clear what kind of thing each function does. For example, `hs_client_circuit_cleanup()` only works during freeing, but its name is still "cleanup". 
c) Come up with invariants to make sure that each type has been done. So if `cleanup_on_free` always comes after `cleanup_on_close` is it possible to ensure that this has indeed happened (to make sure we are not skipping any cleanup steps)?

We don't need to do perfectly all the above, but I feel like this is the messiest (in terms of tech debt) part of this branch so it would be great if we made it nicer.

src/core/or/circuituse.c
dgoulet-tor: So this is a "hard" one in terms of API but we could do better I agree.

Basically, we have HS circuit as a "catch all" circuit. Then we have *client* and *service* and then we have *v2* and *v3*.

So the `circuit_purpose_is_hidden_service()` is the catch all basically. I could add a refactor commit to use the new helper function within that big one?

src/feature/hs/hs_circuit.c
dgoulet-tor: Actually, it does change behavior at this commit which another commit fixes.

I was mistaken at first when I did that one that `hs_circ_cleanup()` was only called for the circuit free() but it is also called for the close().

One approach would be to change the commit structure to add the distinction between `close` and `free` that e3f2df295fd98f1afdd285a4bdd185145d7c73b0 introduces. And then do this commit... We are talking some refactoring :).

src/feature/rend/rendclient.c
dgoulet-tor: The idea behind this is that if you reach the HSv2 subsystem, you are _certain_ that it is a v2 circuit. In other words, HSv2 entry point should be reduced in favor of the `hs_circ_cleanup()` entry point. And thus I removed the check.

I could bring it back with `BUG()` ?

src/feature/rend/rendclient.c
dgoulet-tor: Yes indeed. This is the change I wanted more to get your feedback. I was not entirely pleased and the other option I had in mind was to have 2 different functions that is "on_close" and "on_free".

Originally, I wanted to make it through the pubsub subsystem but that wasn't really possible since, at least for the close() part, we CAN NOT have it async as in we need to remove the circuit from the map immediately.

I'm thinking that we should go with (b) as in 2 functions. The `circuit_free()` code already checks with a BUG() if the circuit was closed before so that state logic should be taken care there.

src/core/or/circuituse.c
asn-d6: Yep that would work. There might need to be some argument interface rearrangemnet but I think it's possible.

src/feature/rend/rendclient.c
asn-d6: Sounds good.

src/feature/rend/rendclient.c
asn-d6: Agreed let's do it with two functions if you also like this idea.

src/core/or/circuituse.c
dgoulet-tor: FIxed in the new PR. Basically, the 3 helpers function now take a purpose and not a circuit so then `circuit_purpose_is_hidden_service()` can use them. It is nicer now as I think it is unified better.

src/feature/hs/hs_circuit.c
dgoulet-tor: I did this work in the new PR. We now have a on_free, on_close and on_purpose functions. Will allow us to be much more future proof.

src/feature/hs/hs_client.c
dgoulet-tor: Well hmmm, this is a bit odd because the note happens for specific conditions so if I had to refactor to a single call, I would need a `bool flag` that tells me we hit the reason to note the issue.

Still want that?

PR #1493, opened 2019-11-04
Doxygen libs

nmathewson:


src/feature/hs/feature_hs.dox
```diff
@@ -1,4 +1,5 @@
 /**
 @dir /feature/hs
-@brief feature/hs
+@brief feature/hs: v3 (current) onion service protocol
```
dgoulet-tor: I would be super greatful if you would put here a ref to the `rend-spec-v3.txt` :D like it was done above:

[dir-spec.txt](https://gitweb.torproject.org/torspec.git/tree/dir-spec.txt),

src/mainpage.dox
```diff
@@ -3,9 +3,120 @@
...
+events.  As Tor runs over time, other events will open, and new events
+will be scheduled.
+
+The codebase is divided into a few top-level subdirectories, each of
+which contains several sub-modules.
+
+   - `ext` -- Code maintained elsewhere that we include in the Tor
+     source distribution.
+
+   - \refdir{lib} -- Lower-level utility code, not necessarily
```
dgoulet-tor: Not sure `refdir` outputs properly:

I get:

`./src/lib "lib" – Lower-level utility code, not necessarily tor-specific.`

src/mainpage.dox
```diff
@@ -3,9 +3,120 @@
...
+started.  This means, however, that some of the older pieces of Tor will have
+some "code smell" in them that could stand a brisk refactoring.  So when we
+describe a piece of code, we'll sometimes give a note on how it got that way,
+and whether we still think that's a good idea.
+
+This document is not an overview of the Tor protocol.  For that, see the
+design paper and the specifications at https://spec.torproject.org/ .
+
+For more information about Tor's coding standards and some helpful
+development tools, see doc/HACKING in the Tor repository.
```
dgoulet-tor: Probably would be nice to have a ref to `doc/HACKING` so user can click on.

src/feature/rend/feature_rend.dox
```diff
@@ -1,4 +1,5 @@
 /**
 @dir /feature/rend
-@brief feature/rend
+@brief feature/rend: version 2 (old) hidden services
```
dgoulet-tor: Same goes for this with `rend-spec.txt`.

src/feature/dirauth/feature_dirauth.dox
```diff
@@ -1,4 +1,6 @@
 /**
-@dir feature/dirauth
-@brief feature/dirauth
+@dir /feature/dirauth
+@brief feature/dirauth: Directory authority implementation.
+
+This module handles running Tor as a directory authority.
```
dgoulet-tor: Maybe ref to `dir-spec.txt`.

src/mainpage.dox
nmathewson: Hm. It works for me -- let's see how it works when we build the website.

PR #1502, opened 2019-11-05
Re-think our doxygen warnings wrt missing docs and fatalness.

nmathewson:
Up till now, we have warned about all missing documentation, which
meant that we could never make doxygen warnings fatal.  This has led
to our doxygen output getting full of errors several times in the
past.

This commit changes our approach to doxygen warnings: missing
documentation warnings are only on when the user asks for them with
--enable-missing-doc-warnings.  When that option is not present,
doxygen respects the --enable-fatal-warnings flag.

Closes ticket 32385.

### DIFF
--- a/changes/ticket32385
+++ b/changes/ticket32385
@@ -2,4 +2,5 @@
     - Our "make doxygen" target now respects --enable-fatal-warnings by
       default, and does not warn about items that are missing documentation.
       To warn about missing documentation, run configure with the
-      "--enable-missing-doc-warnings" flag. Closes ticket 32385.
+      "--enable-missing-doc-warnings" flag: doing suspends fatal warnings
+      for doxygen. Closes ticket 32385.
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -76,6 +76,24 @@ AC_ARG_ENABLE(nss,
 AC_ARG_ENABLE(pic,
    AS_HELP_STRING(--enable-pic, [Build Tor's binaries as position-independent code, suitable to link as a library.]))
 
+AC_ARG_ENABLE(missing-doc-warnings,
+   AS_HELP_STRING(--enable-missing-doc-warnings, [Tell doxygen to warn about missing documentation.]))
```
tlyu: This seems like it would make all Doxygen warnings (not just missing documentation) non-fatal, which seems like a reasonable constraint. Maybe we want to highlight this side effect in the help text and changes file?

changes/ticket32385
```diff
@@ -2,4 +2,5 @@
     - Our "make doxygen" target now respects --enable-fatal-warnings by
       default, and does not warn about items that are missing documentation.
       To warn about missing documentation, run configure with the
-      "--enable-missing-doc-warnings" flag. Closes ticket 32385.
+      "--enable-missing-doc-warnings" flag: doing suspends fatal warnings
```
tlyu: s/doing/doing so/ ?

PR #1505, opened 2019-11-05
ticket 30984 control refactor

tlyu:


src/feature/control/control_proto.c
```diff
@@ -275,3 +277,152 @@ control_write_data(control_connection_t *conn, const char *data)
...
+}
+
+/** Add a "250 OK" line to a set of control reply lines */
+void
+control_reply_add_done(smartlist_t *reply)
+{
+  control_reply_add_str(reply, 250, "OK");
+}
+
+void
```
teor2345: Doc?
And the remaining functions in this file.

src/feature/control/control_proto.c
```diff
@@ -275,3 +277,152 @@ control_write_data(control_connection_t *conn, const char *data)
...
+ * line list.
+ *
+ * @param reply smartlist of pointers to control_reply_line_t
+ * @param code numeric control reply code
+ * @param flags kvline encoding flags
+ * @param key key
+ * @param val value
+ */
+void
+control_reply_add_1kv(smartlist_t *reply, int code, int flags,
```
teor2345: This name is hard to read and type, because "1" and "l" and "I" look similar:
```suggestion
control_reply_add_single_kv(smartlist_t *reply, int code, int flags,
```

src/feature/control/control_cmd.c
```diff
@@ -299,12 +299,12 @@ handle_control_getconf(control_connection_t *conn,
...
         const char *name = option_get_canonical_name(q);
-        control_reply_add_1kv(answers, 250, KV_OMIT_VALS, name, "");
+        control_reply_add_one_kv(answers, 250, KV_OMIT_VALS, name, "");
       }
 
       while (answer) {
         config_line_t *next;
-        control_reply_add_1kv(answers, 250, KV_RAW, answer->key,
+        control_reply_add_one_kv(answers, 250, KV_RAW, answer->key,
                               answer->value);
```
teor2345: Nit: If you want to fix the spacing on this line before we merge, that would be good. Leaving it would also be ok.

PR #1518, opened 2019-11-07
Ticket32404

nmathewson:


src/lib/conf/conftypes.h
```diff
@@ -64,7 +64,8 @@ typedef enum config_type_t {
   CONFIG_TYPE_LINELIST_V,   /**< Catch-all "virtual" option to summarize
                              * context-sensitive config lines when fetching.
                              */
-  CONFIG_TYPE_OBSOLETE,     /**< Obsolete (ignored) option. */
+  CONFIG_TYPE_OBSOLETE,     /**< Ignored (obsolete) option. Uses no storage. */
```
teor2345: Please explain the difference between OBSOLETE and IGNORE: I think the difference is that OBSOLETE warns, and IGNORE does not.

Also, I don't think any options are IGNORE right now? They are all OBSOLETE?

PR #1529, opened 2019-11-11
Ticket32315

liberat2:


changes/bug32315
```diff
@@ -0,0 +1,4 @@
+  o Major bugfixes (networking):
```
teor2345: We'll probable end up categorising this as minor when we do the changelog, because no-one noticed it was broken for a few years.
```suggestion
  o Minor bugfixes (networking):
```

teor2345:
Backport branch is now #1674, with a small fix to the changes file.

PR #1539, opened 2019-11-15
Travis: use -std=gnu99 when compiling

teor2345:
Require C99 standards-conforming code in Travis CI, but allow GNU gcc
extensions. Also activates clang's -Wtypedef-redefinition warnings.

Closes ticket 32500.


teor2345:
Squashed and merged.

PR #1541, opened 2019-11-15
Ticket32500 042

teor2345:


teor2345:
No conflicts with 0.4.2 when merging forward the latest changes.
We only need to merge #1539 and #1540.

PR #1554, opened 2019-11-20
Ticket32020 043 02

dgoulet-tor:


src/feature/hs/hs_circuit.c
```diff
@@ -1197,20 +1215,17 @@ hs_circ_send_establish_rendezvous(origin_circuit_t *circ)
...
-  if (circ->purpose == CIRCUIT_PURPOSE_S_ESTABLISH_INTRO ||
-      circ->purpose == CIRCUIT_PURPOSE_S_INTRO) {
-    hs_service_intro_circ_has_closed(TO_ORIGIN_CIRCUIT(circ));
-  }
+  /* NOTE: Following code must remain light and fast. The circuit close
+   * function has to remain very fast since it is called as part of multiple
+   * main loop event. */
 
   /* Clear HS circuitmap token for this circ (if any). Very important to be
    * done after the HS subsystem has been notified of the close else the
```
asn-d6: I'm a bit confused on what this comment means now. What does "important to be done after the HS subsystem has been notified of the close" in this context? I thought that now this function is responsible for notifying the HS subsystem of the close, so what is important to be done before the circuitmap clear?

src/feature/hs/hs_circuit.c
```diff
@@ -1224,3 +1239,41 @@ hs_circ_cleanup(circuit_t *circ)
...
+
+  /* NOTE: Bulk of the work of cleaning up a circuit is done here. */
+
+  if (circuit_purpose_is_hs_client(circ->purpose)) {
+    cleanup_on_free_client_circ(circ);
+  }
+
+  /* We have no assurance that the given HS circuit has been closed before and
+   * thus removed from the HS map. This actually happens in unit tests. */
+  if (circ->hs_token) {
```
asn-d6: We could in theory call `hs_circuitmap_remove_circuit()` without checking for `hs_token` everytime, since the function does it for us. If we want....

src/feature/hs/hs_client.c
```diff
@@ -1522,6 +1522,55 @@ get_hs_client_auths_map(void)
...
+ * circuit is freed. */
+void
+hs_client_circuit_cleanup_on_free(const circuit_t *circ)
+{
+  bool has_timed_out;
+  rend_intro_point_failure_t failure = INTRO_POINT_FAILURE_GENERIC;
+  const origin_circuit_t *orig_circ = NULL;
+
+  tor_assert(circ);
+  tor_assert(CIRCUIT_IS_ORIGIN(circ));
```
asn-d6: Please add an assert for hs_ident here just to make sure we dont do any UAF or something in the future.

src/feature/rend/rendclient.c
```diff
@@ -1250,3 +1250,65 @@ rend_parse_service_authorization(const or_options_t *options,
...
+  tor_assert(CIRCUIT_IS_ORIGIN(circ));
+
+  reason = circ->marked_for_close_reason;
+  orig_reason = circ->marked_for_close_orig_reason;
+  ocirc = CONST_TO_ORIGIN_CIRCUIT(circ);
+
+  has_timed_out = (reason == END_CIRC_REASON_TIMEOUT);
+  ip_is_redundant = (orig_reason == END_CIRC_REASON_IP_NOW_REDUNDANT);
+
+  switch (circ->purpose) {
```
asn-d6: Please add an assert for rend_data here just to aovid any UAF here in the future.

src/feature/hs/hs_circuit.c
asn-d6: As a side note here, since we are being super nice and clean and making nice functions for our cleanup, I would really like a single high-level comment (on top of our cleanup functions) that details exactly our high-level cleanup strategy. We  are nicely detailing what each function should do, but it would be nice to have centralized what we want to have happened after all these functions are called.

changes/ticket32020
```diff
@@ -0,0 +1,4 @@
+  o Major bugfixes (onion service):
+    - Report back HS circuit failure back into the HS subsytem so we take
```
asn-d6: If this fix has reachability consequences (I think it does) we might want to mention it explicitly. We've been fixing lots of reachability issues in v3 lately, and it would be nice to be able to get a quick list at some point to show our progress (and also to know if a possible bug has been fixed on master).

src/feature/hs/hs_client.c
dgoulet-tor: Fixup: 5e77694957

src/feature/hs/hs_circuit.c
dgoulet-tor: Good idea. See fixup commit: f50e95bb58

src/feature/hs/hs_circuit.c
dgoulet-tor: New commit for this optimization: 031e35b128

src/feature/rend/rendclient.c
dgoulet-tor: Fixup: 2e17c328e8

changes/ticket32020
dgoulet-tor: Fixup: 90c21ed9fb

src/feature/hs/hs_circuit.c
dgoulet-tor: I have reverted in ed9d1a240fa728fe because most tests closing/freeing circuits don't call `hs_init()` and thus the HS circuit map asserts in `hs_circuitmap_remove_circuit()`.

PR #1562, opened 2019-11-25
Ticket 32604: Add two new config directives for onion services.

moonsikpark:
Add HiddenServiceExportRendPoint and HiddenServiceExportInstanceID
config directive to export rendezvous point information and instance
ID along with circuit ID. Closes ticket #32604.

### DIFF
--- a/changes/ticket32604
+++ b/changes/ticket32604
@@ -1,0 +1,7 @@
+  o Minor features (onion services):
+    - Version 3 onion services can now use HiddenServiceExportRendPoint
+      to get the IP and port of the rendezvous point they are connecting.
+    - Version 3 onion services can now use HiddenServiceExportInstanceID
+      which enables them to differenciate traffic when multiple tor
+      instances share the same backend.
+    - Closes ticket 32604. Patch by Moonsik Park.
--- a/doc/tor.1.txt
+++ b/doc/tor.1.txt
@@ -3044,6 +3044,31 @@ The next section describes the per service options that can only be set
    The HAProxy version 1 protocol is described in detail at
    https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt
 
+[[HiddenServiceExportRendPoint]] **HiddenServiceExportRendPoint** **0**|**1**::
+   When HiddenServiceExportCircuitID along with this feature is enabled, export IP
+   and port of the rendezvous point along with the circuit ID. +
+ +
+   When this feature is enabled, tor will export IP and port of the rendezvous
+   point in the following way. +
+ +
+   "PROXY TCP6 fc00::AAAA:BBBB:CCCC:0:ffff:ffff ::1 65535 42\r\n" +
+ +
+      ip = (0xAAAA << 16) + 0xBBBB; +
+      port = 0xCCCC; +
+ +
+   For example, if the IP is 0x7f000001 and the port is 0x1bb, you can translate
+   to 127.30.0.1:443. (Default: 0)
+
+[[HiddenServiceExportInstanceID]] **HiddenServiceExportInstanceID** __N__::
+   When HiddenServiceExportCircuitID is enabled, export an instance ID along
+   with the circuit ID. The maximum allowed value is 65535. +
+ +
+   When this feature is enabled, tor will export an instance ID in the following way. +
+ +
+   "PROXY TCP6 fc00::0:0:0:AAAA:ffff:ffff ::1 65535 42\r\n" +
+ +
+   Where 0xAAAA will be your provided instance ID in hex. (Default: 0)
+
 [[HiddenServiceMaxStreams]] **HiddenServiceMaxStreams** __N__::
    The maximum number of simultaneous streams (connections) per rendezvous
    circuit. The maximum value allowed is 65535. (Setting this to 0 will allow
--- a/src/app/config/or_options_st.h
+++ b/src/app/config/or_options_st.h
@@ -1103,6 +1103,12 @@ struct or_options_t {
    **/
   int DormantCanceledByStartup;
 
+  /** Boolean: true if we should export rend point when exporting CircuitID. */
+  int HiddenServiceExportRendPoint;
+
+  /** Instance ID to export when exporting CircuitID. */
+  int HiddenServiceExportInstanceID;
+
   /**
    * Configuration objects for individual modules.
    *
--- a/src/core/or/connection_edge.c
+++ b/src/core/or/connection_edge.c
@@ -863,18 +863,35 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
   if (protocol != HS_CIRCUIT_ID_PROTOCOL_HAPROXY)
     return;
 
+  origin_circuit_t *circ;
   char *buf = NULL;
   const char dst_ipv6[] = "::1";
   /* See RFC4193 regarding fc00::/7 */
-  const char src_ipv6_prefix[] = "fc00:dead:beef:4dad:";
+  const char src_ipv6_prefix[] = "fc00:";
   uint16_t dst_port = 0;
   uint16_t src_port = 1; /* default value */
   uint32_t gid = 0; /* default value */
+  uint32_t rp_addr = 0;
+  uint16_t rp_port = 0;
 
   /* Generate a GID and source port for this client */
   if (edge_conn->on_circuit != NULL) {
-    gid = TO_ORIGIN_CIRCUIT(edge_conn->on_circuit)->global_identifier;
+    circ = TO_ORIGIN_CIRCUIT(edge_conn->on_circuit);
+    gid = circ->global_identifier;
     src_port = gid & 0x0000ffff;
+    if (get_options()->HiddenServiceExportRendPoint) {
+      crypt_path_t *cpath = circ->cpath;
+      crypt_path_t *cpath_last = NULL;
+      if (cpath) {
+        for (; cpath_last != cpath; cpath = cpath_last) {
+          cpath_last = cpath->next;
+        }
+      }
+      if (cpath_last) {
+        rp_addr = tor_addr_to_ipv4h(&cpath_last->extend_info->addr);
+        rp_port = cpath_last->extend_info->port;
+      }
+    }
   }
 
   /* Grab the original dest port from the hs ident */
@@ -883,8 +900,10 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
   }
 
   /* Build the string */
-  tor_asprintf(&buf, "PROXY TCP6 %s:%x:%x %s %d %d\r\n",
+  tor_asprintf(&buf, "PROXY TCP6 %s:%x:%x:%x:%x:%x:%x %s %d %d\r\n",
                src_ipv6_prefix,
+               rp_addr >> 16, rp_addr & 0x0000ffff, rp_port,
+               get_options()->HiddenServiceExportInstanceID,
                gid >> 16, gid & 0x0000ffff,
                dst_ipv6, src_port, dst_port);
 
--- a/src/test/test_hs_service.c
+++ b/src/test/test_hs_service.c
@@ -2127,7 +2127,7 @@ test_export_client_circuit_id(void *arg)
   /* Check contents */
   cp1 = buf_get_contents(conn->outbuf, &sz);
   tt_str_op(cp1, OP_EQ,
-            "PROXY TCP6 fc00:dead:beef:4dad::0:29a ::1 666 42\r\n");
+            "PROXY TCP6 fc00::0:0:0:0:0:29a ::1 666 42\r\n");
 
   /* Change circ GID and see that the reported circuit ID also changes */
   or_circ->global_identifier = 22;
@@ -2144,7 +2144,7 @@ test_export_client_circuit_id(void *arg)
   export_hs_client_circuit_id(edge_conn, service->config.circuit_id_protocol);
   cp1 = buf_get_contents(conn->outbuf, &sz);
   tt_str_op(cp1, OP_EQ,
-            "PROXY TCP6 fc00:dead:beef:4dad::ffff:ffff ::1 65535 42\r\n");
+            "PROXY TCP6 fc00::0:0:0:0:ffff:ffff ::1 65535 42\r\n");
   tor_free(cp1);
 
   /* Check that GID with UINT16_MAX works. */
@@ -2153,7 +2153,7 @@ test_export_client_circuit_id(void *arg)
   export_hs_client_circuit_id(edge_conn, service->config.circuit_id_protocol);
   cp1 = buf_get_contents(conn->outbuf, &sz);
   tt_str_op(cp1, OP_EQ,
-            "PROXY TCP6 fc00:dead:beef:4dad::0:ffff ::1 65535 42\r\n");
+            "PROXY TCP6 fc00::0:0:0:0:0:ffff ::1 65535 42\r\n");
   tor_free(cp1);
 
   /* Check that GID with UINT16_MAX + 7 works. */
@@ -2161,7 +2161,7 @@ test_export_client_circuit_id(void *arg)
 
   export_hs_client_circuit_id(edge_conn, service->config.circuit_id_protocol);
   cp1 = buf_get_contents(conn->outbuf, &sz);
-  tt_str_op(cp1, OP_EQ, "PROXY TCP6 fc00:dead:beef:4dad::1:6 ::1 6 42\r\n");
+  tt_str_op(cp1, OP_EQ, "PROXY TCP6 fc00::0:0:0:0:1:6 ::1 6 42\r\n");
 
  done:
   UNMOCK(connection_write_to_buf_impl_);
--- a/src/app/config/config.c
+++ b/src/app/config/config.c
@@ -513,6 +513,8 @@ static const config_var_t option_vars_[] = {
   VAR("HiddenServiceMaxStreamsCloseCircuit",LINELIST_S, RendConfigLines, NULL),
   VAR("HiddenServiceNumIntroductionPoints", LINELIST_S, RendConfigLines, NULL),
   VAR("HiddenServiceExportCircuitID", LINELIST_S,  RendConfigLines, NULL),
+  V(HiddenServiceExportRendPoint, BOOL, "0"),
+  V(HiddenServiceExportInstanceID, POSINT, "0"),
   VAR("HiddenServiceEnableIntroDoSDefense", LINELIST_S, RendConfigLines, NULL),
   VAR("HiddenServiceEnableIntroDoSRatePerSec",
       LINELIST_S, RendConfigLines, NULL),
@@ -3938,6 +3940,10 @@ options_validate_cb(const void *old_options_, void *options_, char **msg)
 
   if (options_validate_scheduler(options, msg) < 0) {
     return -1;
+  }
+
+  if (options->HiddenServiceExportInstanceID > 65535) {
+    REJECT("HiddenServiceExportInstanceID must be in range of 0 to 65535");
   }
 
   return 0;


src/core/or/connection_edge.c
```diff
@@ -863,18 +863,35 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
...
+    circ = TO_ORIGIN_CIRCUIT(edge_conn->on_circuit);
+    gid = circ->global_identifier;
     src_port = gid & 0x0000ffff;
+    if (get_options()->HiddenServiceExportRendPoint) {
+      crypt_path_t *cpath = circ->cpath;
+      crypt_path_t *cpath_last = NULL;
+      if (cpath) {
+        for (; cpath_last != cpath; cpath = cpath_last) {
+          cpath_last = cpath->next;
+        }
```
dgoulet-tor: So I believe, on an `origin_circuit_t`, you can avoid going through the cpath and just use the build state:

`circ->build_state->chosen_exit`

Which is conveniently the `extend_info_t` you need below. Just be careful, the `build_state` could be NULL but unlikely so you could use your macro `BUG()` for that and bail if so.

src/core/or/connection_edge.c
```diff
@@ -863,18 +863,35 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
   if (protocol != HS_CIRCUIT_ID_PROTOCOL_HAPROXY)
     return;
 
+  origin_circuit_t *circ;
```
dgoulet-tor: I would recommend using a `const` here and then `CONST_TO_ORIGIN_CIRCUIT()` for extra safety.

src/app/config/or_options_st.h
```diff
@@ -1103,6 +1103,12 @@ struct or_options_t {
    **/
   int DormantCanceledByStartup;
 
+  /** Boolean: true if we should export rend point when exporting CircuitID. */
+  int HiddenServiceExportRendPoint;
+
+  /** Instance ID to export when exporting CircuitID. */
+  int HiddenServiceExportInstanceID;
```
dgoulet-tor: Documenting the Min, Max and Default here would be desirable.

doc/tor.1.txt
```diff
@@ -3044,6 +3044,31 @@ The next section describes the per service options that can only be set
...
+[[HiddenServiceExportInstanceID]] **HiddenServiceExportInstanceID** __N__::
+   When HiddenServiceExportCircuitID is enabled, export an instance ID along
+   with the circuit ID. The maximum allowed value is 65535. +
+ +
+   When this feature is enabled, tor will export an instance ID in the following way. +
+ +
+   "PROXY TCP6 fc00::0:0:0:AAAA:ffff:ffff ::1 65535 42\r\n" +
+ +
+   Where 0xAAAA will be your provided instance ID in hex. (Default: 0)
+
```
dgoulet-tor: I can't tell on Github but just make sure that the text paragraph are 79 chars aligned maximum :). If so, great!

moonsikpark: Fixed in `288823c`.

Before

```
>>> for line in diff.split('\n'):
...     print(f'{"!" if len(line) > 79 else ""}{len(line):02}', line)
... 
79 [[HiddenServiceExportRendPoint]] **HiddenServiceExportRendPoint** **0**|**1**::
!82    When HiddenServiceExportCircuitID along with this feature is enabled, export IP
64    and port of the rendezvous point along with the circuit ID. +
02  +
78    When this feature is enabled, tor will export IP and port of the rendezvous
32    point in the following way. +
02  +
61    "PROXY TCP6 fc00::AAAA:BBBB:CCCC:0:ffff:ffff ::1 65535 42
03 " +
02  +
37       ip = (0xAAAA << 16) + 0xBBBB; +
22       port = 0xCCCC; +
02  +
!80    For example, if the IP is 0x7f000001 and the port is 0x1bb, you can translate
34    to 127.30.0.1:443. (Default: 0)
00 
75 [[HiddenServiceExportInstanceID]] **HiddenServiceExportInstanceID** __N__::
76    When HiddenServiceExportCircuitID is enabled, export an instance ID along
61    with the circuit ID. The maximum allowed value is 65535. +
02  +
!87    When this feature is enabled, tor will export an instance ID in the following way. +
02  +
55    "PROXY TCP6 fc00::0:0:0:AAAA:ffff:ffff ::1 65535 42
03 " +
02  +
70    Where 0xAAAA will be your provided instance ID in hex. (Default: 0)
00 

```
After

```
>>> for line in diff.split('\n'):
...     print(f'{"!" if len(line) > 79 else ""}{len(line):02}', line)
... 
79 [[HiddenServiceExportRendPoint]] **HiddenServiceExportRendPoint** **0**|**1**::
72    When HiddenServiceExportCircuitID along with this feature is enabled,
74    export IP and port of the rendezvous point along with the circuit ID. +
02  +
78    When this feature is enabled, tor will export IP and port of the rendezvous
32    point in the following way. +
02  +
61    "PROXY TCP6 fc00::AAAA:BBBB:CCCC:0:ffff:ffff ::1 65535 42
03 " +
02  +
37       ip = (0xAAAA << 16) + 0xBBBB; +
22       port = 0xCCCC; +
02  +
70    For example, if the IP is 0x7f000001 and the port is 0x1bb, you can
44    translate to 127.30.0.1:443. (Default: 0)
00 
75 [[HiddenServiceExportInstanceID]] **HiddenServiceExportInstanceID** __N__::
76    When HiddenServiceExportCircuitID is enabled, export an instance ID along
61    with the circuit ID. The maximum allowed value is 65535. +
02  +
63    When this feature is enabled, tor will export an instance ID
26    in the following way. +
02  +
55    "PROXY TCP6 fc00::0:0:0:AAAA:ffff:ffff ::1 65535 42
03 " +
02  +
70    Where 0xAAAA will be your provided instance ID in hex. (Default: 0)
```

src/app/config/or_options_st.h
moonsikpark: Fixed in `ec8f17d`.

src/core/or/connection_edge.c
moonsikpark: Fixed in `b0e2baf`.

src/core/or/connection_edge.c
moonsikpark: Fixed in `0599c72`.

dgoulet-tor: Oh I think you can simply remove the fallback through the `cpath` and just bail out after the BUG() with a simple `return;`.

The point is that it is really not suppose to happen but if so, we'll notice and we'll be able to fix it. Chances are that the `cpath` is broken if the `chosen_exit` is not present. Not only that but you might end up using a `cpath` that is not at all the rendezvous node for instance.

moonsikpark: Fixed in `a932db5`.

changes/ticket32604
```diff
@@ -0,0 +1,7 @@
+  o Minor features (onion services):
+    - Version 3 onion services can now use HiddenServiceExportRendPoint
+      to get the IP and port of the rendezvous point they are connecting.
```
ahf: I think I'd have written "connecting to" here?

doc/tor.1.txt
```diff
@@ -3044,6 +3044,32 @@ The next section describes the per service options that can only be set
...
+   When this feature is enabled, tor will export IP and port of the rendezvous
+   point in the following way. +
+ +
+   "PROXY TCP6 fc00::AAAA:BBBB:CCCC:0:ffff:ffff ::1 65535 42\r\n" +
+ +
+      ip = (0xAAAA << 16) + 0xBBBB; +
+      port = 0xCCCC; +
+ +
+   For example, if the IP is 0x7f000001 and the port is 0x1bb, you can
+   translate to 127.30.0.1:443. (Default: 0)
```
ahf: Am I missing something here or will this not work with IPv6?

src/core/or/connection_edge.c
```diff
@@ -863,18 +863,31 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
   if (protocol != HS_CIRCUIT_ID_PROTOCOL_HAPROXY)
     return;
 
+  const origin_circuit_t *circ;
```
ahf: Please initialize this to `NULL`.

src/core/or/connection_edge.c
```diff
@@ -863,18 +863,31 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
...
   /* Generate a GID and source port for this client */
   if (edge_conn->on_circuit != NULL) {
-    gid = TO_ORIGIN_CIRCUIT(edge_conn->on_circuit)->global_identifier;
+    circ = CONST_TO_ORIGIN_CIRCUIT(edge_conn->on_circuit);
+    gid = circ->global_identifier;
     src_port = gid & 0x0000ffff;
+    if (get_options()->HiddenServiceExportRendPoint) {
+      extend_info_t *extend_info = circ->build_state->chosen_exit;
+      if (BUG(extend_info == NULL)) {
+        tor_free(buf);
```
ahf: I think you should replace this `tor_free(buf)` here with a `goto` statement to the end of the function where `tor_free(buf)` is called as well.

src/core/or/connection_edge.c
moonsikpark: Fixed in `99808ad`.

src/core/or/connection_edge.c
moonsikpark: Fixed in `1609fe9`.

changes/ticket32604
moonsikpark: Fixed in `f38218a`.

doc/tor.1.txt
```diff
@@ -3044,6 +3044,32 @@ The next section describes the per service options that can only be set
    The HAProxy version 1 protocol is described in detail at
    https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt
 
+[[HiddenServiceExportRendPoint]] **HiddenServiceExportRendPoint** **0**|**1**::
+   When HiddenServiceExportCircuitID along with this feature is enabled,
+   export IP and port of the rendezvous point along with the circuit ID. +
```
teor2345: Why require HiddenServiceExportCircuitID to be enabled?

doc/tor.1.txt
```diff
@@ -3044,6 +3044,32 @@ The next section describes the per service options that can only be set
...
+   "PROXY TCP6 fc00::AAAA:BBBB:CCCC:0:ffff:ffff ::1 65535 42\r\n" +
+ +
+      ip = (0xAAAA << 16) + 0xBBBB; +
+      port = 0xCCCC; +
+ +
+   For example, if the IP is 0x7f000001 and the port is 0x1bb, you can
+   translate to 127.30.0.1:443. (Default: 0)
+
+[[HiddenServiceExportInstanceID]] **HiddenServiceExportInstanceID** __N__::
+   When HiddenServiceExportCircuitID is enabled, export an instance ID along
```
teor2345: Why require HiddenServiceExportCircuitID to be enabled?

doc/tor.1.txt
moonsikpark: Because that directive specifies the protocol(haproxy proxy). We could create another directive like `HiddenServiceExportProtocol` and make `HiddenServiceExportCircuitID` a bool, but that way we will introduce breaking changes.

@teor2345 @dgoulet @ahf opinions?

doc/tor.1.txt
teor2345: Instead, we could enable this feature when HiddenServiceExportInstanceID is non-zero.

src/core/or/connection_edge.c
```diff
@@ -863,18 +863,30 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
...
   uint32_t gid = 0; /* default value */
+  uint32_t rp_addr = 0;
+  uint16_t rp_port = 0;
 
   /* Generate a GID and source port for this client */
   if (edge_conn->on_circuit != NULL) {
-    gid = TO_ORIGIN_CIRCUIT(edge_conn->on_circuit)->global_identifier;
+    circ = CONST_TO_ORIGIN_CIRCUIT(edge_conn->on_circuit);
+    gid = circ->global_identifier;
     src_port = gid & 0x0000ffff;
```
teor2345: Putting part of the circuit id in the port isn't documented, please document it as part of this patch.

doc/tor.1.txt
moonsikpark: Good point. As Tor will move on to accept IPv6-only gradually, as agreed I think it's better to add first 12 chars of relay fingerprint instead of the (ip, orport) tuple.

src/core/or/connection_edge.c
```diff
@@ -863,18 +863,30 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
...
+    if (get_options()->HiddenServiceExportRendPoint) {
+      extend_info_t *extend_info = circ->build_state->chosen_exit;
+      if (BUG(extend_info == NULL)) {
+        goto cleanup;
+      }
+      rp_addr = tor_addr_to_ipv4h(&extend_info->addr);
```
teor2345: This code will fail on IPv6 single onion services, please check that the address is IPv4 before converting.
(As an IPv6 fallback, you can try looking up the IPv4 address in the consensus, but it won't always be there, because the client and service can have different consensuses.)

src/core/or/connection_edge.c
```diff
@@ -883,14 +895,19 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
   }
 
   /* Build the string */
-  tor_asprintf(&buf, "PROXY TCP6 %s:%x:%x %s %d %d\r\n",
+  tor_asprintf(&buf, "PROXY TCP6 %s:%x:%x:%x:%x:%x:%x %s %d %d\r\n",
```
teor2345: These should probably be "%04x"?

src/core/or/connection_edge.c
```diff
@@ -883,14 +895,19 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
...
                src_ipv6_prefix,
+               rp_addr >> 16, rp_addr & 0x0000ffff, rp_port,
```
teor2345: These should all be cast to uint16_t.

src/test/test_hs_service.c
```diff
@@ -2153,15 +2153,15 @@ test_export_client_circuit_id(void *arg)
...
+            "PROXY TCP6 fc00::0:0:0:0:0:ffff ::1 65535 42\r\n");
   tor_free(cp1);
 
   /* Check that GID with UINT16_MAX + 7 works. */
   or_circ->global_identifier = UINT16_MAX + 7;
 
   export_hs_client_circuit_id(edge_conn, service->config.circuit_id_protocol);
   cp1 = buf_get_contents(conn->outbuf, &sz);
-  tt_str_op(cp1, OP_EQ, "PROXY TCP6 fc00:dead:beef:4dad::1:6 ::1 6 42\r\n");
+  tt_str_op(cp1, OP_EQ, "PROXY TCP6 fc00::0:0:0:0:1:6 ::1 6 42\r\n");
```
teor2345: Please add tests for the 2 new options, not just the old ones.

doc/tor.1.txt
moonsikpark: Fixed in `5399de4`.

src/core/or/connection_edge.c
moonsikpark: Fixed in `5399de4`.

src/core/or/connection_edge.c
moonsikpark: Code removed in `5399de4`.

src/core/or/connection_edge.c
moonsikpark: Code removed in `5399de4`.

src/core/or/connection_edge.c
moonsikpark: Manpage says `All other values in the header can safely be ignored.`

@ahf: Why are we putting part of the circuit id in place of source port? Should I document it, remove it or what?

src/test/test_hs_service.c
moonsikpark: Fixed in `4a80f4f`.

doc/tor.1.txt
moonsikpark: See https://github.com/torproject/tor/pull/1562#discussion_r353520163 .

doc/tor.1.txt
teor2345: That's ok, but you should be aware that the fingerprint is SHA1(encoded(relay-public-key)). So it's relatively easy to brute-force 48 bits of that key.

So unlike an IPv4 address, it won't be possible to uniquely identify most relays using the partial fingerprint:
* there can be two relays in the consensus with the same 48 bit prefix
* there can be a relay in the consensus, and an attacker can set up an unpublished relay with the same fingerprint prefix

If that fits your use case, that's ok. But we do need to document that the partial fingerprint does not allow you to reliably identify a relay.

src/core/or/connection_edge.c
teor2345: Removing it is a breaking change, if someone is relying on the implementation. (Rather than the documentation.)

So it's safer to document it. If we decide to remove it, we should replace it with part of the fingerprint (so it varies in a similar way), and document that breaking change in the changes file.

src/core/or/connection_edge.c
teor2345: All integers in this printf that are 4 hex digits wide should be cast to uint16_t. If we do the cast, then they can't possibly exceed 4 hex digits. If we don't, then any bugs produce bad format strings.

src/core/or/connection_edge.c
teor2345: If we change the source port, we'll also need to update the tests.

doc/tor.1.txt
teor2345: It's not a particularly clean design, but I think it's ok as long as it's documented. If I was designing this from scratch, we'd have one option for the protocol, and then just export all the fields. (Or have one option per field.)

src/core/or/connection_edge.c
```diff
@@ -863,18 +863,34 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
...
+                                           DIGEST_LEN);
+      tor_strlower(rp_identity);
+      tor_asprintf(&rp_identity_buf, "%.4s:%.4s:%.4s",
+                   rp_identity,
+                   rp_identity + 4,
+                   rp_identity + 8);
+    }
   }
 
   /* Grab the original dest port from the hs ident */
```
teor2345: Here's another undocumented feature: the destination port is the virtual port from the onion service protocol.

changes/ticket32604
```diff
@@ -0,0 +1,7 @@
+  o Minor features (onion services):
+    - Version 3 onion services can now use HiddenServiceExportRendPoint
+      to get first 12 chars of the rendezvous point's fingerprint.
+    - Version 3 onion services can now use HiddenServiceExportInstanceID
+      which enables them to differenciate traffic when multiple tor
+      instances share the same backend.
+    - Closes ticket 32604. Patch by Moonsik Park.
```
teor2345: This isn't quite how we format changes files. I'll suggest some changes on individual lines, but you'll need to re-wrap the lines.

changes/ticket32604
```diff
@@ -0,0 +1,7 @@
+  o Minor features (onion services):
+    - Version 3 onion services can now use HiddenServiceExportRendPoint
```
teor2345:
```suggestion
    - When HiddenServiceExportCircuitID is set, version 3 onion services
      can also set HiddenServiceExportRendPoint
```

changes/ticket32604
```diff
@@ -0,0 +1,7 @@
...
+      to get first 12 chars of the rendezvous point's fingerprint.
```
teor2345:
```suggestion
      to export the first 12 chars of the rendezvous point's fingerprint
      using the haproxy protocol.
```

changes/ticket32604
```diff
@@ -0,0 +1,7 @@
...
+    - Version 3 onion services can now use HiddenServiceExportInstanceID
+      which enables them to differenciate traffic when multiple tor
+      instances share the same backend.
```
teor2345: Please make similar changes to this change log entry.

changes/ticket32604
```diff
@@ -0,0 +1,7 @@
+  o Minor features (onion services):
+    - Version 3 onion services can now use HiddenServiceExportRendPoint
+      to get first 12 chars of the rendezvous point's fingerprint.
+    - Version 3 onion services can now use HiddenServiceExportInstanceID
+      which enables them to differenciate traffic when multiple tor
```
teor2345:
```suggestion
      to differentiate traffic when multiple tor
```

changes/ticket32604
teor2345: Each changelog entry can be sorted independently, so there needs to be a "Closes" line in each entry, not just the last entry. Closes lines do not start with "- ".
```suggestion
    - Closes ticket 32604. Patch by Moonsik Park.
```

src/core/or/connection_edge.c
moonsikpark: Fixed in `8078900`.

src/core/or/connection_edge.c
teor2345: It will need tests, too :-)

teor2345: It will need tests, too :-)

src/core/or/connection_edge.c
```diff
@@ -883,14 +899,20 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
   }
 
   /* Build the string */
-  tor_asprintf(&buf, "PROXY TCP6 %s:%x:%x %s %d %d\r\n",
+  tor_asprintf(&buf, "PROXY TCP6 %s:%s:%04x:%04x:%04x %s %d %d\r\n",
                src_ipv6_prefix,
-               gid >> 16, gid & 0x0000ffff,
+               rp_identity_buf ? rp_identity_buf : "0000:0000:0000",
+               get_options()->HiddenServiceExportInstanceID,
```
teor2345: This was actually the one I was most concerned about :-)
```suggestion
               (uint16_t) (get_options()->HiddenServiceExportInstanceID),
```

doc/tor.1.txt
moonsikpark: @teor2345 How can there can be two relays in the consensus with the same fingerprint? Using the same relay key on different machines?

changes/ticket32604
moonsikpark: Fixed in `2582e28`.

changes/ticket32604
moonsikpark: Fixed in `2582e28`.

changes/ticket32604
moonsikpark: Fixed in `2582e28`.

changes/ticket32604
moonsikpark: Fixed in `2582e28`.

changes/ticket32604
moonsikpark: Fixed in `2582e28`.

doc/tor.1.txt
teor2345: The directory authorities only put one relay in the consensus for each unique fingerprint. But this code only uses 48 out of the 160 fingerprint bits. That's enough to identify most relays, but a determined attacker could fake those 48 fingerprint bits. (Also, SHA1 is being phased out, because it is insecure even when you use all 160 bits.)

So we need to document that it's usually right, but not cryptographically secure.

teor2345: I think it's probably easiest if I fix this documentation?

changes/ticket32604
teor2345: I think it's probably easiest if I fix the changes file?

changes/ticket32604
teor2345: I think it's probably easiest if I fix the remaining spelling mistake in the changes file?

src/core/or/connection_edge.c
teor2345: @ahf: whatever we decide to do with the source port, we should do the same thing with the destination port. I think we should just document them, to avoid breaking existing setups.

doc/tor.1.txt
teor2345: Fixed in 8ef44ca.

src/core/or/connection_edge.c
teor2345: See the remaining issues list in:
https://trac.torproject.org/projects/tor/ticket/32604#comment:19

src/core/or/connection_edge.c
teor2345: Documented in 49127e7.

changes/ticket32604
teor2345: Fixed in 3a8dee4.

changes/ticket32604
teor2345: Fixed in 3a8dee4.

teor2345:
See also:
* 2f1aafa where I rewrite the man page option descriptions so they are consistent
* d68c71b where I fix the double-free
* 3093619 where I use the "free after every use" pattern to make memory issues less likely

src/test/test_hs_service.c
```diff
@@ -2285,9 +2285,6 @@ test_export_client_circuit_id(void *arg)
   cp1 = buf_get_contents(conn->outbuf, &sz);
   tt_str_op(cp1, OP_EQ,
             "PROXY TCP6 fc00::ffff:ffff:ffff:ffff:ffff:ffff ::1 65535 42\r\n");
-  tor_free(cp1);
-  extend_info_free(chosen_exit);
-  or_circ->build_state->chosen_exit = NULL;
```
teor2345: It's actually safe to call tor_free() and extend_info_free() here, because they set their arguments to NULL.

We sometimes write code with *_free() at the end of every test, so that the next person who adds a test doesn't have to remember to add the extra *_free() calls?

src/test/test_hs_service.c
```diff
@@ -2285,14 +2285,9 @@ test_export_client_circuit_id(void *arg)
...
 
  done:
   UNMOCK(connection_write_to_buf_impl_);
-  extend_info_free(chosen_exit);
-  or_circ->build_state->chosen_exit = NULL;
```
teor2345: It would also help to add a comment explaining that circuit_free() also frees chosen_exit.

moonsikpark: Fixed in `06a16fc`.

src/test/test_hs_service.c
moonsikpark: Fixed in `06a16fc`.

doc/tor.1.txt
```diff
@@ -3023,27 +3023,76 @@ The next section describes the per service options that can only be set
...
+   is 255. (Default: 0)
+ +
+   When **HiddenServiceExportCircuitID** is set to \'haproxy', tor will export
+   an instance ID in the following way. +
+ +
+   "PROXY TCP6 fdAA::0000:0000:0000:0000:0000:0000 ::1 0 0\r\n" +
+ +
+   HiddenServiceExportInstanceID = 0xAA +
+ +
+   The byte order of the instance id varies, based on the host's byte
```
teor2345: This comment isn't true any more: a 1-byte field doesn't have a byte order.

PR #1563, opened 2019-11-25
Bug32562

asn-d6:


src/feature/hs/hs_client.c
```diff
@@ -1445,6 +1445,79 @@ client_dir_fetch_unexpected(dir_connection_t *dir_conn, const char *reason,
...
+static char *
+get_client_auth_creds_filename(const char *onion_address,
+                               const char *dir)
+{
+  char *full_fname = NULL;
+  char *fname;
+
+  tor_asprintf(&fname, "%s.auth_private", onion_address);
+  full_fname = hs_path_from_filename(dir, fname);
+  tor_free(fname);
```
dgoulet-tor: Maybe you meant to use `full_name` ?

src/feature/hs/hs_client.c
```diff
@@ -1445,6 +1445,79 @@ client_dir_fetch_unexpected(dir_connection_t *dir_conn, const char *reason,
                                 NULL);
 }
 
+/** Get the full filename for storing the client auth credentials for the
+ *  service in <b>onion_address</b>. The base directory is <b>dir</b>. */
```
dgoulet-tor: Seems this can never fail that is never return NULL. I would add a comment about that.

src/feature/hs/hs_client.c
```diff
@@ -1612,6 +1612,90 @@ get_creds_from_client_auth_filename(const char *filename,
...
+    log_warn(LD_REND, "Failed to remove client auth file (Not a file) (%s).",
+             creds_file_path);
+    goto end;
+  }
+
+  if (tor_unlink(creds_file_path) != 0) {
+    log_warn(LD_REND, "Failed to remove client auth file (%s).",
+             creds_file_path);
+    goto end;
+  }
```
dgoulet-tor: Hmmm, don't we have a TOCTOU issue race? Seems possibly harmless but you never know. I would probably just let `tor_unlink()` fail here instead of looking if the file is legit before?

src/feature/hs/hs_client.c
asn-d6: I actually did mean `full_fname`, in the sense that it's the full path instead of just the actual filename. Let me know if you want me to change it to something else.

src/feature/hs/hs_client.c
asn-d6: Fixed in e7f8319022.

src/feature/hs/hs_client.c
asn-d6: Fixed in 2852f8d732.

FWIW I was following the logic from `expire_old_onion_keys()` but this new logic is also fine.

PR #1586, opened 2019-12-05
Reject 0.2.9 and 0.4.0 in dirserv_rejects_tor_version()

neelchauhan:


src/test/test_process_descs.c
```diff
@@ -37,10 +34,11 @@ test_process_descs_versions(void *arg)
     { "Tor 0.3.4.100", true },
     { "Tor 0.3.5.1-alpha", true },
     { "Tor 0.3.5.6-rc", true},
+    { "Tor 0.4.0.1-alpha", true },
+    { "Tor 0.4.0.5", true },
     // new enough to be supported
     { "Tor 0.3.5.7", false },
     { "Tor 0.3.5.8", false },
-    { "Tor 0.4.0.1-alpha", false },
     { "Tor 0.4.1.5", false },
```
teor2345: Please add test cases for the current releases 0.4.2 and 0.4.3.

src/feature/dirauth/process_descs.c
```diff
@@ -331,22 +331,20 @@ dirserv_rejects_tor_version(const char *platform,
...
 
-  /* Series between Tor 0.3.0 and 0.3.4 inclusive are unsupported, and some
-   * have bug #27841, which makes them broken as intro points. Reject them.
-   *
-   * Also reject unstable versions of 0.3.5, since (as of this writing)
-   * they are almost none of the network. */
-  if (tor_version_as_new_as(platform,"0.3.0.0-alpha-dev") &&
-      !tor_version_as_new_as(platform,"0.3.5.7")) {
+  /* Series between Tor 0.3.6 and 0.4.0 inclusive are unsupported. Reject
+   * them. */
```
teor2345: Let's document what actually happened to 0.3.6.
```suggestion
   * them. 0.3.6.0-alpha-dev only existed for a short time, before it was renamed to 0.4.0.0-alpha-dev. */
```

src/feature/dirauth/process_descs.c
```diff
@@ -331,22 +331,20 @@ dirserv_rejects_tor_version(const char *platform,
...
+  if (tor_version_as_new_as(platform,"0.3.6.0-alpha-dev") &&
+      !tor_version_as_new_as(platform,"0.4.1.1-alpha")) {
```
teor2345: We should also decide if we want to reject 0.4.1 alphas and rcs.

PR #1597, opened 2019-12-11
Ticket32542 043 01

dgoulet-tor:


doc/tor.1.txt
```diff
@@ -1456,6 +1454,11 @@ The following options are useful only for clients (that is, if
             error is returned: address checksum doesn't match, ed25519 public
             key is invalid or the encoding is invalid. (v3 only)
 
+          X'F7' Onion Service Introduction Timed Out
```
asn-d6: We decided this together but I dont quite remember why we chose to have F2 and F7 as distinct error codes. Is it possible to document this rationale somewhere?

src/feature/hs/hs_circuit.c
```diff
@@ -620,6 +620,20 @@ setup_introduce1_data(const hs_desc_intro_point_t *ip,
...
+static void
+cleanup_on_close_client_circ(circuit_t *circ)
+{
+  tor_assert(circ);
+
+  if (circuit_is_hs_v3(circ)) {
+    hs_client_circuit_cleanup_on_close(circ);
+  }
+  /* It is possible the circuit has an HS purpose but no identifier (rend_data
+   * or hs_ident). Thus possible that this passess through. */
```
asn-d6: nit: passess -> passes

src/feature/hs/hs_client.c
```diff
@@ -956,6 +956,37 @@ client_get_random_intro(const ed25519_public_key_t *service_pk)
...
 
+/** Return true iff all intro points for the given service have timed out. */
+static bool
+intro_points_all_timed_out(const ed25519_public_key_t *service_pk)
+{
+  bool ret = false;
+
+  tor_assert(service_pk);
+
+  const hs_descriptor_t *desc = hs_cache_lookup_as_client(service_pk);
```
asn-d6: Let's check that we actually got desc != NULL here before dereferencing it.

src/feature/hs/hs_client.c
```diff
@@ -956,6 +956,37 @@ client_get_random_intro(const ed25519_public_key_t *service_pk)
   return ei;
 }
 
+/** Return true iff all intro points for the given service have timed out. */
```
asn-d6: It's a bit weird commit structure here where the first commit is about intro, the second comimt is rend, and the third commit is again about intro and also editing the firstr commit. I had to rebase it and squash the intro stuff into a single commit to be able to review it effectively.

src/feature/hs/hs_client.c
```diff
@@ -956,6 +956,27 @@ client_get_random_intro(const ed25519_public_key_t *service_pk)
   return ei;
 }
 
+/** Called when a rendezvous circuit has timed out. Every streams attached to
+ * the circuit will get set with the SOCKS5_HS_REND_FAILED (0xF3) extended
+ * error code so if the connection to the rendezvous point ends up not
+ * working, this code could be sent back as a reason. */
+static void
+socks_report_rend_circuit_timed_out(const origin_circuit_t *rend_circ)
```
asn-d6: I'm not sure if 'report' is the right thing to call this. It's more of a 'mark' thing than a 'report', since by calling this function we dont report immediately.

src/feature/hs/hs_client.c
```diff
@@ -956,6 +956,27 @@ client_get_random_intro(const ed25519_public_key_t *service_pk)
...
+{
+  tor_assert(rend_circ);
+
+  /* For each entry connections attached to this rendezvous circuit, report
+   * the error. */
+  for (edge_connection_t *edge = rend_circ->p_streams; edge;
+       edge = edge->next_stream) {
```
asn-d6: Could this circuit have no streams? Do we care? Will this crash?

src/feature/hs/hs_client.c
```diff
@@ -1288,6 +1302,10 @@ client_desc_has_arrived(const smartlist_t *entry_conns)
     if (!hs_client_any_intro_points_usable(identity_pk, desc)) {
       log_info(LD_REND, "Hidden service descriptor is unusable. "
                         "Closing streams.");
+      /* Report the extended socks error code that we were unable to introduce
```
asn-d6: I'm wondering if this is the right place to call this function. In particular, we seem to only get here when we get a new desc, but isn't that too late? This means that we fetched two descriptors before throwing the error page. Shouldn't we throw the error when we finish up all intros of the first descriptor? Perhaps we should be calling this from another place where `hs_client_any_intro_points_usable()` is used?

doc/tor.1.txt
dgoulet-tor: https://trac.torproject.org/projects/tor/ticket/32542#comment:2

src/feature/hs/hs_client.c
dgoulet-tor: Right so the code is a bit messy to follow but the structure is a commit per error code.

src/feature/hs/hs_client.c
dgoulet-tor: Replaced to `_mark_`. See fixup commit.

src/feature/hs/hs_client.c
dgoulet-tor: I believe it is not possible. If the SOCKS connection is closed, before we remove the stream from the circuit, we would close the circuit and thus updating the error code and then closing the SOCKs connection.

If it ever happens, won't matter, the loop will just be NOP.

src/feature/hs/hs_client.c
dgoulet-tor: Not sure... Tor client behavior has always been to try all intro points, if they all failed, refetch new descriptor and if it failed again, that means the service is gone and close SOCKS connection.

Thing is that the client doesn't know if the descriptor it has is out of date so the second retry is important for reachability but also attempt to get a newer descriptor before the cache entry expires.

src/feature/hs/hs_client.c
dgoulet-tor: Fixed, see latest commit.

src/feature/hs/hs_circuit.c
dgoulet-tor: Fixed, see latest commit.

PR #1601, opened 2019-12-14
Flag a v3 intro point as bad if rendezvous fails during cleanup

neelchauhan:


src/feature/hs/hs_client.c
```diff
@@ -1778,6 +1778,16 @@ hs_client_circuit_cleanup_on_free(const circuit_t *circ)
...
     break;
+  case CIRCUIT_PURPOSE_C_REND_READY_INTRO_ACKED:
+    log_info(LD_REND, "Failed v3 intro circ for service %s to intro point %s "
+                      "(rendevous point failed).",
+        safe_str_client(ed25519_fmt(&orig_circ->hs_ident->identity_pk)),
+        safe_str_client(build_state_get_exit_nickname(
+                                                     orig_circ->build_state)));
+    hs_cache_client_intro_state_note(&orig_circ->hs_ident->identity_pk,
+                                     &orig_circ->hs_ident->intro_auth_pk,
+                                     failure);
```
dgoulet-tor: So I have a couple of concerns here.

1. This notes down a failure regardless of the failure type. I think we want this only when circuit times out since this would "assume" that the introduction never reached the rendezvous point and thus potential malicious IP.

2. We should really try at least once more in my opinion before flagging it bad and not immediately. This means that you would need to find the intro circuit associated with this rendezvous one (I think it is easy with the HS circuit map lookup) and put in a counter in that circuit hs identifier possibly. Above let say 2 tries, the IP should be flagged as bad.

src/feature/hs/hs_client.c
```diff
@@ -1778,6 +1778,16 @@ hs_client_circuit_cleanup_on_free(const circuit_t *circ)
                                      &orig_circ->hs_ident->intro_auth_pk,
                                      failure);
     break;
+  case CIRCUIT_PURPOSE_C_REND_READY_INTRO_ACKED:
+    log_info(LD_REND, "Failed v3 intro circ for service %s to intro point %s "
+                      "(rendevous point failed).",
+        safe_str_client(ed25519_fmt(&orig_circ->hs_ident->identity_pk)),
+        safe_str_client(build_state_get_exit_nickname(
+                                                     orig_circ->build_state)));
```
dgoulet-tor: Hmmm, this circuit is a rendezvous circuit, not the introduction circuit so you are logging and flagging the wrong circuit here.

PR #1607, opened 2019-12-16
Dirauth config

nmathewson:


src/feature/dirauth/.may_include
```diff
@@ -1 +1,2 @@
 *.h
+*.inc
```
teor2345: I don't think it needs to include any *.inc outside of feature/dirauth/ ?
```suggestion
feature/dirauth/*.inc
```

src/feature/dirauth/dirauth_stub.c
```diff
@@ -11,9 +11,23 @@
...
+#undef CONF_CONTEXT
+
+static const config_format_t dirauth_options_stub_fmt = {
+  .vars = dirauth_options_t_vars,
+};
 
 const struct subsys_fns_t sys_dirauth = {
   .name = "dirauth",
   .supported = false,
   .level = 70,
```
teor2345: If we duplicate the level in *_sys and *_stub, they will get out of sync.

Maybe supported modules shouldn't need a level?
Or maybe they should have an otherwise-illegal level number, using a named #define for readability?

src/feature/relay/relay_stub.c
```diff
@@ -0,0 +1,20 @@
...
+ **/
+
+#include "orconfig.h"
+#include "feature/relay/relay_sys.h"
+#include "lib/subsys/subsys.h"
+
+const struct subsys_fns_t sys_relay = {
+  .name = "relay",
+  .supported = false,
+  .level = 50,
```
teor2345: Similar to dirauth_stub, the level should not be duplicated.

teor2345:
I think this ticket was squashed and merged?

PR #1610, opened 2019-12-16
Pre formatter cleanups

nmathewson:


src/lib/cc/compat_compiler.h
```diff
@@ -72,15 +70,15 @@
...
           PRAGMA_DIAGNOSTIC_(push) \
-          PRAGMA_DIAGNOSTIC_(ignored PRAGMA_JOIN_STRINGIFY_(-W,warningopt))
+          PRAGMA_DIAGNOSTIC_(ignored warningopt)
 #    define ENABLE_GCC_WARNING(warningopt) \
           PRAGMA_DIAGNOSTIC_(pop)
 #else /* !(defined(__clang__) || GCC_VERSION >= 406) */
 /* older version of gcc: no push/pop support. */
 #    define DISABLE_GCC_WARNING(warningopt) \
-         PRAGMA_DIAGNOSTIC_(ignored PRAGMA_JOIN_STRINGIFY_(-W,warningopt))
+         PRAGMA_DIAGNOSTIC_(ignored PRAGMA_JOIN_STRINGIFY_ warningopt)
```
tlyu: Should probably completely delete `PRAGMA_JOIN_STRINGIFY_` as was done on L73. Same on L81.

Preprocessor substitution of empty token for undefined macros probably makes this compile, but we probably shouldn't leave it lying around.

PR #1618, opened 2019-12-19
Test headers

teor2345:


scripts/test/build_headers.sh
```diff
@@ -3,42 +3,129 @@
...
+# We deliberately skip win32: it only contains a Windows orconfig.h
+for hdr in $(cd "${SOURCE_DIR}"/src && find lib core feature app test tools -name '*.h'); do
+    hdr_path="${SOURCE_DIR}"/src/"${hdr}"
+    name=$(basename "$hdr" .h)
+    tc_name="${name}"_h
+    tc_path="${TEST_DIR}"/"$tc_name".c
 
-for hdr in $(cd ./src && find lib core feature app -name '*.h'); do
-  name=$(basename "$hdr" .h)
+    grep '^ *# *if\(def *\| .*defined(\)[A-Z_]*\(PRIVATE\|INTERNAL\|EXPOSE\)[A-Z_]*' "$hdr_path" | \
```
nmathewson: I thought that we were going to standardize on only using PRIVATE.  Do we have a ticket for that?

scripts/test/build-headers.sh
```diff
@@ -139,18 +139,41 @@ done
...
+fi
+
+# Check for errors
+if test "$fail_count" -gt "$exception_count"; then
+    echo "Too many header compilation failures!"
+    echo "Expected ${exception_count}, got ${fail_count}."
+    EXIT_STATUS=1
+fi
+
+# Ignore headers that we expected to fail, but actually didn't fail
```
nmathewson: Instead of ignoring these, maybe we should do a nonfatal notice that we can now safely remove the entry from the exceptions?

scripts/test/build_headers.sh
teor2345: I'm happy to standardise on [A-Z_]*_PRIVATE. I don't know if we have a ticket for that. (Maybe we should also remove the unused PRIVATE defined at the same time.)

When that ticket merges, we can simplify this pattern, and the exclude list.

scripts/test/build-headers.sh
teor2345: We can't say that, because while we may be able to safely remove the entry on this platform, it might be there because it fails on other platforms.

This message already lists the differences:
https://github.com/torproject/tor/pull/1618/files#diff-3cd653bd07713d0f33273db2fffbcac6R163

scripts/test/build_headers.sh
teor2345: I opened trac 32818 for this change.

PR #1625, opened 2019-12-20
router: Stop advertising incorrect auto IPv6 ORPorts

teor2345:
When IPv6 ORPorts are set to "auto", tor relays and bridges would
advertise an incorrect port in their descriptor.

This may be a low-severity memory safety issue, because the published
port number may be derived from uninitialised or out-of-bounds memory
reads.

Fixes bug 32588; bugfix on 0.2.3.9-alpha.


teor2345:
Obsoleted by #1654, leaving open for CI.

teor2345:
Obsoleted by #1654.

PR #1626, opened 2019-12-20
Ticket32822

teor2345:


src/feature/relay/router.c
```diff
@@ -1433,6 +1433,47 @@ router_get_advertised_or_port_by_af(const or_options_t *options,
...
 }
 
+/** As router_get_advertised_or_port(), but returns the IPv6 address and
+ *  port. Returns a null address and zero port, if no ORPort is found. */
+void
+router_get_advertised_ipv6_or_ap(const or_options_t *options,
+                                 tor_addr_port_t *ipv6_ap_out)
+{
+  if (BUG(!ipv6_ap_out))
+    return;
```
dgoulet-tor: I would use `tor_assert()` here because this is a caller convention setup rather than external input we can't control.

Furthermore, tor will misbehave after this return by accessing stack junk :S.

src/feature/relay/router.c
```diff
@@ -1433,6 +1433,47 @@ router_get_advertised_or_port_by_af(const or_options_t *options,
...
+  if (! tor_addr_is_internal(addr, 0) || ! default_auth) {
+    tor_addr_copy(&ipv6_ap_out->addr, addr);
+    ipv6_ap_out->port = port;
+  } else {
+    char addrbuf[TOR_ADDR_BUF_LEN];
+    log_warn(LD_CONFIG,
+             "Unable to use configured IPv6 address \"%s\" in a "
+             "descriptor. Skipping it. "
+             "Try specifying a globally reachable address explicitly.",
+             tor_addr_to_str(addrbuf, addr, sizeof(addrbuf), 1));
```
dgoulet-tor: To avoid that extra `addrbuf[]`, you could use `fmt_addrport()`. No strong opinion there.

src/feature/relay/router.c
teor2345: Fixed in 532d9b6, squashed into #1654.

I'll force-push the squashed and merged branch to this PR.

src/feature/relay/router.c
teor2345: Fixed in f63e110, squashed into #1654.

I'll force-push the squashed and merged branch to this PR.

PR #1628, opened 2019-12-20
Ticket22029 attempt

nmathewson:


src/feature/dirauth/process_descs.c
```diff
@@ -174,27 +214,52 @@ dirserv_load_fingerprint_file(void)
...
+    } else {
+      ed25519_not_ok = -1;
+    }
+
+    /* If this is not a valid key, log and skip */
+    if (ed25519_not_ok && rsa_not_ok) {
+      log_warn(LD_CONFIG, "Invalid fingerprint (nickname '%s', "
+               "fingerprint %s). Skipping.", nickname, fingerprint);
+      continue;
+    }
```
dgoulet-tor: Ok this `if()` got me thinking for a bit on why both need to be bad to skip it and warn. It is correct but I would document why.

Having `rsa_not_ok = -1` doesn't necessarily means the RSA fingerprint is "wrong" but also that there is simply none and the ED key is probably used.

Just a comment explaining that since it is not that straight forward.

src/feature/dirauth/process_descs.c
```diff
@@ -304,13 +376,15 @@ dirserv_router_get_status(const routerinfo_t *router, const char **msg,
 /** Return true if there is no point in downloading the router described by
  * <b>rs</b> because this directory would reject it. */
 int
-dirserv_would_reject_router(const routerstatus_t *rs)
+dirserv_would_reject_router(const routerstatus_t *rs,
+                            const vote_routerstatus_t *vrs)
 {
   uint32_t res;
+  struct ed25519_public_key_t pk;
+  memcpy(&pk.pubkey, vrs->ed25519_id, ED25519_PUBKEY_LEN);
```
dgoulet-tor: Lets use `ed25519_pubkey_copy()`.

src/feature/dirauth/process_descs.c
nmathewson: I'm adding a fixup commit for this, but I don't know if it's what you want.  If it isn't, could you suggest some text that explains the issue better?

src/feature/dirauth/process_descs.c
nmathewson: We can't: ed25519_pubkey_copy() takes two ed25519_public_key_t pointers, but one of the arguments to this memcpy is a raw uint8_t array.

PR #1632, opened 2019-12-26
#32846:Tor Manual:Alphabetize Client Options

swati-thacker13:
Alphabetized options in the Client option category, with the exception of a few options that may not seem to be alphabetically sorted because they made more sense being elsewhere. I have added notes in anchors for these exceptions. 

Please review this ticket.


tlyu:
New pull request #1643

PR #1639, opened 2020-01-07
[Tor Trac #32363]: Check for leading zeros in tor_inet_aton()

neelchauhan:
Ticket: https://trac.torproject.org/projects/tor/ticket/32363

### DIFF
--- a/src/lib/net/inaddr.c
+++ b/src/lib/net/inaddr.c
@@ -11,7 +11,9 @@
 #include "lib/net/inaddr.h"
 
 #include "lib/cc/torint.h"
+#include "lib/container/smartlist.h"
 #include "lib/log/util_bug.h"
+#include "lib/malloc/malloc.h"
 #include "lib/net/inaddr_st.h"
 #include "lib/string/compat_ctype.h"
 #include "lib/string/compat_string.h"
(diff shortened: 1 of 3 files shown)


src/lib/net/inaddr.c
```diff
@@ -39,8 +39,31 @@ tor_inet_aton(const char *str, struct in_addr *addr)
...
+   * heap.
+   *
+   * Here, we run until idx <= strlen(str) so we don't skip the last character
+   * and therefore can process it carefully. However, we cannot do a buffer
+   * overread because idx == strlen(str) does not read further.
+   *
+   * Also, the tor_sscanf() call above prevents an overflow from occuring. */
+  for (size_t idx = 0; idx <= strlen(str); idx++) {
+    if (str[idx] == '.' || idx == strlen(str)) {
+      if (strlen(octet) > 1 && octet[0] == '0')
```
dgoulet-tor: `octet` contains junk at the start of the loop so what happens if we end up here at the first iteration where `octet` was never set?

neelchauhan: The tor_sscanf() check should prevent this.

dgoulet-tor: Ok I think I know what you mean as in we can only get in that `if()` with a malformed address and the sscanf doesn't allow that.

I would really love to still do have `char octet[4] = {0};` just for safety and future proofing this. As in setting it to zeroes at first.

PR #1644, opened 2020-01-09
Ticket32487

nmathewson:


doc/HACKING/Module.md
```diff
@@ -11,12 +11,18 @@ selectively enable or disable, at `configure` time.
...
-dirauth is located in its own directory in `src/feature/dirauth/`.
+The dirauth code is located in its own directory in `src/feature/dirauth/`.
 
-Relay is located in directories named `src/*/*relay` and `src/*/*dircache`,
-which are being progressively refactored and disabled.
+The relay code is located in a directory named `src/*/*relay`, which is
+being progressively refactored and disabled.
+
+The dircache code is located in `src/*/*dircache`.  Right now, it is
+enabled if and only if the relay module is disabled.  (We are treating
```
teor2345:
```suggestion
disabled if and only if the relay module is disabled.  (We are treating
```

src/core/mainloop/fooz
```diff
@@ -0,0 +1,10 @@
+
```
teor2345: This file appears to have been mistakenly committed.

src/feature/dircache/dircache_stub.c
```diff
@@ -0,0 +1,152 @@
...
+
+int
+connection_dirserv_flushed_some(dir_connection_t *conn)
+{
+  (void) conn;
+  tor_assert_nonfatal_unreached_once();
+  return -1;
+}
+
+int
```
teor2345: We usually stub out functions like this using macros, because it lets the optimiser eliminate more dead code. This comment applies to all the stub functions that return a value, particularly if they have no side effects.

(The functions with side effects could also become inline functions in the header.)

For example:
https://github.com/torproject/tor/blob/master/src/feature/relay/routermode.h#L28

src/feature/dircache/dircache_stub.c
```diff
@@ -0,0 +1,152 @@
...
+directory_fetches_from_authorities(const or_options_t *options)
+{
+  (void) options;
+  return 0;
+}
+
+int
+directory_fetches_dir_info_early(const or_options_t *options)
```
teor2345: `FetchDirInfoEarly` is listed as a general option in the man page. It does work on clients, and it's used by some custom client setups (like bandwidth scanners, and other directory monitors).

I think it's ok to disable this option when dircache mode is disabled, but we need to document that in the man page. Maybe it should be moved to an "Advanced Options" section, with a note that these options don't work on clients, when compiled with dircache mode disabled?

src/feature/dircache/dircache_stub.c
```diff
@@ -0,0 +1,152 @@
...
+{
+  (void) options;
+  return 0;
+}
+
+int
```
teor2345: Despite its name, this is actually a client function, which depends on the value of `UseBridges`.

src/feature/dircache/dircache_stub.c
teor2345: (This function calls `directory_fetches_from_authorities()`, which checks the value of `FetchDirInfoEarly`.)

src/feature/dircache/dircache_stub.c
```diff
@@ -0,0 +1,152 @@
...
+directory_permits_begindir_requests(const or_options_t *options)
+{
+  (void) options;
+  return 0;
+}
+
+int
+directory_too_idle_to_fetch_descriptors(const or_options_t *options,
```
teor2345: Despite its name, this is actually a client function:
* it depends on `rep_hist_circbuilding_dormant()`, which is a client function, and
* it depends on `FetchUselessDescriptors`, which is a general option in the man page, but probably belongs in an "Advanced Options" section, like `FetchDirInfoEarly`.

teor2345: (Unlike `FetchDirInfoEarly`, disabling the dircache module doesn't actually disable `FetchUselessDescriptors`, but I wouldn't mind if it did.)

src/feature/dircache/dircache_stub.c
teor2345: These comments also apply to `FetchDirInfoExtraEarly`.

configure.ac
```diff
@@ -289,6 +289,14 @@ AM_COND_IF(BUILD_MODULE_RELAY,
            AC_DEFINE([HAVE_MODULE_RELAY], [1],
                      [Compile with Relay feature support]))
 
+dnl Dircache module.  (This cannot be enabled or disabled independently of
```
teor2345: Please add a note that this module isn't listed in `--list-modules`.

teor2345: We also don't have a `have_module_dircache()` function. I don't mind either way, but we should either have the function, or have a comment in `dircache.h` explaining why it doesn't exist.

We don't check for dircache in `list_enabled_modules()`. That's actually a good thing, because if we did, we'd have to make a large number of changes to the test cases in `test_parseconf.sh`. Let's add comments in `list_enabled_modules()` and `test_parseconf.sh` explaining why the dircache module isn't listed.

If you do want to list the dircache module, please make test_parseconf.sh ignore the dircache module, so we don't have to modify a whole bunch of tests. (By adding `dircache_` to the `relay_` disabled module file names.)

src/feature/dircache/dircache_stub.c
teor2345: There are a bunch of related options used by `should_use_directory_guards()`:
* UseEntryGuards
* DownloadExtraInfo
* FetchDirInfoEarly
* FetchDirInfoExtraEarly
* FetchUselessDescriptors
https://github.com/torproject/tor/blob/4f02812242d1fd90d859eb98ac3fb1ed182f18cf/src/feature/dirclient/dirclient.c#L310

For consistency, we should probably either keep all these options, or disable all of them. If we decide to disable them, let's do it in a separate ticket, and make `directory_fetches_dir_info_early()` a dircommon function for now?

src/feature/dircache/dircache_stub.c
teor2345: But if we decide to disable `FetchUselessDescriptors` when dircache is disabled, let's do it in a separate ticket.

src/core/mainloop/fooz
nmathewson: Removed in 8df602ebb702b1

doc/HACKING/Module.md
nmathewson: Fixed in 44dd6074ff560b

src/app/config/config.c
```diff
@@ -2733,6 +2733,9 @@ list_enabled_modules(void)
 {
   printf("%s: %s\n", "relay", have_module_relay() ? "yes" : "no");
   printf("%s: %s\n", "dirauth", have_module_dirauth() ? "yes" : "no");
+  // We don't list dircache, because it cannot be enabled or disabled
+  // independently from relay.  Listing it here would proliferate
+  // tests in test_parseconf.c to no useful purpose.
```
teor2345: It's actually the parseconf script, not the unit tests :-)
```suggestion
  // test variants in test_parseconf.sh to no useful purpose.
```

PR #1647, opened 2020-01-09
Ticket32709 043 01

dgoulet-tor:


src/feature/hs/hs_ob.c
```diff
@@ -0,0 +1,182 @@
...
+
+  for (line = opts->MasterOnionAddress; line; line = line->next) {
+    /* Allocate config list if need be. */
+    if (!config->ob_master_pubkeys) {
+      config->ob_master_pubkeys = smartlist_new();
+    }
+    uint8_t version = 0;
+    ed25519_public_key_t *pubkey = tor_malloc_zero(sizeof(*pubkey));
+
+    if (hs_parse_address(line->value, pubkey, NULL, &version) < 0) {
```
asn-d6: I get an assertion failure here:
```
[err] Bug: Tor 0.4.3.0-alpha-dev (git-77ef50d86b646f63): Assertion global_options failed in get_options_mutable at . Stack trace: (on Tor 0.4.3.0-alpha-dev 77ef50d86b646f63)
[err] Bug:     tor(log_backtrace_impl+0x56) [0x564e089aa2c6] (on Tor 0.4.3.0-alpha-dev
[err] Bug:     tor(tor_assertion_failed_+0x148) [0x564e089a5408] (on Tor 0.4.3.0-alpha-dev
[err] Bug:     tor(get_options_mutable+0x62) [0x564e0891b302] (on Tor 0.4.3.0-alpha-dev
[err] Bug:     tor(escaped_safe_str+0x1d) [0x564e0891b7cd] (on Tor 0.4.3.0-alpha-dev
[err] Bug:     tor(hs_parse_address+0xf8) [0x564e088a4748] (on Tor 0.4.3.0-alpha-dev
[err] Bug:     tor(hs_ob_parse_config_file+0x193) [0x564e0893ba83] (on Tor 0.4.3.0-alpha-dev
[err] Bug:     tor(hs_config_service_all+0xd6c) [0x564e0893a33c] (on Tor 0.4.3.0-alpha-dev 77ef50d86b646f63)
```

We seem to be trying to do this too early and it's invoking get_options when its not yet ready.

I dirtily fixed this by removing the calls to escaped log, but we need to do something better here.

src/feature/hs/hs_cell.c
```diff
@@ -853,17 +853,46 @@ hs_cell_parse_introduce2(hs_cell_introduce2_data_t *data,
...
+
+    SMARTLIST_FOREACH_BEGIN(service->config.ob_master_pubkeys,
+                            const ed25519_public_key_t *, pkey) {
+      ed25519_public_key_t blinded_pubkey;
+      uint8_t subcredential[DIGEST256_LEN];
+
+      hs_build_blinded_pubkey(pkey, NULL, 0, hs_get_time_period_num(0),
+                              &blinded_pubkey);
+      hs_get_subcredential(pkey, &blinded_pubkey, subcredential);
+      memcpy(&new_data, data, sizeof(new_data));
```
asn-d6: OK this works just fine in Chutney! However, it seems like `data->client_pk` is zeroed out and that makes `hs_ntor_service_get_rendezvous1_keys()` fail in:
```
  curve25519_handshake(dh_result1,
                       &service_ephemeral_rend_keypair->seckey,
                       client_ephemeral_enc_pubkey);
```

I'm not sure what's going on but I fixed this by commenting out the `memwipe(&data->client_pk, 0, sizeof(curve25519_public_key_t));` in `hs_cell_parse_introduce2()`. It worked after this and I managed to actually connect to onionbalance and pass data around.

src/feature/hs/hs_ob.c
```diff
@@ -0,0 +1,182 @@
...
+      log_warn(LD_REND, "OnionBalance: MasterOnionAddress %s is invalid",
```
asn-d6: This seems to expect the address without the ".onion". Can we perhaps tolerate the .onion as well, since that might be more natural to the operators?

src/feature/hs/hs_ob.c
```diff
@@ -0,0 +1,182 @@
...
+               line->value);
+      tor_free(pubkey);
+      goto end;
+    }
+    smartlist_add(config->ob_master_pubkeys, pubkey);
```
asn-d6: Nitpick: Let's add a log message here saying that it got registered?

src/feature/hs/hs_service.h
```diff
@@ -248,10 +248,14 @@ typedef struct hs_service_config_t {
...
 
-  /* DoS defenses. For the ESTABLISH_INTRO cell extension. */
+  /** DoS defenses. For the ESTABLISH_INTRO cell extension. */
   unsigned int has_dos_defense_enabled : 1;
   uint32_t intro_dos_rate_per_sec;
   uint32_t intro_dos_burst_per_sec;
+
+  /** If set, contains the Onion Balance master public key (.onion addresses)
+   * that this service serves as backend. */
+  smartlist_t *ob_master_pubkeys;
```
asn-d6: Do we need to free this somewhere?

src/feature/hs/hs_cell.c
```diff
@@ -853,17 +853,46 @@ hs_cell_parse_introduce2(hs_cell_introduce2_data_t *data,
...
+                                                      encrypted_section_len);
+      if (intro_keys) {
+        break;
+      }
+    } SMARTLIST_FOREACH_END(pkey);
+  } else {
+    /* Get the right INTRODUCE2 ntor keys and verify the cell MAC */
+    intro_keys = get_introduce2_keys_and_verify_mac(data, encrypted_section,
+                                                    encrypted_section_len);
+    if (!intro_keys) {
```
asn-d6: Perhaps this (!intro_keys) case can be moved out of the if clause so that it warns in case the MAC didnt line up in the OB case either.

src/feature/hs/hs_cell.c
```diff
@@ -853,17 +853,46 @@ hs_cell_parse_introduce2(hs_cell_introduce2_data_t *data,
...
+  memcpy(&data->client_pk.public_key, encrypted_section,
+         CURVE25519_PUBKEY_LEN);
+
+  /* If we are configured as an Onion Balance instance, we need to try
+   * validation with the configured master public keys given in the config
+   * file. This is because the master identity key and the blinded key is put
+   * in the INTRODUCE2 cell by the client thus it will never validate with
+   * this instance public key. */
+  if (service->config.ob_master_pubkeys) {
+    hs_cell_introduce2_data_t new_data;
```
asn-d6: Nitpick: Can we move this whole block to the OB subsystem we just made?

src/feature/hs/hs_cell.c
```diff
@@ -853,17 +853,46 @@ hs_cell_parse_introduce2(hs_cell_introduce2_data_t *data,
...
+
+    SMARTLIST_FOREACH_BEGIN(service->config.ob_master_pubkeys,
```
asn-d6: I think we might need to iterate over the three possible timeperiods (-1, 0, +1) as we discussed in jitsi. Consider this:

Client has the 23:00 consensus and backend service has the 01:00 consensus. This means that Client is in TP #N-1 and Service is in TP #N. Hence when it does `hs_get_time_period_num(0)` here it will not synch with what the client does.

What do you think?

src/feature/hs/hs_cell.c
asn-d6: Ah the bug here seems to be introduced by the memwipe I added in the refactoring commit ee3f36476a87cab24881d56b8c9227c80af19521 . Perhaps we can kill that commit? Or fix it? Up to you.

dgoulet-tor: Agree. I've removed it for now in the rebased version of this branch.

src/feature/hs/hs_ob.c
dgoulet-tor: We'll have to do a special `hs_parse_address()` that returns an error string instead of directly logging. So then, we can use the `options_t` that `hs_config_service_all()` has accessed to.

src/feature/hs/hs_ob.c
dgoulet-tor: Time for a `hs_parse_address_with_onion` type of function :)

src/feature/hs/hs_cell.c
```diff
@@ -747,6 +747,61 @@ hs_cell_parse_intro_established(const uint8_t *payload, size_t payload_len)
...
+    /* The MAC field is at the very end of the ENCRYPTED section. */
+    size_t mac_offset = encrypted_section_len - sizeof(mac);
+    /* Compute the MAC. Use the entire encoded payload with a length up to the
+     * ENCRYPTED section. */
+    compute_introduce_mac(data->payload,
+                          data->payload_len - encrypted_section_len,
+                          encrypted_section, encrypted_section_len,
+                          intro_keys->mac_key, sizeof(intro_keys->mac_key),
+                          mac, sizeof(mac));
+    if (tor_memcmp(mac, encrypted_section + mac_offset, sizeof(mac))) {
```
nmathewson: This is not a new bug, but the code here should be using tor_memneq().  tor_memcmp() is only used when we actually want a tristate result.

src/feature/hs/hs_cell.c
```diff
@@ -747,6 +747,61 @@ hs_cell_parse_intro_established(const uint8_t *payload, size_t payload_len)
...
+             "compute key material");
+    goto err;
+  }
+
+  /* Validate MAC from the cell and our computed key material. The MAC field
+   * in the cell is at the end of the encrypted section. */
+  {
+    uint8_t mac[DIGEST256_LEN];
+    /* The MAC field is at the very end of the ENCRYPTED section. */
+    size_t mac_offset = encrypted_section_len - sizeof(mac);
```
nmathewson: I'd be more comfortable if we explicitly checked that encrypted_section_len was at least sizeof(mac).

src/app/config/config.c
```diff
@@ -517,6 +517,8 @@ static const config_var_t option_vars_[] = {
       LINELIST_S, RendConfigLines, NULL),
   VAR("HiddenServiceEnableIntroDoSBurstPerSec",
       LINELIST_S, RendConfigLines, NULL),
+  VAR("HiddenServiceOnionBalanceConfigFile",
```
nmathewson: Design question: why is this a separate configuration file, instead of a set of options for use with the service?

src/feature/hs/hs_ob.c
```diff
@@ -0,0 +1,220 @@
...
+  .vars = config_vars,
+  .extra = &config_extra_vars,
+};
+
+/* Global configuration manager for the config file. */
+static config_mgr_t *config_options_mgr = NULL;
+
+/* Return the configuration manager for the config file. */
+static const config_mgr_t *
+get_config_options_mgr(void)
```
nmathewson: This name (and the other names in this file) should probably be ob-specific, if we decide to keep this as a separate configuration file.

src/feature/hs/hs_ob.c
```diff
@@ -0,0 +1,220 @@
...
+  tor_assert(value);
+  tor_assert(pkey_out);
+
+  extension = strrchr(value, '.');
+  if (!extension) {
+    /* No .onion extension, bad format. */
+    return false;
+  }
+
+  if (strcmp(extension, ".onion")) {
```
nmathewson: I think you could replace the last two checks with `strcmpend(value, ".onion")`

src/app/config/config.c
nmathewson: For example, it seems like it would make more sense to have this be a file at a fixed location in the hidden service directory, maybe with a boolean in the torrc to turn it on.

src/feature/hs/hs_service.h
```diff
@@ -248,10 +248,14 @@ typedef struct hs_service_config_t {
...
   hs_circuit_id_protocol_t circuit_id_protocol;
 
-  /* DoS defenses. For the ESTABLISH_INTRO cell extension. */
+  /** DoS defenses. For the ESTABLISH_INTRO cell extension. */
   unsigned int has_dos_defense_enabled : 1;
   uint32_t intro_dos_rate_per_sec;
   uint32_t intro_dos_burst_per_sec;
+
+  /** If set, contains the Onion Balance master public key (.onion addresses)
+   * that this service serves as backend. */
```
nmathewson: We need to document what type the elements are. (Ed25519 keys? strings?)

src/feature/hs/hs_cell.c
```diff
@@ -856,14 +898,36 @@ hs_cell_parse_introduce2(hs_cell_introduce2_data_t *data,
...
-  intro_keys = get_introduce2_keys_and_verify_mac(data, encrypted_section,
-                                                  encrypted_section_len);
+  /* First bytes of the ENCRYPTED section are the client public key (they are
+   * guaranteed to exist because of the length check above). We are gonna use
+   * the client public key to compute the ntor keys and decrypt the payload:
+   */
+  memcpy(&data->client_pk.public_key, encrypted_section,
+         CURVE25519_PUBKEY_LEN);
+
+  /* If we are configured as an Onion Balance instance, we need to try
```
nmathewson: I think it would be cleaner to have a single function that does this operation for us whether we are an OB instance or not.

src/feature/hs/hs_ob.c
```diff
@@ -172,6 +174,25 @@ ob_option_parse(hs_service_config_t *config, const ob_options_t *opts)
   return ret;
 }
 
+/** For the given master public key and time period, compute the subcredential
+ * and put them into subcredential. */
```
nmathewson: Let's document how big subcredential should be.

src/feature/hs/hs_ob.c
```diff
@@ -218,3 +239,54 @@ hs_ob_parse_config_file(hs_service_config_t *config, const char *fname)
...
   return ret;
 }
+
+/** Compute all possible subcredentials for every onion master key in the
+ * given service config object. The subcredentials is allocated and set as an
+ * continous array containing all possible values.
+ *
+ * The number of subcredential set in the array is returned (length). If the
+ * configuration has no OB master keys configured, 0 is returned and
+ * subcredentials is set to NULL. */
```
nmathewson: Let's document the size of the output array.

src/feature/hs/hs_ob.c
```diff
@@ -218,3 +239,54 @@ hs_ob_parse_config_file(hs_service_config_t *config, const char *fname)
...
+  num_pkeys = smartlist_len(config->ob_master_pubkeys);
+  if (!num_pkeys) {
+    *subcredentials = NULL;
+    goto end;
+  }
+
+  /* Next loop will build 3 subcredentials for each public key. */
+  *subcredentials = tor_malloc_zero(DIGEST256_LEN * 3 * num_pkeys);
+
+  /* Time to build all the subcredentials for each time period: the previous
```
nmathewson: Is this three-subcredentials thing only done for the onionbalance case?  If so, why?  If not, can we share the logic with whatever else does it?

src/feature/hs/hs_ob.c
```diff
@@ -218,3 +239,54 @@ hs_ob_parse_config_file(hs_service_config_t *config, const char *fname)
...
+   * to try them all.
+   *
+   * As a slight optimization, statistically, the current time period (0) will
+   * be the one to work first so we'll put them first in the array to maximize
+   * our chance of success. */
+
+  for (unsigned int i = 0; i <= sizeof(steps); i++) {
+    SMARTLIST_FOREACH_BEGIN(config->ob_master_pubkeys,
+                            const ed25519_public_key_t *, pkey) {
+      build_subcredential(pkey, tp + steps[i], subcredentials[idx]);
```
nmathewson: I'm pretty sure this is going to crash.  You're treating subcredentials as an array of subcredentials, but you allocated an array of uint8_t above.

src/feature/hs/hs_cell.c
```diff
@@ -802,6 +803,47 @@ get_introduce2_keys_and_verify_mac(hs_cell_introduce2_data_t *data,
...
+    /* We are _not_ an OB instance since no configured master onion key(s)
+     * were found and thus no subcredentials were built. */
+    goto end;
+  }
+
+  for (size_t idx = 0; idx < ob_num_subcreds; idx++) {
+    /* Copy current data into a new INTRO2 cell data. We will then change the
+     * subcredential in order to validate. */
+    hs_cell_introduce2_data_t new_data = *data;
+    new_data.subcredential = &ob_subcreds[idx];
```
nmathewson: ob_subcreds[idx] will be the resulting array at position 0, 1, and 2.  But I think you want 0, 1*sizeof a subcredential, 2*sizeof a subcredential.

This makes me worried about testing!  I bet nobody has tested the case where the matching subcredential is not the first one.

src/feature/hs/hs_ob.c
asn-d6: That's a good question. Let me try to answer it:

The reason we are not doing this brute force in the normal hsv3 case, is because the service always knows the right subcredential to use:

In `service_handle_introduce2()` the service gets the descriptor object that corresponds to this intro point, and because each descriptor corresponds to a specific subcredential, we get the right subcredential out of it, and use that to do the decryption. So that works just fine now that each descriptor has different intro points, however we will probably need to change this logic when we do #22893 (potentially by giving a different intro auth key to the same intro point in diff descriptors, so that we can still distinguish which descriptor it came off).

src/feature/hs/hs_cell.c
```diff
@@ -776,6 +776,12 @@ get_introduce2_keys_and_verify_mac(hs_cell_introduce2_data_t *data,
    * in the cell is at the end of the encrypted section. */
   {
     uint8_t mac[DIGEST256_LEN];
+
+    /* Make sure we are now about to underflow. */
```
asn-d6: s/now/not/

src/feature/hs/hs_cell.c
```diff
@@ -802,6 +803,47 @@ get_introduce2_keys_and_verify_mac(hs_cell_introduce2_data_t *data,
...
+get_intro2_keys_as_ob(const hs_service_config_t *config,
+                      const hs_cell_introduce2_data_t *data,
+                      const uint8_t *encrypted_section,
+                      size_t encrypted_section_len)
+{
+  uint8_t *ob_subcreds = NULL;
+  size_t ob_num_subcreds;
+  hs_ntor_intro_cell_keys_t *intro_keys = NULL;
+
+  ob_num_subcreds = hs_ob_get_subcredentials(config, &ob_subcreds);
```
asn-d6: This is a design comment and perhaps a premature optimization, but given that we worry about performance, memory allocation and whatnot I'm gonna mention it. It seems non-optimal that we calculate all the subcredentials (which involves key blinding) even if we don't need them in the end. A more optimized approach would be to create a subcredential and then try it out, then go to the next one, etc. until we find the right one.

src/feature/hs/hs_ob.c
```diff
@@ -220,3 +242,78 @@ hs_ob_parse_config_file(hs_service_config_t *config)
...
+   * total number of keys we are about to process. In other words, for each
+   * key, we allocate 3 subcredential slots. */
+  subcreds = tor_malloc_zero(subcred_len * num_steps * num_pkeys);
+
+  /* For each time period step. */
+  for (unsigned int i = 0; i < num_steps; i++) {
+    SMARTLIST_FOREACH_BEGIN(config->ob_master_pubkeys,
+                            const ed25519_public_key_t *, pkey) {
+      build_subcredential(pkey, tp + steps[i],
+                          &(subcreds[idx * subcred_len]));
```
asn-d6: Can we use `i` instead of `idx` here? Isn't it the same?

src/feature/hs/hs_cell.c
asn-d6: FWIW, I manually tested this with chutney and OB and it works alright. I did that by changing the subcredential order so that position 0 is not first, and then tested a bunch of times.

src/feature/hs/hs_ob.c
```diff
@@ -0,0 +1,337 @@
+/* Copyright (c) 2017-2020, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
+ * \file hs_ob.c
+ * \brief Implement Onion Balance specific code.
+ *
+ * \details
+ *
+ * XXX:
```
asn-d6: There is an XXX here.

src/feature/hs/hs_cell.c
nmathewson: This still looks like a bug?

src/feature/hs/hs_cell.c
nmathewson: It seems to me that we'd even be better off pre-computing all our subcredentials periodically, and not re-doing it every time we're doing a handshake.

src/feature/hs/hs_cell.c
```diff
@@ -802,6 +803,47 @@ get_introduce2_keys_and_verify_mac(hs_cell_introduce2_data_t *data,
...
+    /* We are _not_ an OB instance since no configured master onion key(s)
+     * were found and thus no subcredentials were built. */
+    goto end;
+  }
+
+  for (size_t idx = 0; idx < ob_num_subcreds; idx++) {
+    /* Copy current data into a new INTRO2 cell data. We will then change the
+     * subcredential in order to validate. */
+    hs_cell_introduce2_data_t new_data = *data;
+    new_data.subcredential = &(ob_subcreds[idx * DIGEST256_LEN]);
```
nmathewson: This is less crashy than before, but it's still a little tricky to check.  Maybe we should have a struct subcredential_t, and then use an array of that, rather than an array of char and a bunch of digest calculations?

scripts/maint/practracker/exceptions.txt
```diff
@@ -250,7 +250,7 @@ problem function-size /src/feature/hs/hs_cell.c:hs_cell_parse_introduce2() 152
 problem function-size /src/feature/hs/hs_client.c:send_introduce1() 103
 problem function-size /src/feature/hs/hs_client.c:hs_config_client_authorization() 107
 problem function-size /src/feature/hs/hs_common.c:hs_get_responsible_hsdirs() 102
-problem function-size /src/feature/hs/hs_config.c:config_service_v3() 107
+problem function-size /src/feature/hs/hs_config.c:config_service_v3() 128
```
nmathewson: Okay, but wouldn't it be better to split this up, or introduce helper functions, or something?

PR #1651, opened 2020-01-11
[Tor Trac #30642]: Add an ed25519_identity file to the data directory

neelchauhan:
For the ticket [here](https://trac.torproject.org/projects/tor/ticket/30642).

### DIFF
--- a/src/app/main/main.c
+++ b/src/app/main/main.c
@@ -1033,12 +1033,14 @@ sandbox_init_filter(void)
 
     OPEN_DATADIR("approved-routers");
     OPEN_DATADIR_SUFFIX("fingerprint", ".tmp");
+    OPEN_DATADIR_SUFFIX("ed25519_identity", ".tmp");
     OPEN_DATADIR_SUFFIX("hashed-fingerprint", ".tmp");
     OPEN_DATADIR_SUFFIX("router-stability", ".tmp");
 
     OPEN("/etc/resolv.conf");
 
     RENAME_SUFFIX("fingerprint", ".tmp");
+    RENAME_SUFFIX("ed25519_identity", ".tmp");
     RENAME_KEYDIR_SUFFIX("secret_onion_key_ntor", ".tmp");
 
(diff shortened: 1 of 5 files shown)


src/app/main/main.c
```diff
@@ -1033,12 +1033,14 @@ sandbox_init_filter(void)
 
     OPEN_DATADIR("approved-routers");
     OPEN_DATADIR_SUFFIX("fingerprint", ".tmp");
+    OPEN_DATADIR_SUFFIX("ed25519_identity", ".tmp");
```
teor2345: Maybe this should be "ed25519-identity", to match "hashed-fingerprint" and "router-stability"?
(It appears we use underscores for keys, and dashes for user-readable text files.)

Let's see what @mikeperry-tor thinks.

src/feature/relay/router.c
```diff
@@ -881,6 +881,43 @@ router_write_fingerprint(int hashed)
   return result;
 }
 
+STATIC int
+router_write_ed25519_identity(void)
```
teor2345: Rather than copying most of the code from router_write_fingerprint(), you could add an `int ed25519_identity` to router_write_fingerprint(), and re-use most of the code in that function?

src/feature/relay/router.c
```diff
@@ -829,30 +829,36 @@ router_initialize_tls_context(void)
  * -1 if Tor should die,
  */
 STATIC int
-router_write_fingerprint(int hashed)
+router_write_fingerprint(int hashed, int ed25519_identity)
 {
   char *keydir = NULL, *cp = NULL;
   const char *fname = hashed ? "hashed-fingerprint" :
-                               "fingerprint";
+                      (ed25519_identity ? "ed25519_identity" : "fingerprint");
```
dgoulet-tor: Not sure you can do that. What happens if both `hashed = 1` and `ed25519_identity = 1 ` ?

I think you need to keep the `hashed` as it is and then individually test `ed25519_identity` after so you can write more than one file.

PR #1653, opened 2020-01-12
[Tor Trac #32614]: Flag a v3 intro point as bad if rendezvous fails during cleanup

neelchauhan:
Ticket: https://trac.torproject.org/projects/tor/ticket/32614

### DIFF
--- a/src/feature/hs/hs_client.c
+++ b/src/feature/hs/hs_client.c
@@ -1741,12 +1741,16 @@ hs_client_circuit_cleanup_on_free(const circuit_t *circ)
   bool has_timed_out;
   rend_intro_point_failure_t failure = INTRO_POINT_FAILURE_GENERIC;
   const origin_circuit_t *orig_circ = NULL;
+  origin_circuit_t *intro_circ = NULL;
 
   tor_assert(circ);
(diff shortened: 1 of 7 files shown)


src/feature/hs/hs_client.c
```diff
@@ -1741,13 +1741,17 @@ hs_client_circuit_cleanup_on_free(const circuit_t *circ)
...
 
   tor_assert(circ);
   tor_assert(CIRCUIT_IS_ORIGIN(circ));
 
   orig_circ = CONST_TO_ORIGIN_CIRCUIT(circ);
   tor_assert(orig_circ->hs_ident);
 
+  intro_circ = hs_circuitmap_get_intro_circ_v3_service_side(
+                                          &orig_circ->hs_ident->intro_auth_pk);
+
```
dgoulet-tor: There is quite a bit of wrong with this. This is _service_ side, you can't use the HS circuit map to lookup a service side circuit on the client side.

We actually don't register the client introduction circuit into the map. Also, you assume `intro_circ` is never NULL here which can't be guaranteed.

This needs to be tested properly...

PR #1664, opened 2020-01-15
Ticket32910 044 01

dgoulet-tor:


src/lib/trace/debug.h
```diff
@@ -17,14 +19,20 @@
 
 /* Send every event to a debug log level. This is useful to debug new trace
  * events without implementing them for a specific event tracing framework.
- * Note that the arguments are ignored since at this step we do not know the
- * types and amount there is. */
+ *
+ * NOTE: arguments can't be used becaue there is no easy generic ways to learn
```
nmathewson: Could we take a format string and a bunch of args, and pass them into log_debug?   We'll have to concatenate the format string to the string we're passing in now.

configure.ac
```diff
@@ -251,6 +251,20 @@ AC_ARG_ENABLE(libscrypt,
 
 dnl --- Tracing Options. ---
 
+dnl USDT instrumentation option.
+AC_ARG_ENABLE(tracing-instrumentation-usdt,
+              AS_HELP_STRING([--enable-tracing-instrumentation-usdt],
+                             [build with tracing USDT instrumentation]))
```
nmathewson: Somewhere, we should define what USDT is.

src/lib/trace/lttng/provider_circuit.h
```diff
@@ -0,0 +1,150 @@
...
+ *        circuit subsystem.
+ **/
+
+#undef TRACEPOINT_PROVIDER
+#define TRACEPOINT_PROVIDER tor_circuit
+
+#undef TRACEPOINT_INCLUDE
+#define TRACEPOINT_INCLUDE "./lib/trace/lttng/provider_circuit.h"
+
+#if !defined(_TRACEPOINT_TOR_CIRCUIT_H) || defined(TRACEPOINT_HEADER_MULTI_READ)
```
nmathewson: We shouldn't define macros starting with _; those are reserved.

Also, why do we need this strange include guard?  If we can't do without it, we should have a comment.

nmathewson: Also, what marks code tracepoint as lttng only?  It seems  to depend on lttng, but I don't see the "ifdef" for that.

src/lib/trace/lttng/providers.h
```diff
@@ -0,0 +1,17 @@
+/* Copyright (c) 2020, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
+ * \file providers.h
```
nmathewson: I'd like to avoid a single all-powerful providers.h header; subsystems should just include their own if they need it, right?

src/lib/trace/lttng/provider_circuit.h
```diff
@@ -0,0 +1,150 @@
...
+#define _TRACEPOINT_TOR_CIRCUIT_H
+
+#include <lttng/tracepoint.h>
+
+#include "core/or/circuitlist.h"
+#include "core/or/or.h"
+#include "core/or/origin_circuit_st.h"
+
+TRACEPOINT_ENUM(tor_circuit, purpose,
```
nmathewson: For every single possible event we should have comments explaining what it means.

src/lib/trace/lttng/lttng.h
```diff
@@ -14,8 +14,11 @@
 #include <lttng/tracepoint.h>
 
 /* Map event to an LTTng tracepoint. */
-#define TOR_TRACE_LTTNG(subsystem, event_name, ...) \
-  tracepoint(subsystem, event_name, ## __VA_ARGS__);
+#define TOR_TRACE_LTTNG(subsystem, event_name, args...) \
+  tracepoint(tor_##subsystem, event_name, args)
```
nmathewson: Is this new tor_##subsystem header a global identifier?  If so I would prefer if the prefix were something more tracing-specific:  Like, tor_trc_ instead of tor_.

src/core/or/circuituse.c
```diff
@@ -1680,6 +1681,7 @@ circuit_testing_failed(origin_circuit_t *circ, int at_last_hop)
 void
 circuit_has_opened(origin_circuit_t *circ)
 {
+  tor_trace(circuit, opened, circ);
```
nmathewson: Is this typechecked only when building with lttng?

src/lib/trace/lttng/provider_circuit.h
```diff
@@ -0,0 +1,150 @@
...
+    ctf_enum(tor_circuit, purpose, int, purpose, TO_CIRCUIT(circ)->purpose)
+    ctf_enum(tor_circuit, state, int, state, TO_CIRCUIT(circ)->state)
+  )
+)
+
+/*
+ * Origin circuit events.
+ */
+
+TRACEPOINT_EVENT_INSTANCE(tor_circuit, origin_circuit_t_class, opened,
```
nmathewson: The type "origin_circuit_t" is not visible at this point; it is defined in src/core.  If we require it here, then nothing lower level than src/core will be able to use this header.  Similarly it's problematic to include field access here, since those fields are not visible to src/lib/trace.

I think we need to move this header into src/core.  That may mean some refactoring of the tracing system.

{Same comment for uses of circuit in the rest of this file}

src/lib/trace/trace.c
```diff
@@ -1,4 +1,4 @@
-/* Copyright (c) 2017-2020, The Tor Project, Inc. */
+/* Copyright (c) 2020, The Tor Project, Inc. */
```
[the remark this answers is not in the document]
dgoulet-tor: Hmmm I'll set it back.

src/lib/trace/trace_sys.c
```diff
@@ -0,0 +1,33 @@
...
+static void
+subsys_tracing_shutdown(void)
+{
+  tor_trace_free_all();
+}
+
+const subsys_fns_t sys_tracing = {
+  .name = "tracing",
+  .supported = true,
+  .level = -100,
```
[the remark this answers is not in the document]
dgoulet-tor: That is hella useful to know that scripts! Awesome.

src/lib/trace/debug.h
dgoulet-tor: Hmmm so the tracepoint the way I made the API is:

`tor_trace(<subsystem_name>, <event_name>, args1, arg2, ...)`. The args1 is the information we want to record. So if we put `fmt` as args1, then all tracers will have to record `fmt`.

configure.ac
dgoulet-tor: Within the configure.ac? I thought of updating the `Tracing.md` file and it would explain all this.

src/lib/trace/lttng/provider_circuit.h
dgoulet-tor: Yeah so those crazy ifdef are required by LTTng-UST: https://lttng.org/docs/v2.11/#doc-tpp-header

But we can easily control the `_TRACEPOIN...` so that I can change.

src/lib/trace/lttng/providers.h
dgoulet-tor: The way I did this is that C file include `lib/trace/events.h` alone and then they have access to all providers (for the case of LTTng). Else, we would have to add another layer that has per subsystem events like `lib/trace/events_circuit.h` type of files.

No strong opinion, whatever you prefer. I went for ease of use?

src/lib/trace/lttng/provider_circuit.h
dgoulet-tor: Agree

src/lib/trace/lttng/lttng.h
dgoulet-tor: It is not within the code. It is with the event we export so consider this a bit of an "ABI" that is the tracing event the LTTng tracer will see here will be prefixes with "tor_circuit:<event_name>" for instance.

But apart from that, they don't become global symbol in the code afaik.

src/core/or/circuituse.c
dgoulet-tor: Ah yes good observation. LTTng does type checking but afaik, not the USDT probes (Dtrace, Systemtap, etc...). These tracers will eat what is given and then you need to define the parameter type outside with some other file specific to DTrace.

So right now, that event can be a USDT probes (if enabled) but the object passed will be considered a simple pointer, nothing gets deref or anything. We need to configure that per tracepoint. I haven't done that job for other tracers since supporting them all is quite some work...

src/lib/trace/lttng/provider_circuit.h
dgoulet-tor: Yup agree... Basically, the tracing subsystem ultimately needs visibility of everything so we can do deep-dive into the objects. It could even be legit to be a `src/feature/` ?

For instance, I have tracepoints for the HS subsystem that are not in core/.

PR #1677, opened 2020-01-20
[#33005] Lower log level of standard error messages from PT's.

ahf:
This patch lowers the log level of error messages from PT processes from
warning to debug.

See: https://bugs.torproject.org/33005

### DIFF
--- a/src/feature/client/transports.c
+++ b/src/feature/client/transports.c
@@ -1858,9 +1858,9 @@ managed_proxy_stderr_callback(process_t *process,
   if (BUG(mp == NULL))
     return;
 
-  log_debug(LD_PT,
-            "Managed proxy at '%s' reported via standard error: %s",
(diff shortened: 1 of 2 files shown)


src/feature/client/transports.c
```diff
@@ -1858,7 +1858,9 @@ managed_proxy_stderr_callback(process_t *process,
   if (BUG(mp == NULL))
     return;
 
-  log_warn(LD_PT, "Managed proxy at '%s' reported: %s", mp->argv[0], line);
+  log_debug(LD_PT,
```
teor2345: Should this be log_info()?

I could imagine us wanting to see the logs when a PT user is having trouble. And if we ask for debug logs, they are usually hundreds of megabytes.

ahf: Good catch. I've lowered it to info.

PR #1681, opened 2020-01-22
Ticket33029 043 01

dgoulet-tor:


src/feature/nodelist/dirlist.c
```diff
@@ -49,6 +49,31 @@ static smartlist_t *trusted_dir_servers = NULL;
...
+{
+  if (BUG(!addr) || BUG(!trusted_dir_servers)) {
+    return false;
+  }
+
+  /* Go over all trusted directories. We have very few of these so this should
+   * be fairly fast. In the future, if this shows up in the profile, we can
+   * move to have an address set instead. */
+  SMARTLIST_FOREACH_BEGIN(trusted_dir_servers, const dir_server_t *, ds) {
+    if ((ds->type & type) && tor_addr_eq_ipv4h(addr, ds->addr)) {
```
teor2345: Please check for IPv6 addresses.

It won't cost much extra CPU time, because address types are checked first, and the volume of IPv6 bugs is low. But it will avoid weird bugs in future, when we reach IPv6 feature parity.

teor2345: If the incoming address is an IPv6 address, you'll want to check dir_server_t's ipv6_addr:
https://github.com/torproject/tor/blob/4f02812242d1fd90d859eb98ac3fb1ed182f18cf/src/feature/dirclient/dir_server_st.h#L27

teor2345: I think these issues are ok, but they need to be talked about in the function comments, in case they cause bugs in future:

You're checking the hard-coded (or configured) list of directory authorities, rather than the current consensus. That's probably ok, but it requires the directory authorities to update their configurations when the set of authority addresses changes. (Usually, if authorities advertise a new address in their descriptors, it gets picked up in the consensus, and used by other directory authorities, without a local config change. But this function is an exception to that rule.)

And you're checking the inbound ORPort address, rather than the OutboundBindAddress. How many directory authorities have different inbound and outbound addresses?

src/feature/nodelist/dirlist.c
```diff
@@ -49,6 +49,31 @@ static smartlist_t *trusted_dir_servers = NULL;
  * and all fallback directory servers. */
 static smartlist_t *fallback_dir_servers = NULL;
 
+/** Return true iff the given v4 address is the one of a configured directory
+ * authority matching the given type.
+ *
+ * This needs to be fast because it can be called at each directory connection
+ * in order to learn if the request is from an authority or not. */
```
teor2345: I think these issues are ok, but they need to be talked about in the function comments, in case they cause bugs in future:

You're checking the hard-coded (or configured) list of directory authorities, rather than the current consensus. That's probably ok, but it requires the directory authorities to update their configurations when the set of authority addresses changes. (Usually, if authorities advertise a new address in their descriptors, it gets picked up in the consensus, and used by other directory authorities, without a local config change. But this function is an exception to that rule.)

And you're checking the inbound ORPort address, rather than the OutboundBindAddress. How many directory authorities have different inbound and outbound addresses?

dgoulet-tor: Right so dirauth do have each authority added to their `torrc` at the moment. But we shouldn't rely on that for sure.

I'm confused here about the `OutboundBindAddress` comment. The code is using the address from the `dir_server_t` so if a dirauth uses a different address for outbound connection instead of the one advertised for inbound, we have a problem indeed.

src/core/mainloop/.may_include
```diff
@@ -21,4 +21,6 @@ lib/compress/*.h
 lib/log/*.h
 
 core/mainloop/*.h
-core/mainloop/*.inc
\ No newline at end of file
+core/mainloop/*.inc
+
+feature/nodelist/dirlist.h
```
nmathewson: If we've got to do this, we should be adding feature/nodelist/*.h

src/feature/nodelist/dirlist.c
```diff
@@ -49,6 +49,31 @@ static smartlist_t *trusted_dir_servers = NULL;
...
+bool
+dirlist_addr_is_trusted_dir_by_type(const tor_addr_t *addr,
```
nmathewson: The documentation for this function should emphasize the point that we should never actually trust this function to mean that we have a real directory authority, since it is just an address check, and IP spoofing is possible.

PR #1685, opened 2020-01-27
Ticket32709 044 02

asn-d6:


src/lib/ctime/di_ops.c
```diff
@@ -279,3 +279,30 @@ select_array_member_cumulative_timei(const uint64_t *entries, int n_entries,
...
+ *
+ *      if (s)
+ *         memcpy(dest, src, n);
+ *
+ * except that it tries to run in the same amount of time whether <b>s</b> is
+ * true or not.
+ **/
+void
+memcpy_if_true_timei(bool s, void *dest, const void *src, size_t n)
+{
```
dgoulet-tor: It the "timei" a typo or?

src/feature/hs/hs_circuit.c
```diff
@@ -958,6 +959,42 @@ hs_circ_handle_intro_established(const hs_service_t *service,
...
+  }
+
+  /* This should not happen since we should have made onionbalance
+   * subcredentials when we created our descriptors. */
+  if (BUG(!service->ob_subcreds)) {
+    return -1;
+  }
+
+  /* We are an onionbalance instance: */
+  data->n_subcredentials = (int) service->n_ob_subcreds;
```
dgoulet-tor: That very cast should not be necessary if we change it to a` size_t` as I asked earlier in my review.

src/lib/ctime/di_ops.c
asn-d6: I think it stands for "time invariant". Do you think the function name should be changed? I thought of changing it to `memcpy_if_true_timeinvariant` but not sure if that would be better. I can also add a comment that clarifies that `timei` stands for time invariant, but that seems a bit silly too. Any ideas?

src/core/crypto/hs_ntor.c
```diff
@@ -452,6 +452,28 @@ hs_ntor_service_get_introduce1_keys(
...
+/**
+ * As hs_ntor_service_get_introduce1_keys(), but take multiple subcredentials
+ * as input, and yield multiple sets of keys as output.
+ **/
+int
+hs_ntor_service_get_introduce1_keys_multi(
+            const struct ed25519_public_key_t *intro_auth_pubkey,
+            const struct curve25519_keypair_t *intro_enc_keypair,
+            const struct curve25519_public_key_t *client_ephemeral_enc_pubkey,
+            int n_subcredentials,
```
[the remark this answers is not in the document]
nmathewson: I'm not so sure; this number will never be greater than 4 or 5, right?

src/lib/ctime/di_ops.c
nmathewson: We already have gt_i64_timei(); I'm wondering if there is some other standard string we should use for this stuff.

src/core/crypto/hs_ntor.c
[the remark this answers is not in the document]
nmathewson: That said, it won't hurt to let it be a size_t, and avoiding the possibility of negative values might make it worthwhile.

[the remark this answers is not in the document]
asn-d6: Please see  901ed35 for this fix.

src/feature/hs/hs_circuit.c
asn-d6: please see 901ed35 for this fix.

src/lib/ctime/di_ops.c
dgoulet-tor: Ah great, didn't know what it meant. I'm fine with it.

src/feature/hs/hs_client.c
```diff
@@ -1837,13 +1837,14 @@ hs_client_decode_descriptor(const char *desc_str,
     uint64_t current_time_period = hs_get_time_period_num(0);
     hs_build_blinded_pubkey(service_identity_pk, NULL, 0, current_time_period,
                             &blinded_pubkey);
-    hs_get_subcredential(service_identity_pk, &blinded_pubkey, subcredential);
+    hs_get_subcredential(service_identity_pk, &blinded_pubkey, &
+                         subcredential);
```
dgoulet-tor: Oh wow I thought this was a bug but also "how can this compile?" and then I noticed the `&` on the previous line.

I would recommend to fix the syntax :).

asn-d6: Addressed in 975102869a.

src/feature/hs/hs_ob.c
```diff
@@ -286,10 +287,9 @@ hs_ob_get_subcredentials(const hs_service_config_t *config,
   tor_assert(config);
   tor_assert(subcredentials);
 
+  /* Our caller made sure that we are an OB instance */
   num_pkeys = smartlist_len(config->ob_master_pubkeys);
-  if (!num_pkeys) {
-    goto end;
-  }
+  tor_assert(num_pkeys > 0);
```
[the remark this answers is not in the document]
asn-d6: Agreed. Addressed in 975102869a.

src/feature/hs/hs_ob.c
```diff
@@ -315,18 +319,30 @@ compute_subcredentials(const hs_service_t *service,
...
                             const ed25519_public_key_t *, pkey) {
       build_subcredential(pkey, tp + steps[i], &subcreds[idx]);
       idx++;
     } SMARTLIST_FOREACH_END(pkey);
   }
 
+  /* And then in the end we add the two subcredentials of the current active
+   * instance descriptors */
+  memcpy(&subcreds[idx++],
+         service->desc_current->desc->subcredential.subcred, SUBCRED_LEN);
```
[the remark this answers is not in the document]
asn-d6: Good point. Addressed in 975102869a.

src/feature/hs/hs_service.h
```diff
@@ -305,8 +305,13 @@ typedef struct hs_service_t {
   /** Next descriptor. */
   hs_service_descriptor_t *desc_next;
 
-  /* XXX: Credential (client auth.) #20700. */
-
+  /* If this is an onionbalance instance, this is an array of subcredentials
+   * that should be used when decrypting an INTRO2 cell. If this is not an
+   * onionbalance instance, this is NULL.
+   * See [ONIONBALANCE] section in rend-spec-v3.txt for more details . */
+  hs_subcredential_t *ob_subcreds;
```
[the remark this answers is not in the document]
asn-d6: It is all the subcredentials, but the variable is named `ob_subcreds` because it's used only when onionbalance is enabled. See comment in review below for more info.

src/feature/hs/hs_circuit.c
```diff
@@ -958,6 +959,42 @@ hs_circ_handle_intro_established(const hs_service_t *service,
...
+    return 0;
+  }
+
+  /* This should not happen since we should have made onionbalance
+   * subcredentials when we created our descriptors. */
+  if (BUG(!service->ob_subcreds)) {
+    return -1;
+  }
+
+  /* We are an onionbalance instance: */
```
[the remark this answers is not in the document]
asn-d6: ob_subreds also includes the descriptor's subcredentials as seen in `compute_subcredentials`.

I agree that the function could be simpler that way, but I'm not sure how to do this while preserving the optimal performance of the non-OB mode. In particular, when the INTRO2 is received in the non-OB case, the service can know the exact subcredential that should be used because each intro point corresponds to a specific descriptor and hence to a specific subcredential. So in the non-OB case, we don't even need to try both descriptor subcredentials, we can only try one.

src/feature/hs/hs_ob.c
```diff
@@ -285,14 +285,17 @@ compute_subcredentials(const hs_service_t *service,
...
+  tor_assert(subcredentials_out);
   /* Our caller has checked these too */
   tor_assert(service->desc_current);
   tor_assert(service->desc_next);
 
   /* Our caller made sure that we are an OB instance */
   num_pkeys = smartlist_len(service->config.ob_master_pubkeys);
-  tor_assert(num_pkeys > 0);
+  if (!num_pkeys) {
+    subcredentials_out = NULL;
```
nmathewson: Should this be *subcredentials_out ?

src/feature/hs/hs_circuit.c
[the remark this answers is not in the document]
nmathewson: ah, ok.  This seems like something we should look at in a ticket down the road then; maybe we could have a list of subcredentials for each intro point?  But it's not so bad to do things this way if there's a reason.

src/feature/hs/hs_ob.c
asn-d6: !Yes indeed. See 93cb807 for a unittest and fix

PR #1687, opened 2020-01-28
Use print() function in both Python 2 and Python 3

cclauss:
A partial solution to https://trac.torproject.org/projects/tor/ticket/32791

Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

__xrange()__ was removed in Python 3 in favour of an iterator-base version of __range()__.


teor2345:
Thanks for this PR.

We track reviews using our Trac ticket system, and we don't always notice GitHub PRs straight away.

If you put a link to the PR on the ticket (or open a new ticket in Core Tor/Tor), and put the ticket in needs_review, then we'll review your PR in the next week or two.

I've put the link for this PR here:
https://trac.torproject.org/projects/tor/ticket/32791#comment:2

PR #1698, opened 2020-01-30
Ticket33072 043 01

dgoulet-tor:


src/core/mainloop/connection.c
```diff
@@ -3277,8 +3277,11 @@ connection_bucket_write_limit(connection_t *conn, time_t now)
  * shouldn't send <b>attempt</b> bytes of low-priority directory stuff
  * out to <b>conn</b>.
  *
- * If we are a directory authority, always answer dir requests thus true is
- * always returned.
+ * If we are are a directory authority, true is returned (indicating that we
```
armadev: are are

src/core/mainloop/connection.c
```diff
@@ -3277,8 +3277,11 @@ connection_bucket_write_limit(connection_t *conn, time_t now)
...
+ * should answer the request) if one of these conditions is met:
```
armadev: Careful -- 'true' indicates that we should not answer the request. This is part of the confusion we have from calling our option RejectUnderLoad and then having this function check if we're too loaded. It's hard to count the right number of negatives. Maybe we'll be happier long-term having the function be called "have-enough-room-to-answer" and inverting everything.

src/core/mainloop/connection.c
dgoulet-tor: Fixed

src/core/mainloop/connection.c
dgoulet-tor: Indeed, s/true/false/g in that sentence. Fixed.

PR #1686, opened 2020-01-27
Travis CI: Use flake8 to find Python syntax errors

cclauss:
Blocked by ~#1687~

`from __future__ import print_function` changes the way that Python 2 operates so that legacy __print__ statements are syntax errors.  This PR demonstrates that this is the case in a dozen files.

PR #1712, opened 2020-02-07
Stop assuming that /usr/bin/python exists

AlwaysLivid:
This proposed change will change all occasions of `/usr/bin/python` to the more agnostic `/usr/bin/env python`, similarly to the change [here](https://github.com/torproject/tor/commit/a10d4adc25500e160c71d0d3f32a6f3ebbf3c9fd).

[Relevant Trac ticket](https://trac.torproject.org/projects/tor/ticket/33192#ticket)

### DIFF
--- a/changes/ticket33192
+++ b/changes/ticket33192
@@ -1,0 +1,5 @@
+  o Minor feature (python):
+    - Stop assuming that /usr/bin/python exists. Instead of using a
+      hardcoded path in scripts that still use Python 2, use /usr/bin/env,
+      similarly to the scripts that use Python 3. Fixes bug 33192; bugfix
+      on 0.4.2.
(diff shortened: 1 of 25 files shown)


AlwaysLivid:
I force-pushed a new commit, after the previous CI check failed, due to me using a `#` before the ticket number in the commit message.

AlwaysLivid:
I tested and looked into more files in order to figure out which ones should use Python 2, which ones should use Python 3 and which ones should be considered as version agnostic.

I may have left out something, but apparently, this seems to be breaking the tests which was something that I couldn't foresee in my local environment, so I'll push a new commit to fix that in a bit.

teor2345:
Thanks!

Let us know when you've got things working. You might find some of the changes in #1686 helpful.

AlwaysLivid:
> Thanks!
> 
> Let us know when you've got things working. You might find some of the changes in #1686 helpful.

The only problem that I can really think of right now is the tree thing/merge conflict.

PR #1720, opened 2020-02-11
Ticket32121

nmathewson:


scripts/git/git-merge-forward.sh
```diff
@@ -170,49 +140,16 @@ done
 # Git worktrees to manage #
 ###########################
 
+set -e
```
teor2345: Should we `set -e` at the top of each script?

teor2345: Hmm, maybe that's another ticket?

PR #1727, opened 2020-02-12
Bug33285

teor2345:


src/feature/dirauth/dirvote.h
```diff
@@ -230,6 +230,21 @@ char *networkstatus_get_detached_signatures(smartlist_t *consensuses);
 STATIC microdesc_t *dirvote_create_microdescriptor(const routerinfo_t *ri,
                                                    int consensus_method);
 
+/* These are hardwired, to avoid disaster. */
```
nmathewson: Now that these are named constants, they should have doxygen -- probably including suitable warnings about changing them without thinking.

teor2345: Done in ed1ee20, with warnings for required protocols.

I also referenced proposal 303, which talks about when we should update required protocols.

PR #1731, opened 2020-02-13
Ticket33280 master

teor2345:


src/test/include.am
```diff
@@ -64,7 +64,7 @@ TESTS += src/test/test src/test/test-slow src/test/test-memwipe \
 
 # run a quick test or two
 # this test only uses IPv4
-TEST_CHUTNEY_FLAVOR_QUICK = bridges+hs-v2
+TEST_CHUTNEY_FLAVOR_QUICK = bridges+hs-v23
```
nmathewson: Would it make sense for chutney itself to define a flavor called "quick-test" or something, so that Tor doesn't need to know which flavor chutney thinks is best?  Or is that a mistake?

Makefile.am
```diff
@@ -293,42 +293,111 @@ test-network: need-chutney-path $(TESTING_TOR_BINARY) src/tools/tor-gencert
...
+
+# Make the test network log directory, if it does not exist
+.PHONY: test-network-mkdir
+test-network-mkdir:
+	@mkdir -p $(TEST_NETWORK_ALL_LOG_DIR)
+
+# Clean the test network log directory
+.PHONY: test-network-clean
+test-network-clean:
+	rm -f $(TEST_NETWORK_ALL_LOG_DIR)/*.log $(TEST_NETWORK_ALL_LOG_DIR)/*.trs
```
nmathewson: I was about to ask whether we needed quotes here, but actually: I'd be more comfortable if we had quotes around these variable expansions, whether they are needed or not.

Makefile.am
```diff
@@ -293,42 +293,111 @@ test-network: need-chutney-path $(TESTING_TOR_BINARY) src/tools/tor-gencert
...
+test-network-ipv6:
+	@$(MAKE) test-network-mkdir
+	@$(MAKE) test-network-clean
+	@$(MAKE) test-network-ipv6-impl ipv6_flavors="$(TEST_CHUTNEY_FLAVORS_IPV6)"
+	@$(MAKE) test-network-results
+
+# Make the test network log directory, if it does not exist
+.PHONY: test-network-mkdir
+test-network-mkdir:
+	@mkdir -p $(TEST_NETWORK_ALL_LOG_DIR)
```
nmathewson: Same note here about quotes

src/test/include.am
teor2345: No, the best flavour actually depends on the tor version: newer tor versions may support extra features, and use a newer flavour.

Makefile.am
teor2345: Fixed comprehensively in 454e0af.

Longer-term, we should consider moving scripts out of the Makefile and CI configs, so that:
* shellcheck can check the scripts, and
* we can re-use common code.

Makefile.am
teor2345: Fixed comprehensively in 454e0af.

teor2345: Now 6cabdda, make warned about some non-portable comments.

PR #1741, opened 2020-02-14
Ticket31634

nmathewson:


src/app/main/subsysmgr.c
```diff
@@ -302,7 +302,8 @@ subsystems_dump_list(void)
 {
   for (unsigned i = 0; i < n_tor_subsystems - 1; ++i) {
     const subsys_fns_t *sys = tor_subsystems[i];
-    printf("% 4d\t%s\n", sys->level, sys->name);
+    printf("% 4d\t%16s\t%s\n", sys->level, sys->name,
+           sys->location?sys->location:NULL);
```
teor2345: NULL works on some systems, because the bytes there are '\0'. But it's not clear from man page that `printf("%s", NULL);` is guaranteed to work.

```suggestion
           sys->location?sys->location:"");
```

doc/tor.1.txt
```diff
@@ -184,7 +184,14 @@ The following options in this section are only recognized on the
     signing key certificate and will output, both in the logs as well
     as to stdout, the signing key certificate's expiration time in
     ISO-8601 format.  For example, the output sent to stdout will be
-    of the form: "signing-cert-expiry: 2017-07-25 08:30:15 UTC"
+    of the form: "signing-cert-expiry: 2017-07-25 08:30:15 UTC"h
```
teor2345: Typo?
```suggestion
    of the form: "signing-cert-expiry: 2017-07-25 08:30:15 UTC"
```

scripts/maint/check_subsys_ordering.py
```diff
@@ -0,0 +1,118 @@
...
+import re
+import sys
+
+# Not every subsystem is initialized in a place corresponding to its
+# name.  These are the locations of the ones that are not.
+LOCATIONS = {
+    "btrack" : "feature/control",
+    "network" : "lib/net",
+    "ocirc_event" : "core/or",
+    "orconn_event" : "core/or",
```
teor2345: I thought you removed orcirc_event and orconn_event in an earlier commit?

scripts/maint/check_subsys_ordering.py
```diff
@@ -0,0 +1,118 @@
...
+    "relay" : "feature/relay",
```
teor2345: Why is feature/relay here, but feature/dirauth is not?

scripts/maint/check_subsys_ordering.py
nmathewson: ugh, I believe so.  Also this entire script should be junked: the new function in includes.py is what we should be using instead.

PR #1742, opened 2020-02-14
Ticket31011: Refuse to publish a bridge with internal addr to bridgeauth

cjb:
If a PT:
  * has ServerTransportListenAddr set to an internal address
  * and it's a bridge
  * and it's publishing its descriptor to the default bridge authorities
then fail the config as invalid, since we don't want internal IP info
being sent to bridgeauths for distribution.

And for clients that aren't running with this fix: if we're a bridgeauth,
reject extra-info descs containing those internal addresses.

Fixes #31011.


src/feature/dirauth/process_descs.c
```diff
@@ -818,6 +818,15 @@ dirserv_add_extrainfo(extrainfo_t *ei, const char **msg)
     goto fail;
   }
 
+  /* If the extrainfo descriptor contains an internal address, reject it. */
+  if (dirserv_router_has_valid_address(ri) < 0) {
+    log_notice(LD_DIR, "Rejecting an extrainfo descriptor '%s' "
+               "with an internal address.", ri->nickname);
```
teor2345: You probably want routerinfo_describe() here, nicknames aren't unique.

src/feature/relay/transport_config.c
```diff
@@ -230,6 +230,21 @@ options_validate_server_transport(const or_options_t *old_options,
     char *bindaddr = get_bindaddr_from_transport_listen_line(cl->value, NULL);
     if (!bindaddr)
       REJECT("ServerTransportListenAddr did not parse. See logs for details.");
+
+    tor_addr_t tor_addr;
+    uint16_t tor_port;
+    int rv = tor_addr_port_parse(LOG_WARN, bindaddr, &tor_addr, &tor_port, 0);
+    if (rv < 0) {
+        log_notice(LD_CONFIG, "Your ServerTransportListenAddr failed to parse "
+                              "as a Tor address; check your configuration.");
```
teor2345: This tor_addr_t parsing just accepts IPv4 and IPv6 addresses:
```suggestion
                              "as an IPv4 or IPv6 address; check your configuration.");
```

src/feature/dirauth/process_descs.c
```diff
@@ -818,6 +818,15 @@ dirserv_add_extrainfo(extrainfo_t *ei, const char **msg)
     goto fail;
   }
 
+  /* If the extrainfo descriptor contains an internal address, reject it. */
+  if (dirserv_router_has_valid_address(ri) < 0) {
```
teor2345: You're checking the routerinfo here, but you should be checking the ServerTransportListenAddr in the extrainfo?

PR #1746, opened 2020-02-17
Small changes to scheduler comments and state changes

stevenengler:
See ticket [#33349](https://trac.torproject.org/projects/tor/ticket/33349).

> I have attached a patch that tries to make some of the comments in scheduler.c easier to follow, such as the comments that reference the old channels_waiting_for_cells and channels_waiting_to_write lists that were removed in ​commit 3530825c5. It also simplifies a couple of the channel state changes.

Merged in: f0561861e3

### DIFF
--- a/src/core/or/scheduler_kist.c
+++ b/src/core/or/scheduler_kist.c
@@ -463,6 +463,13 @@ MOCK_IMPL(void, channel_write_to_kernel, (channel_t *chan))
   log_debug(LD_SCHED, "Writing %lu bytes to kernel for chan %" PRIu64,
             (unsigned long)channel_outbuf_length(chan),
             chan->global_identifier);
+  /* Note that 'connection_handle_write()' may change the scheduler state of
+   * the channel during the scheduling loop with
+   * 'connection_or_flushed_some()' -> 'scheduler_channel_wants_writes()'.
+   * This side-effect will only occur if the channel is currently in the
+   * 'SCHED_CHAN_WAITING_TO_WRITE' or 'SCHED_CHAN_IDLE' states, which KIST
+   * rarely uses, so it should be fine unless KIST begins using these states
+   * in the future. */
   connection_handle_write(TO_CONN(BASE_CHAN_TO_TLS(chan)->conn), 0);
 }
 
--- a/changes/ticket33349
+++ b/changes/ticket33349
@@ -1,0 +1,4 @@
+  o Code simplifications and refactoring:
+    - Updated comments in 'scheduler.c' to reflect old code changes,
+      and simplified the scheduler channel state change code. Closes
+      ticket 33349.
--- a/changes/ticket33349
+++ b/changes/ticket33349
@@ -1,4 +1,4 @@
-  o Code simplifications and refactoring:
+  o Code simplification and refactoring:
     - Updated comments in 'scheduler.c' to reflect old code changes,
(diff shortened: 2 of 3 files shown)


src/core/or/scheduler_kist.c
```diff
@@ -463,6 +463,13 @@ MOCK_IMPL(void, channel_write_to_kernel, (channel_t *chan))
   log_debug(LD_SCHED, "Writing %lu bytes to kernel for chan %" PRIu64,
             (unsigned long)channel_outbuf_length(chan),
             chan->global_identifier);
+  /* Note that 'connection_handle_write()' may change the scheduler state of
+   * the channel during the scheduling loop with
+   * 'connection_or_flushed_some()' -> 'scheduler_channel_wants_writes()'.
+   * This side-effect will only occur if the channel is currently in the
+   * 'SCHED_CHAN_WAITING_TO_WRITE' or 'SCHED_CHAN_IDLE' states, which KIST
+   * rarely uses, so it should be fine unless KIST begins using these states
+   * in the future. */
```
dgoulet-tor: Useful comment! That made me curious to figure out the side-effect here of having a scheduler change after this call. Correct me if I'm wrong (this is quite spaghetti code to follow):

When KIST is within a scheduler loop, the channel can only be in `PENDING` or `WAITING_FOR_CELLS`. Actually, I know for a fact that `scheduler_channel_wants_writes()` is 99% called because the outbuf length is _always_ below the low watermark (those marks are a remain of the vanilla scheduler past).

```
  datalen = connection_get_outbuf_len(TO_CONN(conn));
  if (datalen < OR_CONN_LOWWATER) {
    /* Let the scheduler know */
    scheduler_channel_wants_writes(TLS_CHAN_TO_BASE(conn->chan));
  }
```

However, because the channel is in `PENDING` or `WAITING_FOR_CELLS` state, `scheduler_channel_wants_writes()` just results in doing nothing.

And thus no side effect since the channel will not change state within a scheduler loop.

stevenengler: Yep that sounds right to me. Even when it's under the low water mark, the function will have no effect if the state is `PENDING` or `WAITING_FOR_CELLS`, which are the only states KIST uses unless a connection is closing or errors (since with KIST channels always want to write). I just wanted to document this in case someone modifies KIST in the future (for example following the comments [here](https://github.com/torproject/tor/blob/17724a7cdeb9b98539831f1164a82784f1fb050f/src/core/or/scheduler_kist.c#L679-L681)) and doesn't realize that the state may change from somewhere way outside of the scheduling loop. Or for someone like me who's just looking at the code, isn't sure if KIST handles that situation correctly, and then spends an hour working through that side-case to make sure nothing is going wrong :) It might be useful to add something like

```
old_state = chan->scheduler_state;
connection_handle_write(...);
tor_assert_nonfatal_once(chan->scheduler_state == old_state);
```

but for now I just wanted to document it.

dgoulet-tor: Indeed... I agree that a nonfatal assert() would help us in the future _and_ be a clear way to have "code documentation" about the expected re-entry behavior.

stevenengler: Just let me know if you'd like for me to include that in this PR and remove the comment, but I currently have no way of testing it to make sure that it never asserts (such as if a channel is closing or something).

PR #1747, opened 2020-02-18
Bug4631

teor2345:


src/feature/dirauth/dirvote.c
```diff
@@ -3200,6 +3201,25 @@ dirvote_add_vote(const char *vote_body, const char **msg_out, int *status_out)
     goto err;
   }
 
+  if (!time_posted) { /* I imported this one myself */
+    log_notice(LD_DIR, "Retrieved vote from %s.", vi->address);
```
nmathewson: Should we be logging this in all cases where we don't reject the vote?

src/feature/dirauth/dirvote.c
```diff
@@ -3200,6 +3201,25 @@ dirvote_add_vote(const char *vote_body, const char **msg_out, int *status_out)
...
+   * is to discard it, because using it greatly increases the chances
+   * of a split vote for this round (some dir auths got it in time,
+   * some didn't). */
+  if (time_posted && time_posted > voting_schedule.fetch_missing_votes) {
+    char tbuf1[ISO_TIME_LEN+1], tbuf2[ISO_TIME_LEN+1];
+    format_iso_time(tbuf1, time_posted);
+    format_iso_time(tbuf2, voting_schedule.fetch_missing_votes);
+    log_warn(LD_DIR, "Rejecting vote from %s received at %s; "
+             "our cutoff for received votes is %s", vi->address, tbuf1, tbuf2);
+    *msg_out = "Vote received too late, would be dangerous to count it";
```
nmathewson: Should this be a more actionable warning? It could suggest checking your clock is skewed, and checking whether your CPU is hosed. If neither of those is true, then we have a bug somewhere.

src/feature/dirauth/dirvote.c
teor2345: I don't think this log is correct. The vote is either downloaded from another authority, or loaded from disk. The comment "imported" conflicts with the log message "Retrieved", and neither cover both cases.

And I don't think it's helpful to log a notice, we already log lots of info about votes.

I removed it in 15192f88c0.

src/feature/dirauth/dirvote.c
teor2345: Here are the different possible causes:
* skewed clock
* CPU load
* network load
Late votes can be caused by these issues on the local or remote authority.

I edited the log message in d8c719efd96bf1cbc584fe66c2968b9e18fefe96.

nmathewson:
squashed and merged

PR #1737, opened 2020-02-13
#33275:Tor Manual:Alphabetically sort the last four sections

swati-thacker13:
Alphabetically sorted the following four option categories:

- DENIAL OF SERVICE MITIGATION OPTIONS
- DIRECTORY AUTHORITY SERVER OPTIONS
- HIDDEN SERVICE OPTIONS
- TESTING NETWORK OPTIONS

Note the following:

- Exceptions are marked with the //Out of order comment.
- Non-Persistent Options and Client Authorization Options need no changes.

PR #1755, opened 2020-02-18
Ticket33275 alphabetize remaining tor options sections

tlyu:


tlyu:
This replaces #1737, adding a changes file and a missing comment for an  out-of-order section.

PR #1760, opened 2020-02-20
hs-v3: Log reasons why service can't upload descriptor

dgoulet-tor:
When a service can not upload its descriptor(s), we have no logs on why. This
adds logging for each possible reason for each descriptors.

That logging is emitted every second so it is rate limited for each reason and
per descriptor.

Closes #33400

### DIFF
--- a/src/feature/hs/hs_service.c
+++ b/src/feature/hs/hs_service.c
@@ -707,8 +707,8 @@ get_extend_info_from_intro_point(const hs_service_intro_point_t *ip,
 
 /** Return the number of introduction points that are established for the
  * given descriptor. */
-static unsigned int
-count_desc_circuit_established(const hs_service_descriptor_t *desc)
+MOCK_IMPL(STATIC unsigned int,
+count_desc_circuit_established, (const hs_service_descriptor_t *desc))
 {
   unsigned int count = 0;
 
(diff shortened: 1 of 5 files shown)


src/feature/hs/hs_service.c
```diff
@@ -3042,13 +3042,85 @@ service_desc_hsdirs_changed(const hs_service_t *service,
...
+typedef enum {
+  LOG_DESC_UPLOAD_REASON_MISSING_IPS = 0,
+  LOG_DESC_UPLOAD_REASON_IP_NOT_ESTABLISHED = 1,
+  LOG_DESC_UPLOAD_REASON_NOT_TIME = 2,
+  LOG_DESC_UPLOAD_REASON_NO_LIVE_CONSENSUS = 3,
+  LOG_DESC_UPLOAD_REASON_NO_DIRINFO = 4,
+
+  /* Always +1 of the last value. This is used to allocate the static array of
+   * all rate limiting objects. */
+  LOG_DESC_UPLOAD_REASON_NUM_ = 5,
```
asn-d6: I don't like how this "size of enum" object is also part of the enum. It's not a reason itself, so I don't think it should be in the reason enum for semantic reasons. Perhaps it can be defined separately as a #define using `sizeof(log_desc_upload_reason_t)`?

Also I'm not sure I like the trailing `_`. What does it signify?

Perhaps instead of `NUM` we could use `MAX`? I think that's thje unwritten convention in the rest of the code. I think that would be better since there are not actually 5 reasons; they are 4.

src/feature/hs/hs_service.c
```diff
@@ -3042,13 +3042,85 @@ service_desc_hsdirs_changed(const hs_service_t *service,
...
+
+  tor_assert(service);
+  tor_assert(desc);
+  tor_assert(msg);
+
+  /* Make sure the reason value is valid. It should never happen because we
+   * control that value in the code flow but will be apparent during
+   * development if a reason is added but LOG_DESC_UPLOAD_REASON_NUM_ is not
+   * updated. */
+  if (BUG(reason > (LOG_DESC_UPLOAD_REASON_NUM_ - 1) || reason < 0)) {
```
asn-d6: I'm not sure if I like the `- 1` here. I think it's more standard to do a +1 in the array definition, and then do the check as is. Also if we use sizeof as suggested above we won't fear that the size can change under our feet.

src/feature/hs/hs_service.c
```diff
@@ -3042,13 +3042,85 @@ service_desc_hsdirs_changed(const hs_service_t *service,
...
+  is_next_desc = (service->desc_next == desc);
+
+  /* Current descriptor is the first element in the ratelimit object array.
+   * The next descriptor is the second element. */
+  rlim_pos = (is_next_desc ? 1 : 0);
+  /* Get the ratelimit object for the reason _and_ right descriptor. */
+  rlim = &limits[rlim_pos][reason];
+
+  /* Every minute shouldn't be too annoying for log rate limit. This logging
+   * can be emitted every second for each descriptor. However, one case we
```
asn-d6: Hmm, I'm not sure exactly what this comment means. Can you clarify? I think you mean that for the "it's not time yet" reason we increase its rate limit period because we expect it to trigger all the time?

Also `constantly`.

src/test/test_hs_service.c
```diff
@@ -2169,6 +2190,131 @@ test_export_client_circuit_id(void *arg)
...
+    MOCK(router_have_minimum_dir_info,
+         mock_router_have_minimum_dir_info_false);
+    setup_full_capture_of_logs(LOG_INFO);
+    run_upload_descriptor_event(now);
+    expect_log_msg_containing(
+      "Service [scrubbed] can't upload its current descriptor: "
+      "Not enough directory information");
+    teardown_capture_of_logs();
+    UNMOCK(router_have_minimum_dir_info);
+  }
```
asn-d6: That's a nice test, but I'm wondering if it checks that the ratelimiting works as planned. Does it?

src/feature/hs/hs_service.c
dgoulet-tor: Renamed this and set it as a `#define` instead.

src/feature/hs/hs_service.c
dgoulet-tor: Fixed with the previous fix where it is now `MAX` and the +1 is done when we allocate.

src/feature/hs/hs_service.c
dgoulet-tor: Tried to rewrite that, let me know if it makes more sense.

src/test/test_hs_service.c
dgoulet-tor: So this is a HELLA good point because it is not and not only that but the ratelimiting was not being applied. I've fixed the issue, the rlimt_t needs to be initialized statically, not dynamically which the original code was doing.

I've added the test and fixed the issue.

PR #1763, opened 2020-02-24
Ticket33436

nmathewson:


src/feature/dirauth/voting_schedule.c
```diff
@@ -20,50 +20,6 @@
  * Vote scheduling
```
asn-d6: Perhaps it's now time to update the file-level `\brief` comment of `voting_schedule.c` to indicate that this file should only be used by dirauth now?

src/feature/hs_common/shared_random_client.c
```diff
@@ -31,6 +32,24 @@ srv_to_control_string(const sr_srv_t *srv)
...
 
+/**
+ * If we have no consensus and we are not an authority, assume that this is
+ * the voting interval.  We should never actually use this: only authorities
+ * should be trying to figure out the schedule when they don't have a
+ * consensus.
+ **/
+#define DEFAULT_NETWORK_VOTING_INTERVAL (3600)
+
+/* This is an unpleasing workaround for tests.  Our unit tests assume that we
```
asn-d6: Hmm agreed that this is suboptimal. I wonder if another approach would be to mock `authdir_mode()` in the relevant tests, since this macro is always used next to `authdir_mode()`.

This way it would be ultimately true that only dirauths can get into `voting_schedule.c` in a way that we could also soft-assert against it?

src/feature/hs_common/shared_random_client.c
```diff
@@ -242,8 +238,14 @@ sr_state_get_start_time_of_current_protocol_run(void)
   networkstatus_t *ns = networkstatus_get_live_consensus(approx_time());
```
asn-d6: I think we should update the comment above this to reflect the new tristate reality.

src/feature/hs_common/shared_random_client.c
```diff
@@ -242,8 +238,14 @@ sr_state_get_start_time_of_current_protocol_run(void)
...
   if (ns) {
     beginning_of_curr_round = ns->valid_after;
+  } else if (authdir_mode(get_options()) || ASSUME_AUTHORITY_SCHEDULING) {
+    beginning_of_curr_round = dirauth_sched_get_cur_valid_after_time();
   } else {
-    beginning_of_curr_round = get_start_time_of_current_round();
+    tor_assert_nonfatal_unreached_once();
```
asn-d6: Hm. The soft assert here makes me wonder...

I assume you are soft-asserting here because all the HS-relevant code that calls this function has already checked that the consensus is live before getting n here. That's true. However there is a conceivable TOCTOU race-condition that can trigger this code path; if the consensus was live but about to expire during check, but while moving to this function the second has passed, and now it's not live anymore, so we get to this third state and soft-assert.

Is this something that should concern us? The problem is that if this soft-assert ever triggers we might not be sure if it's the edge case above or a different bug.

src/feature/dirauth/voting_schedule.c
nmathewson: good idea.

src/feature/hs_common/shared_random_client.c
nmathewson: Hm.  Maybe instead of using a live consensus we should use the latest consensus we have whether it's live or not. Would that help here, do you think?

src/feature/hs_common/shared_random_client.c
nmathewson: We could mock dirauth mode or set V3AuthoritativeDirectory here, but the problem is how pervasive in the tests it would need to be.  All of our consensus-generation tests need to reach this code, but most of them don't actually set dirauth mode.

Without this kludge, there are 28 tests that fail and would need to get tweaked.  To me that seemed higher than I thought it was reasonable to deal with, but I could try to tweak them all anyway.  Do you think I should go ahead with that?

nmathewson: (Also, it's not just authority-only functions. For example, consider the test `hs_service/intro2_handling`. It tries to get a subcredential, so it calls into the shared_random_client.c code.  In reality that would never happen without a consensus, but in our test it happens without a consensus, and so would trigger a soft assertion.)

src/feature/hs_common/shared_random_client.c
asn-d6: I think that could work indeed. And then we can have a final case where we return the default value and do the soft-assert?

Also we seem to be doing this quadraple if check twice in the code. Can we abstract it somehow?

src/feature/hs_common/shared_random_client.c
asn-d6: ACK makes sense. Let's leave it as is then.

src/feature/hs_common/shared_random_client.c
nmathewson: I've tried to make the changes for this on 16f6b65f98a1892b2bca5d781632a49f9ec0bd5f .  Doing so made the two functions diverge slightly, so I'm not sure about the abstraction you suggest, but I think this behavior is more sensible.

PR #1764, opened 2020-02-25
Ticket33451

nmathewson:


scripts/git/git-install-tools.sh
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/env bash
+
```
sporksmith: Consider using "strict mode": http://redsymbol.net/articles/unofficial-bash-strict-mode/. I see you use -e below, but consider also -u, -o, and pipefail. (I don't see any actual pipe usage here, but doesn't hurt to add it now for when it sneaks in later)

src/feature/dirclient/dirclient.c
```diff
@@ -1964,6 +1965,44 @@ dir_client_decompress_response_body(char **bodyp, size_t *bodylenp,
...
 }
 
+/**
+ * Total number of bytes downloaded of each directory purpose, when
+ * bootstrapped, and when not bootstrapped.
+ *
+ * (For example, the number of bytes downloaded of purpose p while
+ * not fully bootstrapped is total_dl[p][false].)
+ **/
+static uint64_t total_dl[DIR_PURPOSE_MAX_][2];
```
sporksmith: Consider using signed integers since these are intended as integers and not e.g. bit fields. 

From https://google.github.io/styleguide/cppguide.html#Integer_Types:

> Unsigned integers are good for representing bitfields and modular arithmetic. Because of historical accident, the C++ standard also uses unsigned integers to represent the size of containers - many members of the standards body believe this to be a mistake, but it is effectively impossible to fix at this point. The fact that unsigned arithmetic doesn't model the behavior of a simple integer, but is instead defined by the standard to model modular arithmetic (wrapping around on overflow/underflow), means that a significant class of bugs cannot be diagnosed by the compiler. In other cases, the defined behavior impedes optimization.

nmathewson: This commit isn't actually part of this PR; the thing that updates github from Tor's git repository seems to be stalled.

scripts/git/git-install-tools.sh
teor2345: If we're going to make this change, let's do it for all the git scripts, in a separate PR?

teor2345: I opened https://trac.torproject.org/projects/tor/ticket/33603 for this change in tor, I'll also open tickets for other projects.

sporksmith: SGTM

src/feature/dirclient/dirclient.c
sporksmith: Makes sense; disregard

teor2345:
Squashed and merged to master.

PR #1771, opened 2020-02-26
[Tor Trac #33375]: Stop advertising an IPv6 exit policy when DNS is broken for IPv6

neelchauhan:
For the bug [here](https://trac.torproject.org/projects/tor/ticket/33375).

### DIFF
--- a/changes/bug33375
+++ b/changes/bug33375
@@ -1,0 +1,6 @@
+  o Minor bugfixes (coding best practices checks):
+    - Stop advertising an IPv6 exit policy when DNS is broken for IPv6 by
+      marking the relay descriptor dirty for 24 hours, not advertising an
+      IPv6 exit policy with a dirty descriptor, and then resetting the
+      IPv6 DNS broken count every 24 hours. Fixes bug 33375; bugfix on
+      0.2.9.14. Patch by Neel Chauhan.
--- a/changes/bug33375
+++ b/changes/bug33375
@@ -1,6 +1,4 @@
   o Minor bugfixes (coding best practices checks):
     - Stop advertising an IPv6 exit policy when DNS is broken for IPv6 by
-      marking the relay descriptor dirty for 24 hours, not advertising an
-      IPv6 exit policy with a dirty descriptor, and then resetting the
-      IPv6 DNS broken count every 24 hours. Fixes bug 33375; bugfix on
-      0.2.9.14. Patch by Neel Chauhan.
+      marking the relay descriptor dirty for 24 hours. Fixes bug 33375;
+      bugfix on 0.2.4.7-alpha. Patch by Neel Chauhan.
--- a/changes/bug33375
+++ b/changes/bug33375
@@ -1,4 +1,3 @@
-  o Minor bugfixes (coding best practices checks):
-    - Stop advertising an IPv6 exit policy when DNS is broken for IPv6 by
-      marking the relay descriptor dirty for 24 hours. Fixes bug 33375;
-      bugfix on 0.2.4.7-alpha. Patch by Neel Chauhan.
+  o Major bugfixes (ipv6):
+    - Stop advertising an IPv6 exit policy when DNS is broken for IPv6.
+      Fixes bug 33375; bugfix on 0.2.4.7-alpha. Patch by Neel Chauhan.
--- a/src/feature/relay/dns.c
+++ b/src/feature/relay/dns.c
@@ -83,6 +83,9 @@
  * that the resolver is wedged? */
 #define RESOLVE_MAX_TIMEOUT 300
 
+/** If IPv6 DNS fails, how long should we not advertise an IPv6 exit policy? */
+#define IPV6_FAILED_EXIT_POLICY_TIMEOUT 86400 /* 24 hours */
+
 /** Our evdns_base; this structure handles all our name lookups. */
 static struct evdns_base *the_evdns_base = NULL;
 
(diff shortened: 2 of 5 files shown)


changes/bug33375
```diff
@@ -0,0 +1,6 @@
+  o Minor bugfixes (coding best practices checks):
+    - Stop advertising an IPv6 exit policy when DNS is broken for IPv6 by
+      marking the relay descriptor dirty for 24 hours, not advertising an
+      IPv6 exit policy with a dirty descriptor, and then resetting the
+      IPv6 DNS broken count every 24 hours. Fixes bug 33375; bugfix on
+      0.2.9.14. Patch by Neel Chauhan.
```
teor2345: Can you help me understand why you chose 0.2.9.14 as the bugfix version?

IPv6 DNS support was added early in the 0.2 series, maybe 0.2.4 ?

changes/bug33375
```diff
@@ -0,0 +1,6 @@
+  o Minor bugfixes (coding best practices checks):
+    - Stop advertising an IPv6 exit policy when DNS is broken for IPv6 by
+      marking the relay descriptor dirty for 24 hours, not advertising an
+      IPv6 exit policy with a dirty descriptor, and then resetting the
+      IPv6 DNS broken count every 24 hours. Fixes bug 33375; bugfix on
```
teor2345: Can you rephrase the changes file so it's clearer?

It helps to have shorter sentences. Start with a sentence that summarises the change.

You don't need to describe the change in so much detail. Something like "disable the IPv6 exit policy for 24 hours" should be enough.

src/feature/relay/dns.c
```diff
@@ -127,6 +129,8 @@ static uint64_t n_ipv6_requests_made = 0;
 static uint64_t n_ipv6_timeouts = 0;
 /** Global: Do we think that IPv6 DNS is broken? */
 static int dns_is_broken_for_ipv6 = 0;
+/** Has IPv6 DNS failed ever? */
+static int dns_was_broken_for_ipv6 = 0;
```
teor2345: Are this comment and variable name accurate?

The code resets IPv6 DNS after 24 hours, so it would be helpful to say "recent".

src/feature/relay/dns.c
```diff
@@ -2026,6 +2033,26 @@ dns_launch_correctness_checks(void)
   }
 }
 
+/** If appropriate, if IPv6 DNS failed, reset the countner every 24 hours */
+static void
+dns_launch_ipv6_broken_reset(void)
```
teor2345: "launch" could mean now, or later.

Can we say "schedule" instead?

```suggestion
dns_schedule_ipv6_broken_reset(void)
```

src/feature/relay/dns.c
```diff
@@ -2026,6 +2033,26 @@ dns_launch_correctness_checks(void)
...
+  static struct event *launch_event = NULL;
+  struct timeval timeout;
+
+  if (dns_was_broken_for_ipv6)
+    return;
+
+  if (!launch_event)
+    launch_event = tor_evtimer_new(tor_libevent_get_base(),
+                                   dns_reset_ipv6_broken, NULL);
+  timeout.tv_sec = 24 * 60 * 60;
```
teor2345: Please define this value using a named macro, rather than just doing the calculation directly in the code.

Named constants help people understand the code.

src/feature/relay/router.c
```diff
@@ -1939,6 +1939,12 @@ get_my_declared_family(const or_options_t *options)
   return result;
 }
 
+#ifdef HAVE_MODULE_RELAY
+#define IPV6_DNS_NOT_BROKEN !dns_seems_to_be_broken_for_ipv6()
```
teor2345: There could be a critical bug here:

dns_seems_to_be_broken_for_ipv6() checks dns_is_broken_for_ipv6. But the patch adds dns_was_broken_for_ipv6, so it looks like you meant to use dns_was_broken_for_ipv6.

Good unit tests would have caught this bug. Better variable naming might also have avoided it.

I also don't think dns_was_broken_for_ipv6 will need a macro, because it's only defined in this file.

changes/bug33375
```diff
@@ -0,0 +1,4 @@
+  o Minor bugfixes (coding best practices checks):
+    - Stop advertising an IPv6 exit policy when DNS is broken for IPv6 by
+      marking the relay descriptor dirty for 24 hours. Fixes bug 33375;
```
teor2345: This changes file doesn't describe what the patch actually does.

Maybe delete "for IPv6 by marking the relay descriptor dirty".

src/feature/relay/dns.c
```diff
@@ -1486,7 +1491,11 @@ configure_nameservers(int force)
   if (evdns_base_count_nameservers(the_evdns_base) == 1) {
     SET("max-timeouts:", "1000000");
   } else {
-    SET("max-timeouts:", "10");
+    if (options->TestingTorNetwork) {
```
teor2345: This code makes libevent stop using a DNS server when there is a timeout.

Please don't change this code.

We want libevent to switch servers after a few timeouts, so that we only disable IPv6 DNS if all the configured DNS servers are failing.

src/feature/relay/router.c
teor2345: Please remove the macro for IPV6_DNS_NOT_BROKEN.

It's confusing to hide functions behind a macro. And double negatives are also confusing.

teor2345: Instead, put a function-like macro in the !defined(HAVE_MODULE_RELAY) section of dns.h.

src/feature/relay/dns.c
```diff
@@ -1559,7 +1568,9 @@ evdns_callback(int result, char type, int count, int ttl, void *addresses,
         log_notice(LD_EXIT, "More than half of our IPv6 requests seem to "
```
teor2345: Please increase the minimum number of DNS queries in these lines:
```
    if (n_ipv6_timeouts > 10 &&
        n_ipv6_timeouts > n_ipv6_requests_made / 2) {
```

But set it to 1 when TestingTorNetwork is 1.

Also, please replace the numbers by named constants. One of the reasons you changed the wrong code, is that there are multiple "10"s in this file.

PR #1773, opened 2020-03-01
Prop 310

frochet:
Implements Proposal 310 issued from trac ticket 32088.

This patch also covers Nickm's bucket list from the ticket, i.e.:
- tests or the changed behavior. (Note: existing tests have been adapted to support the new behavior)
- a "changes" file
- editing such that "make check-spaces" passes
- Updates to the documentation and names of all the functions whose behavior has changed.

### DIFF
--- a/src/feature/client/entrynodes.c
+++ b/src/feature/client/entrynodes.c
@@ -99,7 +99,7 @@
  * [x] When we are about to pick a guard, make sure that the primary list is
  * full.
  *
- * [x] Before calling sample_reachable_filtered_entry_guards(), make sure
+ * [x] Before calling first_reachable_filtered_entry_guard(), make sure
  * that the filtered, primary, and confirmed flags are up-to-date.
  *
  * [x] Call entry_guard_consider_retry every time we are about to check
@@ -172,6 +172,7 @@ static entry_guard_t *get_sampled_guard_by_bridge_addr(guard_selection_t *gs,
                                               const tor_addr_port_t *addrport);
 static int entry_guard_obeys_restriction(const entry_guard_t *guard,
                                          const entry_guard_restriction_t *rst);
+static int compare_guards_by_sampled_idx(const void **a_, const void **b_);
 
 /** Return 0 if we should apply guardfraction information found in the
  *  consensus. A specific consensus can be specified with the
@@ -890,6 +891,7 @@ entry_guard_add_to_sample_impl(guard_selection_t *gs,
   tor_free(guard->sampled_by_version);
   guard->sampled_by_version = tor_strdup(VERSION);
   guard->currently_listed = 1;
+  guard->sampled_idx = gs->next_sampled_idx++;
   guard->confirmed_idx = -1;
 
   /* non-persistent fields */
@@ -1707,7 +1709,7 @@ entry_guards_update_filtered_sets(guard_selection_t *gs)
 }
 
 /**
- * Return a random guard from the reachable filtered sample guards
+ * Return the first sampled guard from the reachable filtered sample guards
  * in <b>gs</b>, subject to the exclusion rules listed in <b>flags</b>.
  * Return NULL if no such guard can be found.
  *
@@ -1718,7 +1720,7 @@ entry_guards_update_filtered_sets(guard_selection_t *gs)
  * violate it.
  **/
 STATIC entry_guard_t *
-sample_reachable_filtered_entry_guards(guard_selection_t *gs,
+first_reachable_filtered_entry_guard(guard_selection_t *gs,
                                        const entry_guard_restriction_t *rst,
                                        unsigned flags)
 {
@@ -1771,7 +1773,8 @@ sample_reachable_filtered_entry_guards(guard_selection_t *gs,
            flags, smartlist_len(reachable_filtered_sample));
 
   if (smartlist_len(reachable_filtered_sample)) {
-    result = smartlist_choose(reachable_filtered_sample);
+    smartlist_sort(reachable_filtered_sample, compare_guards_by_sampled_idx);
+    result = smartlist_get(reachable_filtered_sample, 0);
     log_info(LD_GUARD, "  (Selected %s.)",
              result ? entry_guard_describe(result) : "<null>");
   }
@@ -1791,6 +1794,21 @@ compare_guards_by_confirmed_idx(const void **a_, const void **b_)
   if (a->confirmed_idx < b->confirmed_idx)
     return -1;
   else if (a->confirmed_idx > b->confirmed_idx)
+    return 1;
+  else
+    return 0;
+}
+/**
+ * Helper: compare two entry_guard_t by their sampled_idx values.
+ * Used to sort the sampled list
+ */
+static int
+compare_guards_by_sampled_idx(const void **a_, const void **b_)
+{
+  const entry_guard_t *a = *a_, *b = *b_;
+  if (a->sampled_idx < b->sampled_idx)
+    return -1;
+  else if (a->sampled_idx > b->sampled_idx)
     return 1;
   else
     return 0;
@@ -1821,6 +1839,8 @@ entry_guards_update_confirmed(guard_selection_t *gs)
   } SMARTLIST_FOREACH_END(guard);
 
   gs->next_confirmed_idx = smartlist_len(gs->confirmed_entry_guards);
+  // We need the confirmed list to always be give guards in sampled order
+  smartlist_sort(gs->confirmed_entry_guards, compare_guards_by_sampled_idx);
 
   if (any_changed) {
     entry_guards_changed_for_guard_selection(gs);
@@ -1849,6 +1869,7 @@ make_guard_confirmed(guard_selection_t *gs, entry_guard_t *guard)
 
   guard->confirmed_idx = gs->next_confirmed_idx++;
   smartlist_add(gs->confirmed_entry_guards, guard);
+  smartlist_sort(gs->confirmed_entry_guards, compare_guards_by_sampled_idx);
 
   // This confirmed guard might kick something else out of the primary
   // guards.
@@ -1912,7 +1933,7 @@ entry_guards_update_primary(guard_selection_t *gs)
 
   /* Finally, fill out the list with sampled guards. */
   while (smartlist_len(new_primary_guards) < N_PRIMARY_GUARDS) {
-    entry_guard_t *guard = sample_reachable_filtered_entry_guards(gs, NULL,
+    entry_guard_t *guard = first_reachable_filtered_entry_guard(gs, NULL,
                                             SAMPLE_EXCLUDE_CONFIRMED|
                                             SAMPLE_EXCLUDE_PRIMARY|
                                             SAMPLE_NO_UPDATE_PRIMARY);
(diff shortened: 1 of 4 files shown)


src/feature/client/entrynodes.h
```diff
@@ -515,15 +527,15 @@ MOCK_DECL(STATIC circuit_guard_state_t *,
 STATIC entry_guard_t *entry_guard_add_to_sample(guard_selection_t *gs,
                                                 const node_t *node);
 STATIC entry_guard_t *entry_guards_expand_sample(guard_selection_t *gs);
-STATIC char *entry_guard_encode_for_state(entry_guard_t *guard);
+STATIC char *entry_guard_encode_for_state(entry_guard_t *guard, int i);
```
teor2345: Looks like a typo or an old variable name
```suggestion
STATIC char *entry_guard_encode_for_state(entry_guard_t *guard, int dense_sampled_index);
```

src/feature/client/entrynodes.c
```diff
@@ -3042,6 +3073,26 @@ entry_guard_parse_from_state(const char *s)
     }
   }
 
+  if (sampled_idx) {
+    int ok = 1;
+    long idx = tor_parse_long(sampled_idx, 10, 0, INT_MAX, &ok, NULL);
+    if (!ok) {
+      log_warn(LD_GUARD, "Guard has invalid sampled_idx %s",
+          escaped(sampled_idx));
```
nmathewson, line 3081: Can we set the sampled_idx to _something_ in this case?  Right now it seems that we're leaving it unassigned at zero, which is not so great.

src/feature/client/entrynodes.c
```diff
@@ -3042,6 +3073,26 @@ entry_guard_parse_from_state(const char *s)
...
+    }
+  } else if (confirmed_idx) {
+    /* This state has been written by an older Tor version which did not have
+     * sample ordering  */
+
+    guard->sampled_idx = guard->confirmed_idx;
+  } else {
+    log_warn(LD_GUARD, "The state file seems to be into a status that could"
+        " yield to weird entry node selection. It might be better to delete"
+        " it");
```
nmathewson, line 3093: Let's not suggest that people delete their state files.  Instead let's warn, and set the sampled_idx to something.  Leaving sampled_idx at zero seems like a worst-case idea, since it makes the guard get selected before any others.

src/feature/client/entrynodes.c
```diff
@@ -3184,13 +3237,22 @@ entry_guards_load_guards_from_state(or_state_t *state, int set)
       tor_assert(gs);
       smartlist_add(gs->sampled_entry_guards, guard);
       guard->in_selection = gs;
+      /* Recompute the next_sampled_id from the state  */
+      if (gs->next_sampled_idx <= guard->sampled_idx) {
+        gs->next_sampled_idx = ++guard->sampled_idx;
```
nmathewson, line 3242: Wait -- why are you incrementing guard->sampled_idx here?  Don't you just want to say "guard->sampled_idx + 1"?

src/feature/client/entrynodes.c
```diff
@@ -1849,6 +1868,9 @@ make_guard_confirmed(guard_selection_t *gs, entry_guard_t *guard)
 
   guard->confirmed_idx = gs->next_confirmed_idx++;
   smartlist_add(gs->confirmed_entry_guards, guard);
+  /** The confirmation ordering might not be the sample ording. We need to
+   * reorder */
```
nmathewson, line 1872: Let's document this new requirement in the documentation for confirmed_entry_guards.

src/feature/client/entrynodes.c
```diff
@@ -2774,7 +2803,7 @@ entry_guards_update_all(guard_selection_t *gs)
  * <b>guard</b> to the state file.
  */
 STATIC char *
-entry_guard_encode_for_state(entry_guard_t *guard)
+entry_guard_encode_for_state(entry_guard_t *guard, int dense_sampled_idx)
```
nmathewson, line 2806: We should document this new argument.

src/feature/client/entrynodes.c
```diff
@@ -1771,7 +1773,7 @@ sample_reachable_filtered_entry_guards(guard_selection_t *gs,
            flags, smartlist_len(reachable_filtered_sample));
 
   if (smartlist_len(reachable_filtered_sample)) {
-    result = smartlist_choose(reachable_filtered_sample);
+    result = smartlist_get(reachable_filtered_sample, 0);
```
nmathewson, line 1776: Let's have a comment here that refers to proposal 310.

In particular, let's explain why this is safe, and why it won't just return the same guard over and over whether it's working or not.

src/feature/client/entrynodes.c
frochet: Ok, setting missing idx starting from the max sample size to avoid those nodes to be selected before any other.

src/feature/client/entrynodes.c
frochet: idem as above

src/feature/client/entrynodes.c
frochet: Wait -- yes. Now I wonder why a test did not catch it. Looking into it.

src/feature/client/entrynodes.c
frochet: ok, done in struct guard_selection_t

src/feature/client/entrynodes.c
frochet: Ok, done. I've also noticed that functions calling first_reachable_filtered_entry_guard needed documentation updates. Function select_filtered_guard_for_circuit seemed said to look into confirmed guards while excluding them; I believe the documentation was wrong but the code was ok (i.e., we had to exclude confirmed guards).

frochet:
I am not sure to get what's going on regarding the mem leak. It looks like some memory is leaking while parsing the config lines of the state file (the 8 lines of sampled_idx). This should be handled by config_free_lines, right? I don't see why this line is not freed. 

=================================================================
==3923==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 17 byte(s) in 8 object(s) allocated from:
    #0 0x7f99c20b6538 in strdup (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x77538)
    #1 0x559e3df5ea40 in tor_strdup_ src/lib/malloc/malloc.c:165
    #2 0x559e3dc9488b in entry_guard_parse_from_state src/feature/client/entrynodes.c:2981
    #3 0x559e3dc9d2d5 in entry_guards_load_guards_from_state src/feature/client/entrynodes.c:3233
    #4 0x559e3dc9d2d5 in entry_guards_parse_state src/feature/client/entrynodes.c:3456
    #5 0x559e3d8a41fb in test_entry_guard_parse_from_state_full src/test/test_entrynodes.c:697
    #6 0x559e3db7bdc5 in testcase_run_bare_ src/ext/tinytest.c:107

Direct leak of 17 byte(s) in 8 object(s) allocated from:
    #0 0x7f99c20b6538 in strdup (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x77538)
    #1 0x559e3df5ea40 in tor_strdup_ src/lib/malloc/malloc.c:165
    #2 0x559e3dc9488b in entry_guard_parse_from_state src/feature/client/entrynodes.c:2981
    #3 0x559e3dc9d2d5 in entry_guards_load_guards_from_state src/feature/client/entrynodes.c:3233
    #4 0x559e3dc9d2d5 in entry_guards_parse_state src/feature/client/entrynodes.c:3456
    #5 0x559e3d8a47ef in test_entry_guard_parse_from_state_full src/test/test_entrynodes.c:714
    #6 0x559e3db7bdc5 in testcase_run_bare_ src/ext/tinytest.c:107

Direct leak of 17 byte(s) in 8 object(s) allocated from:
    #0 0x7f99c20b6538 in strdup (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x77538)
    #1 0x559e3df5ea40 in tor_strdup_ src/lib/malloc/malloc.c:165
    #2 0x559e3dc9488b in entry_guard_parse_from_state src/feature/client/entrynodes.c:2981
    #3 0x559e3dc9d2d5 in entry_guards_load_guards_from_state src/feature/client/entrynodes.c:3233
    #4 0x559e3dc9d2d5 in entry_guards_parse_state src/feature/client/entrynodes.c:3456
    #5 0x559e3d8a3ff9 in test_entry_guard_parse_from_state_full src/test/test_entrynodes.c:691
    #6 0x559e3db7bdc5 in testcase_run_bare_ src/ext/tinytest.c:107

scripts/maint/practracker/exceptions.txt
```diff
@@ -182,10 +182,10 @@ problem function-size /src/feature/client/addressmap.c:addressmap_rewrite() 109
 problem function-size /src/feature/client/bridges.c:rewrite_node_address_for_bridge() 125
 problem function-size /src/feature/client/circpathbias.c:pathbias_measure_close_rate() 108
 problem function-size /src/feature/client/dnsserv.c:evdns_server_callback() 153
-problem file-size /src/feature/client/entrynodes.c 3827
+problem file-size /src/feature/client/entrynodes.c 4000
```
teor2345, line 185: Notes for the reviewer:

173 unchecked extra lines seems a bit high.
```suggestion
problem file-size /src/feature/client/entrynodes.c 3900
```

src/feature/client/entrynodes.c
asn-d6: Hmm, I'd like to echo this. While prop310 is noted now, it's still not clear to me why the same guard wont be returned over and over. Perhaps some additional documentation or spec patch is in order here to increase comprehension?

src/feature/client/entrynodes.c
asn-d6: Hmm, `compare_guards_by_sampled_idx()` only talks about sampled set but here we do it in the confirmed set. So perhaps we should update documentation?

src/feature/client/entrynodes.c
```diff
@@ -2892,6 +3003,7 @@ entry_guard_parse_from_state(const char *s)
   char *pb_collapsed_circuits = NULL;
   char *pb_unusable_circuits = NULL;
   char *pb_timeouts = NULL;
+  int unvalid_sampled_idx = get_max_sample_size_absolute();
```
asn-d6, line 3006: nitpick: Perhaps `invalid` is more right than `unvalid`.

src/feature/client/entrynodes.h
```diff
@@ -116,6 +116,12 @@ struct entry_guard_t {
    * successfully and decide to keep it?) This field is zero if this is not a
    * confirmed guard. */
   time_t confirmed_on_date; /* 0 if not confirmed */
+  /**
+   * In what order was this guard sampled without replacement? Guards with
```
asn-d6, line 120: What does "without replacement" mean in this context?

src/feature/client/entrynodes.h
```diff
@@ -116,6 +116,12 @@ struct entry_guard_t {
...
+   * lower indices appear earlier on the sampled list
+   */
```
asn-d6, line 122: I think we need a patch for `guard-spec.txt` to explain how this new field should work. Our guard spec is quite engineering oriented and basically explains how these fields tie together, so I think mentioning `sampled_idx` and `next_sampled_idx` there is important.

src/feature/client/entrynodes.c
```diff
@@ -3186,13 +3273,22 @@ entry_guards_load_guards_from_state(or_state_t *state, int set)
       tor_assert(gs);
       smartlist_add(gs->sampled_entry_guards, guard);
       guard->in_selection = gs;
+      /* Recompute the next_sampled_id from the state  */
```
asn-d6, line 3276: Hmm, can we document when this can happen? Also why don't we do it everytime and we only do it in the `<=` case?

src/feature/client/entrynodes.h
frochet: I can send a PR to the torspec repository (I guess guard-spec.txt is there)

src/feature/client/entrynodes.c
asn-d6: Looks good thanks!

src/feature/client/entrynodes.h
asn-d6: ACK!

src/feature/client/entrynodes.h
asn-d6: Yes please!

src/feature/client/entrynodes.c
asn-d6: Thanks for the comment!

PR #1775, opened 2020-03-05
Hs options modernize

nmathewson:


src/feature/hs/hs_config.c
```diff
@@ -444,6 +444,11 @@ config_service_v3(const config_line_t *line_,
   return -1;
 }
 
+/**
+ * DOCDOC
+ **/
```
dgoulet-tor: If you have the chance here to maybe add the comment about this. Seems pretty simple, that this is the section header to configure an onion service.

src/feature/hs/hs_config.c
```diff
@@ -471,25 +475,23 @@ config_generic_service(const config_line_t *line_,
...
 
-  tor_assert(line_);
+  tor_assert(line);
   tor_assert(options);
   tor_assert(service);
 
   /* Makes thing easier. */
   config = &service->config;
 
+  if (BUG(strcasecmp(line->key, "HiddenServiceDir")))
```
dgoulet-tor: `SECTION_HEADER` ?

src/feature/hs/hs_config.c
```diff
@@ -471,25 +475,23 @@ config_generic_service(const config_line_t *line_,
...
+    return -1;
+
   /* The first line starts with HiddenServiceDir so we consider what's next is
    * the configuration of the service. */
-  for (line = line_; line ; line = line->next) {
+  for ( ; line ; line = line->next) {
     int ok = 0;
 
     /* This indicate that we have a new service to configure. */
     if (!strcasecmp(line->key, "HiddenServiceDir")) {
```
dgoulet-tor: Not sure this if is needed anymore. Reason is that we have the `if (BUG())` before so we know that when we enter that function, the start is the `HiddenServiceDir` so we could move the `config->directory_path = tor_strdup(line->value);` below even at the top after the check.

src/feature/hs/hs_config.c
```diff
@@ -471,25 +475,23 @@ config_generic_service(const config_line_t *line_,
...
-      /* This function only configures one service at a time so if we've
-       * already seen one, stop right now. */
-      if (dir_seen) {
-        break;
-      }
       /* Ok, we've seen one and we are about to configure it. */
       dir_seen = 1;
```
dgoulet-tor: Which also makes this `dir_seen` irrelevant.

src/feature/hs/hs_config.c
```diff
@@ -359,126 +346,71 @@ config_validate_service(const hs_service_config_t *config)
...
-        }
-        goto err;
-      }
-      dos_burst_per_sec = true;
-      log_info(LD_REND, "Service INTRO2 DoS defenses burst set to: %" PRIu32,
-               config->intro_dos_burst_per_sec);
-      continue;
+  /* Number of introduction points. */
+  if (check_value_oob(hs_opts->HiddenServiceNumIntroductionPoints,
+                      "HiddenServiceNumIntroductionPoints",
```
dgoulet-tor: As a "good practice" with the new conf interface, should we avoid using directly the option string and instead go for the `.name` instead? As in:

`hs_opts->HiddenServiceNumIntroductionPoints.name;`

src/feature/hs/hs_config.c
```diff
@@ -359,126 +346,71 @@ config_validate_service(const hs_service_config_t *config)
...
+
+  /* Is the DoS defense enabled? */
+  config->has_dos_defense_enabled =
+    hs_opts->HiddenServiceEnableIntroDoSDefense;
+
+  /* Rate for DoS defense */
+  if (check_value_oob(hs_opts->HiddenServiceEnableIntroDoSRatePerSec,
+                      "HiddenServiceEnableIntroDoSRatePerSec",
+                      HS_CONFIG_V3_DOS_DEFENSE_RATE_PER_SEC_MIN,
+                      HS_CONFIG_V3_DOS_DEFENSE_RATE_PER_SEC_MAX)) {
```
dgoulet-tor: No idea if there is a way today but would be neat to have this MIN/MAX functionality built in the `CONF_VAR()` perhaps. If not, no biggy, just an idea but if it exists, great.

src/feature/hs/hs_config.c
```diff
@@ -493,9 +425,6 @@ config_service_v3(const config_line_t *line_,
 
   return 0;
  err:
-  if (dup_opt_seen) {
-    log_warn(LD_CONFIG, "Duplicate directive %s.", dup_opt_seen);
-  }
```
dgoulet-tor: So I suppose that duplicate option are found by the config subsystem? Else, we just removed that entirely, that is fine?

src/feature/hs/hs_config.c
```diff
@@ -516,129 +445,74 @@ static const char SECTION_HEADER[] = "HiddenServiceDir";
...
-      config->max_streams_close_circuit =
-        (unsigned int) helper_parse_uint64(line->key, line->value, 0, 1, &ok);
-      if (!ok || have_max_streams_close) {
-        if (have_max_streams_close)
-          dup_opt_seen = line->key;
-        goto err;
+  /* Directory where the service's keys are stored. */
+  tor_assert(hs_opts->HiddenServiceDir);
+  config->directory_path = tor_strdup(hs_opts->HiddenServiceDir);
+  log_info(LD_CONFIG, "HiddenServiceDir=%s. Configuring...",
```
dgoulet-tor: `SECTION_HEADER` ?

src/feature/hs/hs_config.c
nmathewson: Right. Duplicate options aren't allowed by default.

src/feature/hs/hs_config.c
nmathewson: I think this line is gone in the final version? There are no more "BUG("s in hs_config.c.

src/feature/hs/hs_config.c
nmathewson: I think this line is also gone in the final version?

src/feature/hs/hs_config.c
nmathewson: I don't see dir_seen any longer either.

src/feature/hs/hs_config.c
nmathewson: I would love to have this built in to CONF_VAR, but there isn't a way to do that just yet.  We can refactor once there is.

src/feature/hs/hs_config.c
nmathewson: I don't think that field actually has a "name" value.  You'd have to look up the name at the appropriate index of the hs_opts_t_vars table, which isn't totally trivial.

But there is a simpler approach we can use for now; I've added a new commit to add a macro to reduce the duplication.

src/feature/hs/hs_config.c
nmathewson: I've added https://trac.torproject.org/projects/tor/ticket/33547#ticket for the requested enhancement.

PR #1777, opened 2020-03-06
#32143: Run some CI jobs with ALL_BUGS_ARE_FATAL macro

rl1987:
https://trac.torproject.org/projects/tor/ticket/32143

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ matrix:
     - compiler: clang
       os: osx
       ## Turn off some newer features, turn on clang's -Wtypedef-redefinition
-      ## Also, disable ALL_BUGS_FATAL macro.
+      ## Also, disable ALL_BUGS_ARE_FATAL macro.
       env: C_DIALECT_OPTIONS="-std=gnu99" HARDENING_OPTIONS="--enable-expensive-hardening"
 
     ## We run chutney on Linux, because it's faster than chutney on macOS
--- a/.travis.yml
+++ b/.travis.yml
@@ -119,6 +119,8 @@ matrix:
     #- env: CHUTNEY_MAKE="test-network-ipv6" CHUTNEY="yes" CHUTNEY_ALLOW_FAILURES="2" SKIP_MAKE_CHECK="yes"
     #  os: osx
     - env: ALL_BUGS_ARE_FATAL="yes"
+    - env: ALL_BUGS_ARE_FATAL="yes" SKIP_MAKE_CHECK="yes" CHUTNEY="yes"
+    - env: ALL_BUGS_ARE_FATAL="yes" SKIP_MAKE_CHECK="yes" TEST_STEM="yes"
 
 ## (Linux only) Use a recent Linux image (Ubuntu Bionic)
 dist: bionic
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,9 +93,9 @@ matrix:
     #  compiler: clang
     #  os: osx
 
-    - env: ALL_BUGS_ARE_FATAL="yes"
-    - env: ALL_BUGS_ARE_FATAL="yes" SKIP_MAKE_CHECK="yes" CHUTNEY="yes"
-    - env: ALL_BUGS_ARE_FATAL="yes" SKIP_MAKE_CHECK="yes" TEST_STEM="yes"
+    - env: HARDENING_OPTIONS="--enable-assert-fatalism"
+    - env: HARDENING_OPTIONS="--enable-assert-fatalism" SKIP_MAKE_CHECK="yes" CHUTNEY="yes"
+    - env: HARDENING_OPTIONS="--enable-assert-fatalism" SKIP_MAKE_CHECK="yes" TEST_STEM="yes"
 
   ## Allow the build to report success (with non-required sub-builds
   ## continuing to run) if all required sub-builds have succeeded.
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -241,7 +246,8 @@ script:
   - ./autogen.sh
   - CONFIGURE_FLAGS="$ASCIIDOC_OPTIONS $COVERAGE_OPTIONS $HARDENING_OPTIONS $MODULES_OPTIONS $NSS_OPTIONS $OPENSSL_OPTIONS $RUST_OPTIONS --enable-fatal-warnings --disable-silent-rules"
   - echo "Configure flags are $CONFIGURE_FLAGS CC=\"$CC $C_DIALECT_OPTIONS\""
-  - ./configure $CONFIGURE_FLAGS CC="$CC $C_DIALECT_OPTIONS"
+  - if [[ "$ALL_BUGS_ARE_FATAL" == "" ]]; then ./configure $CONFIGURE_FLAGS CC="$CC $C_DIALECT_OPTIONS"; fi
```
nmathewson: I think it would make more sense to use  CONFIGURE_FLAGS or CFLAGS or something for this option, and have only one single `./configure` line.  If we had a separate "if" condition for every way that we can run ./configure, we would quickly hit a combinatorial explosion.

.travis.yml
```diff
@@ -241,7 +246,8 @@ script:
...
+  - if [[ "$ALL_BUGS_ARE_FATAL" != "" ]]; then ./configure CFLAGS="-DALL_BUGS_ARE_FATAL" $CONFIGURE_FLAGS CC="$CC $C_DIALECT_OPTIONS"; fi
```
nmathewson: Using `CFLAGS` like this seems to replace all the other default cflags that we use -- Maybe we should have it be `-DALL_BUGS_ARE_FATAL -O2 -g` for now?  Or maybe we can fix up CFLAGS handling in our configure script

.travis.yml
rl1987: Fixed in 6b3f4bd28febaefc696b03ba32bcc6adefb1b383.

.travis.yml
rl1987: Addressed in a0741a2a3821171f9aeb9928804df6eef1e4f172.

configure.ac
```diff
@@ -228,6 +228,13 @@ if test "x$enable_expensive_hardening" = "xyes" || test "x$enable_fragile_harden
   AC_DEFINE(DEBUG_SMARTLIST, 1, [Enable smartlist debugging])
 fi
 
+AC_ARG_ENABLE(assert-fatalism,
+   AS_HELP_STRING(--enable-assert-fatalism, [force all soft asserts in Tor codebase (tor_assert_nonfatal(), BUG(), etc.) to act as hard asserts (tor_assert() and equivalents); makes Tor fragile and is not recommended for non-debug builds]))
```
teor2345: I'm a bit confused by the option name here.

"Fatalism" is the passive acceptance of the future, and might suggest that we are ignoring asserts.

Can we say "--enable-all-bugs-are-fatal" instead?

teor2345: We also try to avoid double-negatives, particularly in our logs and documentation:
```suggestion
   AS_HELP_STRING(--enable-assert-fatalism, [force all soft asserts in Tor codebase (tor_assert_nonfatal(), BUG(), etc.) to act as hard asserts (tor_assert() and equivalents); makes Tor fragile; is only recommended for debug builds]))
```

.travis.yml
```diff
@@ -93,6 +93,10 @@ matrix:
     #  compiler: clang
     #  os: osx
 
+    - env: HARDENING_OPTIONS="--enable-assert-fatalism"
+    - env: HARDENING_OPTIONS="--enable-assert-fatalism" SKIP_MAKE_CHECK="yes" CHUTNEY="yes"
+    - env: HARDENING_OPTIONS="--enable-assert-fatalism" SKIP_MAKE_CHECK="yes" TEST_STEM="yes"
```
teor2345: We can't add 3 extra jobs in CI. CI is already too slow for some developers.

I think our best option is to:
* add the new configure option to the global HARDENING_OPTIONS at the start of the file, and make sure all the jobs pass

If we have to use allow_failure, we can add the following jobs, without using too much more time:
* test-stem
* plain Linux gcc, no other options
But allow_failure isn't ideal anyway, because someone needs to manually check for failed jobs.

configure.ac
rl1987: Fixed.

rl1987: Fixed.

.travis.yml
rl1987: Added `--enable-all-bugs-are-fatal` to default value of `HARDENING_OPTIONS`. For some builds I set it back to just `--enable-fragile-hardening`, because we want to keep running unit tests that check for assert failure under certain conditions. Such tests are skipped when `ALL_BUGS_ARE_FATAL` is enabled.

scripts/coccinelle/exceptions.txt
```diff
@@ -22,3 +22,4 @@ src/feature/nodelist/networkstatus.c
 src/test/test_address.c
 src/test/test_hs_cache.c
 src/test/test_hs_descriptor.c
+src/feature/dircache/conscache.c
```
teor2345: Instead of adding the whole file to the exceptions, use #ifndef COCCI to wrap the problematic code.

PR #1781, opened 2020-03-09
#26366: Possible duplicated logic in connection_edge_finished_connecting() and connection_exit_connect()

neelchauhan:
PR: https://trac.torproject.org/projects/tor/ticket/26366

### DIFF
--- a/src/core/or/connection_edge.c
+++ b/src/core/or/connection_edge.c
@@ -908,26 +908,19 @@ export_hs_client_circuit_id(edge_connection_t *edge_conn,
   tor_free(buf);
 }
 
-/** Connected handler for exit connections: start writing pending
- * data, deliver 'CONNECTED' relay cells as appropriate, and check
- * any pending data that may have been received. */
-int
-connection_edge_finished_connecting(edge_connection_t *edge_conn)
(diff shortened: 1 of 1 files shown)


src/core/or/connection_edge.c
```diff
@@ -943,14 +936,49 @@ connection_edge_finished_connecting(edge_connection_t *edge_conn)
     int connected_payload_len =
       connected_cell_format_payload(connected_payload, &conn->addr,
                                     edge_conn->address_ttl);
-    if (connected_payload_len < 0)
+    if (connected_payload_len < 0) {
+      if (zero_payload_cleanup) {
```
tlyu, line 940: I'm not sure I like this new parameter whose only purpose is to run some code that is present in `connection_exit_connect()` but absent in `connection_edge_finished_connecting()`. It seems to make the supposed de-duplication effort more complicated than it has to be. Could `connection_exit_connect()` maybe check for a `-1` return from `connection_exit_set_state_open()` instead? It seems to me that there's no other place that `connection_exit_set_state_open()` can return `-1`. (A return of `-1` from `connection_edge_send_command()` maps to `0`.)

PR #1783, opened 2020-03-10
hs-v3: Improve accessor semantic of client cached object

dgoulet-tor:
Add an inline helper function that indicates if the cached object contains a
decrypted descriptor or not.

The descriptor object is NULL if tor is unable to decrypt it (lacking client
authorization) and some actions need to be done only when we have a decrypted
object.

This improves code semantic.

Fixes #33458

### DIFF
--- a/src/feature/hs/hs_cache.c
+++ b/src/feature/hs/hs_cache.c
@@ -27,6 +27,15 @@
 static int cached_client_descriptor_has_expired(time_t now,
            const hs_cache_client_descriptor_t *cached_desc);
 
+/** Helper function: Return true iff the cache entry has a decrypted
+ * descriptor. */
+static inline bool
+entry_has_decrypted_descriptor(const hs_cache_client_descriptor_t *entry)
+{
+  tor_assert(entry);
+  return (entry->desc != NULL);
+}
+
 /********************** Directory HS cache ******************/
 
 /** Directory descriptor cache. Map indexed by blinded key. */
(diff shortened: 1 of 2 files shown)


src/feature/hs/hs_cache.c
```diff
@@ -740,11 +768,15 @@ cache_clean_v3_as_client(time_t now)
...
     /* We just removed an old descriptor. We need to close all intro circuits
-     * so we don't have leftovers that can be selected while lacking a
-     * descriptor. We leave the rendezvous circuits opened because they could
-     * be in use. */
-    hs_client_close_intro_circuits_from_desc(entry->desc);
+     * if the descriptor is decrypted so we don't have leftovers that can be
+     * selected while lacking a descriptor. Circuits are selected by intro
+     * authentication key thus we need the descriptor. We leave the rendezvous
+     * circuits opened because they could be in use. */
+    if (entry_has_decrypted_descriptor(entry)) {
```
asn-d6: Is it possible to have a unittest that triggers this bug, to make sure it's fixed? (and it doesn't trigger later in the code path)

src/feature/hs/hs_cache.c
```diff
@@ -27,6 +27,15 @@
 static int cached_client_descriptor_has_expired(time_t now,
            const hs_cache_client_descriptor_t *cached_desc);
 
+/** Helper function: Return true iff the cache entry has a decrypted
+ * descriptor. */
```
asn-d6: Please let's have some more function doc here. I know there is already in the struct, but let's be extra paranoid here to ensure that people that see this function learn about this behavior.

src/feature/hs/hs_cache.c
```diff
@@ -659,14 +683,18 @@ cache_store_as_client(hs_cache_client_descriptor_t *client_desc)
...
+    /* Without a decrypted descriptor, we'll always replace the one we have
+     * with the new one we just fetched because we can't inspect the revision
+     * counter in the plaintext data. */
+    if (!entry_has_decrypted_descriptor(client_desc)) {
       remove_v3_desc_as_client(cache_entry);
       cache_client_desc_free(cache_entry);
       goto store;
     }
 
+    /* From this point on, we know that the decrypted descriptor is in the
```
asn-d6: Ye it's kinda weird that now the check of ->desc is not explicit so these kind of comments are helpful.... Anyhow nothing to change here, just mentioning this weirdness.

PR #1790, opened 2020-03-11
Ticket32910 044 03

dgoulet-tor:


src/core/or/trace_probes_circuit.h
```diff
@@ -7,320 +7,16 @@
...
-    ctf_enum(tor_circuit, state, int, old, old_state)
-    ctf_enum(tor_circuit, state, int, new, new_state)
-  )
-)
-
-#endif /* TOR_TRACE_PROBES_CIRCUIT_H */
-
-/* Must be included after the probes declaration. */
-#include <lttng/tracepoint-event.h>
+#include "lttng_circuit.inc"
```
nmathewson, line 18: We should include this by path, like we do for all our other include files: core/or/lttng_circuit.inc

doc/HACKING/Tracing.md
```diff
@@ -4,6 +4,23 @@ This document describes how the event tracing subsystem works in tor so
...
+## WARNING ##
+
+Tracing the tor daemon **always** generates sensitive data if used in
+production (on the public network).
+
+It **is** ethical for researchers to use tracing for their own tor client (for
+example: building paths, timings, or performance).
+
+It is **NOT** ethical to archive, publish or keep data containing other users'
+activity such as relay data or anything that handles users' traffic.
```
nmathewson, line 16: Maybe we should add a note that it's the same as debug logs in this respect.

PR #1791, opened 2020-03-11
hs-v3: Relax severity of a log message when decoding descriptors.

asn-d6:
Make it LOG_PROTOCOL_WARN and also add the expiration timestamp in there to
ease debugging in the future.

### DIFF
--- a/src/feature/hs/hs_descriptor.c
+++ b/src/feature/hs/hs_descriptor.c
@@ -56,6 +56,7 @@
 #define HS_DESCRIPTOR_PRIVATE
 
 #include "core/or/or.h"
+#include "app/config/config.h"
 #include "trunnel/ed25519_cert.h" /* Trunnel interface. */
 #include "feature/hs/hs_descriptor.h"
 #include "core/or/circuitbuild.h"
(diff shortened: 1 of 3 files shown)


src/feature/hs/hs_descriptor.c
```diff
@@ -1283,11 +1284,20 @@ cert_is_valid(tor_cert_t *cert, uint8_t type, const char *log_obj_type)
...
-             tor_cert_describe_signature_status(cert));
+    if (cert->cert_expired) {
+      char expiration_str[ISO_TIME_LEN+1];
+      format_iso_time(expiration_str, cert->valid_until);
+      log_fn(LOG_PROTOCOL_WARN, LD_REND, "Invalid signature for %s: %s (%s)",
+             log_obj_type, tor_cert_describe_signature_status(cert),
+             expiration_str);
+    } else {
+      log_warn(LD_REND, "Invalid signature for %s: %s",
+               log_obj_type, tor_cert_describe_signature_status(cert));
```
dgoulet-tor: Wasn't the idea of the ticket to make this one proto warn? Because anyone can trigger that remotely or can they with your patch?

asn-d6: It's true that people can trigger this remotely, but so is the case for hundreds of Tor's logs. AFAIK only the expiration case of this message is actually getting frequently triggered and that's why I only tried to hide that (and also add some more information for debugging).

I thought that by allowing the other messages get past, we might find a bug or something at some point.

Let me know and I can put everything at log proto warn.

PR #1801, opened 2020-03-17
Ticket33633

teor2345:


src/feature/relay/circuitbuild_relay.h
```diff
@@ -20,6 +20,9 @@ struct created_cell_t;
 struct circuit_t;
 struct or_circuit_t;
 
+#define EXTEND_TO_CLIENT_PROTOCOL_WARNING \
```
nmathewson: Let's not start using this log declaration style: if proliferated, I think it would make our code harder to read. Instead let's add a function or a macro that logs this message.

teor2345: Fixed in 6f124a3.

I had the changes in a fixup commit, but I accidentally squashed it when I was rebasing to deal with practracker conflicts. (See 988d947.)

All other commits are the same as when you reviewed.

PR #1803, opened 2020-03-17
Ticket33624 044 01

dgoulet-tor:


configure.ac
```diff
@@ -861,7 +861,8 @@ if test "x$enable_static_libevent" = "xyes"; then
    if test "x$tor_cv_library_libevent_dir" = "x(system)"; then
      AC_MSG_ERROR("You must specify an explicit --with-libevent-dir=x option when using --enable-static-libevent")
    else
-     TOR_LIBEVENT_LIBS="$TOR_LIBDIR_libevent/libevent.a $STATIC_LIBEVENT_FLAGS"
+     dnl If libevent was _not_ installed, the library is in .libs/.
+     TOR_LIBEVENT_LIBS="$TOR_LIBDIR_libevent/libevent.a $TOR_LIBDIR_libevent/.libs/libevent.a $STATIC_LIBEVENT_FLAGS"
```
nmathewson: Does this work? I think this would make us expect different libraries in the two different locations.  What happens if both libraries are present? Maybe we should check which one we have.

configure.ac
```diff
@@ -958,7 +958,7 @@ AC_ARG_WITH(ssl-dir,
   ])
 
 AC_MSG_NOTICE([Now, we'll look for OpenSSL >= 1.0.1])
-TOR_SEARCH_LIBRARY(openssl, $tryssldir, [-lssl -lcrypto $TOR_LIB_GDI $TOR_LIB_WS32],
+TOR_SEARCH_LIBRARY(openssl, $tryssldir, [-lssl -lcrypto -ldl -lpthread $TOR_LIB_GDI $TOR_LIB_WS32],
```
nmathewson: I think these libraries are posix-only?

PR #1811, opened 2020-03-19
Bug32588 035 tests

nmathewson:


src/test/test_config.c
```diff
@@ -4860,6 +4860,64 @@ test_config_parse_port_config__ports__server_options(void *data)
...
+  or_options_t *opts = options_new();
+  int port;
+  const tor_addr_t *addr;
+
+  // no ports are configured? We get NULL.
+  port = get_first_advertised_port_by_type_af(CONN_TYPE_OR_LISTENER,
+                                              AF_INET);
+  tt_int_op(port, OP_EQ, 0);
+  addr = get_first_advertised_addr_by_type_af(CONN_TYPE_OR_LISTENER,
+                                              AF_INET);
```
teor2345: We're missing an IPv6 case here. I don't know how important that is.

Maybe I'll just fill it in, then merge?

src/test/test_router.c
```diff
@@ -231,11 +237,59 @@ test_router_check_descriptor_bandwidth_changed(void *arg)
...
+
+  // Now set up a dummy listener.
+  MOCK(get_connection_array, mock_get_connection_array);
+  fake_connection_array = smartlist_new();
+  listener = listener_connection_new(CONN_TYPE_OR_LISTENER, AF_INET);
+  TO_CONN(listener)->port = 54321;
+  smartlist_add(fake_connection_array, TO_CONN(listener));
+
+  // We should get a port this time.
+  tt_int_op(54321, OP_EQ, router_get_advertised_or_port_by_af(opts, AF_INET));
```
teor2345: Maybe we should swap the IPv4 and IPv6 auto cases here?
Because the IPv6 auto was the case that was failing.

I might just make this change before I merge.

changes/bug32588
```diff
@@ -0,0 +1,6 @@
+  o Minor bugfixes (relays, brudges, memory safety):
+    - Stop advertising incorrect IPv6 ORPorts in relay and bridge descriptors,
+      when the IPv6 port was configured as "auto". This may be a low-severity
+      memory safety issue, because the published port number may be derived
+      from uninitialised or out-of-bounds memory reads.
```
teor2345: I have to change this changes file to remove "memory safety" before I merge.

PR #1812, opened 2020-03-19
Ticket33643 skip 035

nmathewson:


teor2345:
Seems to be a duplicate of #1813, but targeted to master.

GitHub makes duplicate PRs like this, if you press the PR button too soon, after switching the target branch.

PR #1825, opened 2020-03-20
#33618: Add IPv6 Support to is_local_addr()

europe-asia-america:
#33618

https://trac.torproject.org/projects/tor/ticket/33618#no4

Passes `make test-network-all`, but that doesn't mean that the feature implemented actually works. Either we need to create a unit test to constantly check for it (I'm unsure whether there is one related to that right now) or the code must be reviewed.


src/lib/net/address.c
```diff
@@ -852,6 +852,24 @@ tor_addr_is_valid_ipv4n(uint32_t v4n_addr, int for_listening)
   return v4n_addr != 0;
 }
 
+/* IPv6 equivalent for tor_addr_is_valid() */
+int
+tor_addr_ipv6_is_valid(const tor_addr_t *addr)
```
teor2345: Hi, you don't need a separate tor_addr_is_valid() function for IPv6. tor_addr_is_valid() already has enough IPv6 to do what you need.

You can just say: "tor_addr_is_valid(ipv6_addr, 0)".

teor2345:
> Passes make test-network-all, but that doesn't mean that the feature implemented actually works. Either we need to create a unit test to constantly check for it (I'm unsure whether there is one related to that right now) or the code must be reviewed.

We do reviews and unit tests. Please create unit tests, and make the changes I've asked for in the code.

If you add new commits to the branch, the pull request will automatically update.

src/lib/net/address.h
```diff
@@ -308,6 +308,7 @@ int tor_addr_is_null(const tor_addr_t *addr);
 int tor_addr_is_loopback(const tor_addr_t *addr);
 
 int tor_addr_is_valid(const tor_addr_t *addr, int for_listening);
+int tor_addr_ipv6_is_valid(const tor_addr_t *addr);
```
teor2345: Please also remove the function declaration
```suggestion
```

src/lib/net/address.c
```diff
@@ -818,7 +818,8 @@ tor_addr_is_loopback(const tor_addr_t *addr)
 /* Is addr valid?
  * Checks that addr is non-NULL and not tor_addr_is_null().
  * If for_listening is true, IPv4 addr 0.0.0.0 is allowed.
- * It means "bind to all addresses on the local machine". */
+ * It means "bind to all addresses on the local machine".
+ * Supports checking IPv6 adresses. */
```
teor2345: Thanks for this extra comment, but it's not really needed. All Tor's address functions support IPv4 and IPv6. (Unless we have named or documented them to only support particular kinds of addresses.)

Normally extra comments are ok. But someone else is changing this function in another ticket, and if you both change nearby lines, we will get merge conflicts. Here is the other ticket:
https://trac.torproject.org/projects/tor/ticket/33679#comment:3

teor2345:
> > Either we need to create a unit test to constantly check for it (I'm unsure whether there is one related to that right now) or the code must be reviewed.
> 
> We do reviews and unit tests. Please create unit tests...

How can we help you create unit tests?

Some of our current tests mock (replace) is_local_addr(), but none of them actually test it.

It will be easier to test is_local_addr() if you split it into two different functions:
* is_local_addr_impl(), which:
  * accepts as arguments:
    * a bool that says if we are enforcing distinct subnets
    * the other node's address
    * this node's IPv4 address
    * this node's IPv6 address, and
  * does the checks that is_local_addr() does now
* is_local_addr(), which:
  * accepts as an argument:
    * addr, the other node's address, and
  * passes to is_local_addr_impl():
    * get_options()->EnforceDistinctSubnets
    * addr
    * last_resolved_addr
    * &(router_get_my_routerinfo()->ipv6_addr), if valid

Then you can add tests for is_local_addr_impl() in test_config.c.

Mrigyen:
Two things:

First, I have modified exceptions.txt. nickm wants someone to work on splitting up config.c since it's too big according to the original exceptions.txt and its best practices list.

Second, teor, could you please verify that I've divided is_local_addr() into two functions appropriately? If not, could you then give me some hints as to how to get started with writing the unit test? Because I've been going through the existing ones and they are intimidating.

src/app/config/config.c
```diff
@@ -2994,14 +2997,32 @@ resolve_my_address(int warn_severity, const or_options_t *options,
 MOCK_IMPL(int,
 is_local_addr, (const tor_addr_t *addr))
 {
-  if (tor_addr_is_internal(addr, 0))
+  int EnforceDistinctSubnets = get_options()->EnforceDistinctSubnets;
+
+  if (tor_addr_family(addr) == AF_INET6) {
```
teor2345: is_local_addr() should be a simple wrapper for is_local_addr_impl().

It should contain a single call to is_local_addr_impl(), like this:
`is_local_addr_impl(EnforceDistinctSubnets, addr, last_resolved_addr, valid_ipv6_addr);`

valid_ipv6_addr should be:
* `&(router_get_my_routerinfo()->ipv6_addr)` if it is valid
* NULL if it is not valid

You should not need to do any checks on other_addr.

src/app/config/config.c
```diff
@@ -2994,14 +2997,32 @@ resolve_my_address(int warn_severity, const or_options_t *options,
...
+  /* Check for an internal IPv4 or IPv6 address */
+  if (tor_addr_is_internal(other_addr, 0))
     return 1;
+
   /* Check whether ip is on the same /24 as we are. */
-  if (get_options()->EnforceDistinctSubnets == 0)
+  if (EnforceDistinctSubnets == 0)
     return 0;
-  if (tor_addr_family(addr) == AF_INET) {
-    uint32_t ip = tor_addr_to_ipv4h(addr);
```
teor2345: This code is correct, please don't delete it.

src/app/config/config.c
```diff
@@ -3010,9 +3031,16 @@ is_local_addr, (const tor_addr_t *addr))
      * it was on net 0, which is already done by tor_addr_is_internal.
      */
     if ((last_resolved_addr & (uint32_t)0xffffff00ul)
-        == (ip & (uint32_t)0xffffff00ul))
+        == (addr_ipv4 & (uint32_t)0xffffff00ul))
```
teor2345: This code is correct, please don't change it.

Instead, replace last_resolved_addr with addr_ipv4

src/app/config/config.c
```diff
@@ -3010,9 +3031,16 @@ is_local_addr, (const tor_addr_t *addr))
...
       return 1;
   }
+
+  /* Check for the same IPv6 /48 as the directory server */
+  if (tor_addr_family(other_addr) == AF_INET6) {
+    if (router_get_my_routerinfo() && tor_addr_is_valid(addr_ipv6, 0))
```
teor2345: If is_local_addr() has already checked router_get_my_routerinfo(), you don't need to check it again here.

teor2345:
> First, I have modified exceptions.txt. nickm wants someone to work on splitting up config.c since it's too big according to the original exceptions.txt and its best practices list. Perhaps that should be a new ticket?

Yes, that's a big job, and it should be a new ticket.

> Second, teor, could you please verify that I've divided is_local_addr() into two functions appropriately?

I've asked you to make some changes. is_local_addr() should be a small wrapper function, which passes the specific arguments to is_local_addr_impl(). That way, you can test is_local_addr_impl() by passing different arguments.

> Could you then give me some hints as to how to get started with writing the unit test? Because I've been going through the existing ones and they are intimidating.

Here's an example of a similar test for tor_addr_is_internal():

It uses these two macros:
https://github.com/torproject/tor/blob/384a771fccb0b0ab8b95122815b33ef26bf042e8/src/test/test_addr.c#L98

And these tests cases:
https://github.com/torproject/tor/blob/384a771fccb0b0ab8b95122815b33ef26bf042e8/src/test/test_addr.c#L345

There's a bit of work setting up the test wrapper functions. But you can focus on writing test cases for now. We want to test every combination:
* EnforceDistinctSubnets on and off
* IPv4 local and not local
* IPv6 local and not local
* IPv6 local: valid and invalid IPv6

Ideally, we also have tests for either side of any boundary condition.

For example, we should test IPv4 addresses in the same /23, /24, and /25.
And IPv6 addresses in the same /47, /48, and /49

teor2345:
I think things got a little confused here, because of all the changes.

I've tried to help by fixing some of the bugs in the new code. I made a new PR, and described the changes in: https://github.com/torproject/tor/pull/1849#issuecomment-606973738. Can you try writing unit tests based on #1849?

(I also fixed some bugs in our old code, that were making these changes harder.)

PR #1831, opened 2020-03-20
ticket#33679

MrSquanchee:


src/test/test_addr.c
```diff
@@ -1671,8 +1671,53 @@ test_addr_octal(void *arg)
...
+                                      &test_addr->addr.in6_addr);
+  TEST_ADDR_VALIDITY(test_addr, 0, 0);
+  TEST_ADDR_VALIDITY(test_addr, 1, 1);
+  done:
+  tor_free(test_addr);
+}
+
 #ifndef COCCI
-#define ADDR_LEGACY(name)                                               \
+#define ADDR_LEGACY(name) \
```
teor2345: Please don't change whitespace, unless you've actually edited the line.

It makes diffs hard to read.

src/test/test_addr.c
```diff
@@ -1671,8 +1671,53 @@ test_addr_octal(void *arg)
...
+  TEST_ADDR_VALIDITY(test_addr, 0, 1);
+  TEST_ADDR_VALIDITY(test_addr, 1, 1);
+
+  tor_free(test_addr);
+
+  /* Tests for IPv6 addresses. */
+  test_addr = tor_malloc(sizeof(tor_addr_t));
+  test_addr->family = AF_INET6;
+
+  /* Obtain 2000::1a00::1000:fc098 . */
```
teor2345: This test is hard to read, because the order is:
IPv4 null
IPv4 not null
IPv6 not null
IPv6 null

Please use a consistent order.

src/test/test_addr.c
```diff
@@ -1671,8 +1671,53 @@ test_addr_octal(void *arg)
...
 
+#define TEST_ADDR_VALIDITY(a, lis, rv) STMT_BEGIN \
+    tt_int_op(tor_addr_is_valid(a, lis), OP_EQ, rv); \
+  STMT_END;
+
+static void
+test_addr_is_valid(void *arg)
+{
+  (void)arg;
+  tor_addr_t *test_addr = tor_malloc(sizeof(tor_addr_t));
```
teor2345: You're missing tests for Unix addresses, and the unspecified null address.

You can use tor_addr_make_null(AF_UNSPEC) to make an unspecified null address.

src/lib/net/address.h
```diff
@@ -308,7 +308,11 @@ int tor_addr_is_null(const tor_addr_t *addr);
 int tor_addr_is_loopback(const tor_addr_t *addr);
 
 int tor_addr_is_valid(const tor_addr_t *addr, int for_listening);
-int tor_addr_is_valid_ipv4n(uint32_t v4n_addr, int for_listening);
+
+/* If for_listening is true allow v4n_addr = 0 too.
+ * else allow only non-zero v4n_addr. */
+#define tor_addr_is_valid_ipv4n(v4n_addr,for_listening) \
+        (for_listening == 1 || v4n_addr != 0)
```
teor2345: Is this really simpler than the original function?

I find it harder to read multi-part expressions on a single line, than the original code.

src/lib/net/address.c
```diff
@@ -859,13 +842,9 @@ tor_addr_is_valid_ipv4n(uint32_t v4n_addr, int for_listening)
...
-  /* Any port value is valid with for_listening. */
-  if (for_listening) {
-    return 1;
-  }
-
-  /* Otherwise, zero ports are invalid. */
-  return port != 0;
+/* If for_listening is 1 any port is valid.
+ * Otherwise, zero ports are invalid. */
+  return for_listening || port != 0;
```
teor2345: Is this really simpler than the original function?

I find it harder to read multi-part expressions on a single line, than the original code.

src/lib/net/address.c
```diff
@@ -827,29 +828,11 @@ tor_addr_is_valid(const tor_addr_t *addr, int for_listening)
...
-    return 1;
-  }
-
-  /* Otherwise, zero addresses are invalid. */
-  return v4n_addr != 0;
+  /* For the case of IPv4 it doesn't matter if it is in network order
+   * or host order because we are testing it against 0 for validity.
+   * If for_listening is true i.e. 1 then it doesn't matter if addr
+   * is null or not i.e. 0.0.0.0 for IPv4 and :: for IPv6. */
+  return for_listening || !tor_addr_is_null(addr);
```
teor2345: This code gives an incorrect result for Unix socket addresses, and the unspecified null address.

IPv4 0.0.0.0 and IPv6 :: are the only documented exceptions. So you can't remove the AF_INET and AF_INET6 checks.

src/lib/net/address.c
```diff
@@ -773,8 +773,12 @@ tor_addr_is_null(const tor_addr_t *addr)
...
   switch (tor_addr_family(addr)) {
     case AF_INET6: {
-      uint32_t *a32 = tor_addr_to_in6_addr32(addr);
-      return (a32[0] == 0) && (a32[1] == 0) && (a32[2] == 0) && (a32[3] == 0);
+      const char *ipv6 = (const char *)tor_addr_to_in6_addr8(addr);
+      int result = 1, i;
+      for (i = 0; i < 16; ++i) {
+        result &= ipv6[i] == 0;
+      }
+      return result;
```
teor2345: I think this code does exactly the same thing as the previous code.

If you think it does something different, please provide a test case that produces a different result on the old and new code.

If you think it is the same, please don't modify this code.

In general, we try to change one function per commit or pull request. That way, we know why the tests are failing.

src/test/test_addr.c
```diff
@@ -1750,6 +1750,49 @@ test_addr_is_valid(void *arg)
   ;
 }
 
+#define TEST_ADDR_IS_NULL(a, rv) STMT_BEGIN               \
```
teor2345: Thanks for these new unit tests.

Let's keep them, even if we don't change tor_addr_is_null().

src/lib/net/address.c
teor2345: We also try to minimise the changes, so they are easy to review,

src/test/test_addr.c
```diff
@@ -1671,6 +1671,128 @@ test_addr_octal(void *arg)
   ;
 }
 
+#define Get_IPv4(test_addr, str) STMT_BEGIN                     \
```
teor2345: We name function-like macros using snake_case:
```suggestion
#define get_ipv4(test_addr, str) STMT_BEGIN                     \
```

src/test/test_addr.c
```diff
@@ -1671,6 +1671,128 @@ test_addr_octal(void *arg)
...
+    test_addr = tor_malloc_zero(sizeof(tor_addr_t));            \
+    test_addr->family = AF_UNIX;                                \
+  STMT_END;
+
+#define Get_AF_UNSPEC(test_addr) STMT_BEGIN                     \
+    test_addr = tor_malloc_zero(sizeof(tor_addr_t));            \
+    test_addr->family = AF_UNSPEC;                              \
+  STMT_END;
+
+#define TEST_ADDR_VALIDITY(a, lis, rv) STMT_BEGIN               \
```
teor2345: All capitals is also ok for macros.

src/test/test_addr.c
```diff
@@ -1671,6 +1671,128 @@ test_addr_octal(void *arg)
   ;
 }
 
+#define Get_IPv4(test_addr, str) STMT_BEGIN                     \
+    test_addr = tor_malloc(sizeof(tor_addr_t));                 \
+    test_addr->family = AF_INET;                                \
+    tor_inet_aton(str, &test_addr->addr.in_addr);               \
```
teor2345: Please check the return value of tor_inet_aton().

src/test/test_addr.c
```diff
@@ -1671,6 +1671,128 @@ test_addr_octal(void *arg)
...
+  STMT_END;
+
+#define Get_IPv6(test_addr, str) STMT_BEGIN                     \
+    test_addr = tor_malloc(sizeof(tor_addr_t));                 \
+    test_addr->family = AF_INET6;                               \
+    tor_inet_pton(AF_INET6, str, &test_addr->addr.in6_addr);    \
```
teor2345: Please check the return value of tor_inet_pton().

src/test/test_addr.c
```diff
@@ -1671,6 +1671,128 @@ test_addr_octal(void *arg)
...
+  /* Tests for IPv6 addresses. */
+
+  /* Test for null IPv6 address. */
+  Get_IPv6(test_addr, "::");
+  TEST_ADDR_VALIDITY(test_addr, 0, 0);
+  TEST_ADDR_VALIDITY(test_addr, 1, 1);
+  tor_free(test_addr);
+
+  /* Test for non-null IPv6 address. */
+  Get_IPv6(test_addr, "2000::1a00::1000:fc098");
```
teor2345: This is not an IPv6 address, it has two "::".

Looks like Windows and macOS/Linux handle this failure differently.
That shouldn't happen, I've opened a ticket to fix it:
https://trac.torproject.org/projects/tor/ticket/33768

PR #1843, opened 2020-03-30
Bug33545

asn-d6:


src/feature/control/control_hs.c
```diff
@@ -50,11 +50,18 @@ parse_private_key_from_control_port(const char *client_privkey_str,
...
   if (base64_decode((char*)privkey->secret_key, sizeof(privkey->secret_key),
                     key_blob,
-                   strlen(key_blob)) != sizeof(privkey->secret_key)) {
+                    strlen(key_blob)) != sizeof(privkey->secret_key)) {
     control_printf_endreply(conn, 512, "Failed to decode x25519 private key");
     goto err;
   }
 
+  if (fast_mem_is_zero((const char*)&privkey->secret_key,
+                       sizeof(privkey->secret_key))) {
```
dgoulet-tor, line 59: I believe it should be `const char*) privkey->secret_key` here without the reference `&`.

Since your unit test is working, maybe in C the result in the same where foo[] directly used or with a reference but unclear? Anyway, we should just be consistent here.

asn-d6: Killer find!

Addressed this in d37c532.

I also made sure that C will return the same pointer regardless of whether we use `&` or not there. Seems to be some property of static arrays but I didn't manage to find it in the standard. But that's the reason the unittest was succeding before.

PR #1847, opened 2020-03-31
configure: Fix enabled module variable expansion

dgoulet-tor:
Fixes #33646

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2665,7 +2665,7 @@ PPRINT_SUBTITLE([Modules])
 
 m4_foreach_w([mname], MODULES,
   [
-    test "x$enable_module_[]mname[]" != "xno" && value=1 || value=0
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -2665,7 +2665,7 @@ PPRINT_SUBTITLE([Modules])
 
 m4_foreach_w([mname], MODULES,
   [
-    test "xenable_module_mname" != "xno" && value=1 || value=0
+    test "x$enable_module_[]mname[]" != "xno" && value=1 || value=0
```
teor2345, line 2668: That still doesn't work, all modules should be "no" when relay is disabled:
https://travis-ci.org/github/torproject/tor/jobs/669204177#L2324

Looks like you're using the input variable, rather than the module variable?

Maybe try something like this:
```suggestion
    MODULE=m4_toupper([]mname[])
    AM_COND_IF(BUILD_MODULE_[]MODULE[], value=1, value=0)
```

teor2345:
Squashed and merged to 0.4.3 and later.

PR #1853, opened 2020-04-04
ticket 33812

Suraj-Upadhyay:
Also refactored some code in order to be able
to test these functions private to the module
rephist.c

### DIFF
--- a/changes/ticket33812
+++ b/changes/ticket33812
@@ -1,7 +1,7 @@
   o Code simplification and refactoring:
     - Move the bandwidth structure and function declarations
       to new file. Include this  module in other files.
-      Remove REPHIST_PRIVATE.
+      Remove REPHIST_PRIVATE. Closes ticket 33812.
   o Testing:
     - Add unit tests for bandwidth statistics manipulation functions.
       Closes ticket 33812. Patch by MrSquanchee
--- a/changes/ticket33812
+++ b/changes/ticket33812
@@ -1,0 +1,7 @@
+  o Code simplification and refactoring:
+    - Move the bandwidth structure and function declarations
+      to new file. Include this  module in other files.
+      Remove REPHIST_PRIVATE.
+  o Testing:
+    - Add unit tests for bandwidth statistics manipulation functions.
+      Closes ticket 33812. Patch by MrSquanchee
--- a/src/feature/stats/bw_array.h
+++ b/src/feature/stats/bw_array.h
@@ -12,6 +12,7 @@
 #include "feature/stats/bw_array_st.h"
 
 STATIC void commit_max(bw_array_t *b);
+STATIC void add_obs(bw_array_t *b, time_t when, uint64_t n);
 STATIC void advance_obs(bw_array_t *b);
 STATIC bw_array_t *bw_array_new(void);
 STATIC void bw_array_free_(bw_array_t *b);
@@ -20,28 +21,6 @@ STATIC void bw_arrays_init(void);
 STATIC void bw_arrays_free_all(void);
 STATIC size_t
 rep_hist_fill_bandwidth_history(char *buf, size_t len, const bw_array_t *b);
-
-/** Add <b>n</b> bytes to the number of bytes in <b>b</b> for second
- * <b>when</b>. */
-STATIC inline void
-add_obs(bw_array_t *b, time_t when, uint64_t n)
-{
-  if (when < b->cur_obs_time)
-    return; /* Don't record data in the past. */
-
-  /* If we're currently adding observations for an earlier second than
-   * 'when', advance b->cur_obs_time and b->cur_obs_idx by an
-   * appropriate number of seconds, and do all the other housekeeping. */
-  while (when > b->cur_obs_time) {
-    /* Doing this one second at a time is potentially inefficient, if we start
-       with a state file that is very old.  Fortunately, it doesn't seem to
-       show up in profiles, so we can just ignore it for now. */
-    advance_obs(b);
-  }
-
-  b->obs[b->cur_obs_idx] += n;
-  b->total_in_period += n;
-}
 
 #define bw_array_free(val) \
   FREE_AND_NULL(bw_array_t, bw_array_free_, (val))
--- a/src/feature/stats/bw_array.h
+++ b/src/feature/stats/bw_array.h
@@ -1,0 +1,49 @@
+/* Copyright (c) 2007-2020, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
+ * @file bw_array_st.h
+ * @brief Bandwidth array manipulation functions.
+ **/
+
+#ifndef BW_ARRAY_H
+#define BW_ARRAY_H
+
+#include "feature/stats/bw_array_st.h"
+
+STATIC void commit_max(bw_array_t *b);
+STATIC void advance_obs(bw_array_t *b);
+STATIC bw_array_t *bw_array_new(void);
+STATIC void bw_array_free_(bw_array_t *b);
+STATIC uint64_t find_largest_max(bw_array_t *b);
+STATIC void bw_arrays_init(void);
+STATIC void bw_arrays_free_all(void);
+STATIC size_t
+rep_hist_fill_bandwidth_history(char *buf, size_t len, const bw_array_t *b);
+
+/** Add <b>n</b> bytes to the number of bytes in <b>b</b> for second
+ * <b>when</b>. */
+STATIC inline void
+add_obs(bw_array_t *b, time_t when, uint64_t n)
+{
+  if (when < b->cur_obs_time)
+    return; /* Don't record data in the past. */
+
+  /* If we're currently adding observations for an earlier second than
+   * 'when', advance b->cur_obs_time and b->cur_obs_idx by an
+   * appropriate number of seconds, and do all the other housekeeping. */
+  while (when > b->cur_obs_time) {
+    /* Doing this one second at a time is potentially inefficient, if we start
+       with a state file that is very old.  Fortunately, it doesn't seem to
+       show up in profiles, so we can just ignore it for now. */
+    advance_obs(b);
+  }
+
+  b->obs[b->cur_obs_idx] += n;
+  b->total_in_period += n;
+}
+
+#define bw_array_free(val) \
+  FREE_AND_NULL(bw_array_t, bw_array_free_, (val))
+
+#endif /* !defined(BW_ARRAY_H). */
--- a/src/feature/stats/bw_array.h
+++ b/src/feature/stats/bw_array.h
@@ -2,14 +2,16 @@
 /* See LICENSE for licensing information */
 
 /**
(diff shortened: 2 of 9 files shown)


src/test/test_stats.c
```diff
@@ -245,6 +246,43 @@ test_rephist_mtbf(void *arg)
   tor_free(ddir_fname);
 }
 
+static void
+test_commit_max(void *arg)
```
MrSquanchee, line 250: I don't know if this a good way to test this function because
I am manipulating the data structure directly which is implementation dependent.
Is this ok ???

src/feature/stats/rephist.h
```diff
@@ -99,10 +99,60 @@ extern struct bw_array_t *write_array;
 #endif
 
 #ifdef REPHIST_PRIVATE
-typedef struct bw_array_t bw_array_t;
+
```
MrSquanchee, line 102: I had to do this because bw_array_t() was private to the module
but still not accessible after #define  REPHIST_PRIVATE.
Is this ok ??

teor2345: Please put the struct in a separate file called "bw_array_st.h".
Then you can include the struct in the test file, and the code file.

That's what we do with other structs.
For examples, see:
https://github.com/torproject/tor/tree/master/src/core/or

src/test/test_stats.c
teor2345: Yes, that's fine.

src/feature/stats/rephist.c
```diff
@@ -1122,15 +1082,22 @@ static bw_array_t *dir_read_array = NULL;
     directory protocol. */
 static bw_array_t *dir_write_array = NULL;
 
-/** Set up [dir_]read_array and [dir_]write_array, freeing them if they
- * already exist. */
-static void
-bw_arrays_init(void)
+/* Free up [dir_]read_array and [dir_]write_array. */
+STATIC void
+bw_arrays_destroy(void)
```
teor2345, line 1087: We call our free functions "*_free" or "*_free_all".

src/test/test_stats.c
```diff
@@ -245,6 +246,117 @@ test_rephist_mtbf(void *arg)
...
+  now += 53;
+  bw = 97;
+  test_add_obs_(b, now, now, bw, tot);
+
+  now += 60*60;
+  bw = 90;
+  test_add_obs_(b, now, now, bw, tot);
+
+  /* We should not test for new period because the time complexity
+   * will be very high. */
```
teor2345, line 351: Can you explain what you mean by "the time complexity will be very high" ?

MrSquanchee: The loop [here](https://github.com/torproject/tor/blob/686494f0f71b9235399b8241aba3e0c2fcb03ea1/src/feature/stats/rephist.c#L1074) will iterate
86400 times for a day.
And you can see the comment below 
describing this behavior.
Should I test it for a day forward, nevertheless ??

src/feature/stats/rephist.c
MrSquanchee: I will correct every typo in my last commit.

src/test/test_stats.c
teor2345: How slow is the test when you run it?

80,000 is not much, for a computer that runs a billion operations per second.

src/feature/stats/bw_array_st.h
```diff
@@ -0,0 +1,68 @@
+
```
ahf, line 1: I think in Tor the general purpose of files ending `_st.h` is to contain *only* the `struct` and no methods related to the struct.

src/feature/stats/rephist.c
```diff
@@ -973,47 +971,45 @@ rep_hist_load_mtbf_data(time_t now)
...
-/** How large are the intervals for which we track and report bandwidth use? */
-#define NUM_SECS_BW_SUM_INTERVAL (24*60*60)
-/** How far in the past do we remember and publish bandwidth use? */
-#define NUM_SECS_BW_SUM_IS_VALID (5*24*60*60)
-/** How many bandwidth usage intervals do we remember? (derived) */
-#define NUM_TOTALS (NUM_SECS_BW_SUM_IS_VALID/NUM_SECS_BW_SUM_INTERVAL)
+/** Recent history of bandwidth observations for read operations. */
+static bw_array_t *read_array = NULL;
+/** Recent history of bandwidth observations for write operations. */
+STATIC bw_array_t *write_array = NULL;
```
ahf, line 977: Nit: I don't think this one have to be `STATIC`.

MrSquanchee: It was STATIC from before. Should I change it ??

teor2345: If it's used in the unit tests, say "STATIC", which means "only available in this file in the tor binary, but available to the unit tests".

If it's not used in the unit tests, say "static", which means "only available in this file" (and not available to the unit tests).

MrSquanchee: The variable here is also used in unit tests.  So leaving it as it is.

src/feature/stats/bw_array.h
```diff
@@ -0,0 +1,28 @@
+/* Copyright (c) 2007-2020, The Tor Project, Inc. */
+/* See LICENSE for licensing information */
+
+/**
+ * @file bw_array_st.h
```
ahf, line 5: Typo in the filename here.

src/feature/stats/bw_array.h
```diff
@@ -0,0 +1,28 @@
...
+ * @brief Bandwidth array manipulation functions.
+ **/
+
+#ifndef BW_ARRAY_H
+#define BW_ARRAY_H
+
+#include "feature/stats/bw_array_st.h"
+
+STATIC void commit_max(bw_array_t *b);
```
ahf, line 14: The `STATIC` methods should be protected by an `BW_ARRAY_PRIVATE` guard, such that people don't accidentally use the static functions in the wrong places. Have a look at `src/lib/geoip/geoip.h` for an example of how it is done elsewhere.

changes/ticket33812
```diff
@@ -0,0 +1,7 @@
+  o Code simplification and refactoring:
+    - Move the bandwidth structure and function declarations
+      to new file. Include this  module in other files.
+      Remove REPHIST_PRIVATE.
```
teor2345, line 4: Each changes file paragraph needs a ticket number.

src/feature/stats/bw_array.h
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @file bw_array_st.h
+ * @brief Bandwidth array manipulation functions.
+ **/
+
+#ifndef BW_ARRAY_H
+#define BW_ARRAY_H
+
+#include "feature/stats/bw_array_st.h"
```
teor2345, line 12: Consider replacing the struct declaration with a forward declaration:
```suggestion
struct bw_array_t;
```

src/test/test_stats.c
```diff
@@ -246,6 +247,244 @@ test_rephist_mtbf(void *arg)
...
+  set_test_bw_lines(25000, 27500, 15000, 17500, now - 6*60*60);
+  /* Day 4. */
+  now += 86400;
+  set_test_bw_lines(90000, 76000, 60000, 45000, now - 6*60*60);
+  /* Day 5. */
+  now += 86400;
+  set_test_bw_lines(500, 55000, 30000, 35000, now - 6*60*60);
+  set_test_bw_lines(0, 0, 0, 0, now);
+  format_iso_time(t, now);
+  tor_snprintf(checkstr, len, "write-history %s (86400 s) "
```
teor2345, line 473: Pass the length of checkstr to snprintf(), not a calculated length.
```suggestion
  tor_snprintf(checkstr, sizeof(checkstr), "write-history %s (86400 s) "
```

teor2345: Oh right. Now I understand what's happening here.

It's slightly safer to use stack-allocated buffers for snprintf(), so that the actual size and snprintf size never get out of sync. But this code is also ok.

src/test/test_stats.c
```diff
@@ -30,6 +30,7 @@
 #define CIRCUITLIST_PRIVATE
 #define MAINLOOP_PRIVATE
 #define STATEFILE_PRIVATE
+#define CONFIG_PRIVATE
```
teor2345, line 33: Is CONFIG_PRIVATE actually needed here?

src/feature/stats/bw_array.h
teor2345: This entire header is a private API, as is bw_array_st.h. They should both be protected by _PRIVATE guard macros.

That way, they won't be used outside the rephist module.

src/feature/stats/bw_array.h
```diff
@@ -0,0 +1,28 @@
...
+STATIC bw_array_t *bw_array_new(void);
+STATIC void bw_array_free_(bw_array_t *b);
+STATIC uint64_t find_largest_max(bw_array_t *b);
+STATIC void bw_arrays_init(void);
+STATIC void bw_arrays_free_all(void);
+STATIC size_t
+rep_hist_fill_bandwidth_history(char *buf, size_t len, const bw_array_t *b);
+
+#define bw_array_free(val) \
+  FREE_AND_NULL(bw_array_t, bw_array_free_, (val))
```
teor2345, line 26: This macro is defined twice.

src/feature/stats/rephist.c
```diff
@@ -973,47 +971,67 @@ rep_hist_load_mtbf_data(time_t now)
...
+  b = tor_malloc_zero(sizeof(bw_array_t));
+  rephist_total_alloc += sizeof(bw_array_t);
+  start = time(NULL);
+  b->cur_obs_time = start;
+  b->next_period = start + NUM_SECS_BW_SUM_INTERVAL;
+  return b;
+}
+
+#define bw_array_free(val) \
+  FREE_AND_NULL(bw_array_t, bw_array_free_, (val))
```
teor2345, line 1000: This macro is defined twice.

src/test/test_stats.c
MrSquanchee: No, that wasn't necessary.

src/feature/stats/bw_array.h
MrSquanchee: Fixed this.

src/feature/stats/bw_array.h
MrSquanchee: Fixed it.

src/feature/stats/bw_array.h
MrSquanchee: Thanks for the suggestion. Made the contents of both the files private.

src/feature/stats/bw_array.h
MrSquanchee: Suggestion implemented.

src/feature/stats/rephist.c
MrSquanchee: Fixed This.

PR #1854, opened 2020-04-07
client: Revert setting PreferIPv6 on by default

dgoulet-tor:
This change broke torsocks that by default is expecting an IPv4 for hostname
resolution because it can't ask tor for a specific IP version with the SOCKS5
extension.

PreferIPv6 made it that sometimes the IPv6 could be returned to torsocks that
was expecting an IPv4.

Torsocks is probably a very unique case because the runtime flow is that it
hijacks DNS resolution (ex: getaddrinfo()), gets an IP and then sends it back
for the connect() to happen.

The libc has DNS resolution functions that allows the caller to request a
specific INET family but torsocks can't tell tor to resolve the hostname only
to an IPv4 or IPv6 and thus by default fallsback to IPv4.

Reverting this change into 0.4.3.x series but we'll keep it in the 0.4.4.x
series in the hope that we add this SOCKS5 extension to tor for DNS resolution
and then change torsocks to use that.

Fixes #33804

### DIFF
--- a/doc/tor.1.txt
+++ b/doc/tor.1.txt
@@ -1492,7 +1492,7 @@ The following options are useful only for clients (that is, if
         requests on this connection. This option is only relevant when SOCKS5
         is in use, because SOCKS4 can't handle IPv6.  (Allowing IPv6 is the
         default.)
-    **NoPreferIPv6**;;
+    **PreferIPv6**;;
         Tells exits that, if a host has both an IPv4 and an IPv6 address,
         we would prefer to connect to it via IPv4. (IPv4 is the default.)
     **NoDNSRequest**;;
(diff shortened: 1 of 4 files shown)


doc/tor.1.txt
```diff
@@ -1494,8 +1494,7 @@ The following options are useful only for clients (that is, if
         default.)
     **NoPreferIPv6**;;
```
teor2345, line 1495: We use the non-default flag:
```suggestion
    **PreferIPv6**;;
```

doc/tor.1.txt
```diff
@@ -1494,8 +1494,7 @@ The following options are useful only for clients (that is, if
...
         Tells exits that, if a host has both an IPv4 and an IPv6 address,
-        we would prefer to connect to it via IPv4. (IPv6 is the default in
-        recent versions of Tor.)
+        we would prefer to connect to it via IPv4. (IPv4 is the default.)
```
teor2345, line 1497: We use the non-default flag:
```suggestion
        we would prefer to connect to it via IPv6. (IPv4 is the default.)
```

teor2345:
Squashed and merged to 0.4.3 only.

PR #1858, opened 2020-04-08
Ticket33072 042 01

dgoulet-tor:


scripts/maint/practracker/exceptions.txt
```diff
@@ -56,7 +56,7 @@ problem dependency-violation /src/core/crypto/onion_crypto.c 5
 problem dependency-violation /src/core/crypto/onion_fast.c 1
 problem dependency-violation /src/core/crypto/onion_tap.c 3
 problem dependency-violation /src/core/crypto/relay_crypto.c 9
-problem file-size /src/core/mainloop/connection.c 5569
+problem file-size /src/core/mainloop/connection.c 5694
```
teor2345, line 59: Wow, you shouldn't need to fix practracker on 0.4.2 or 0.4.3.

I have removed check-best-practices from check-local on all maint and release branches. So practracker only runs by default on master.

src/core/mainloop/connection.c
```diff
@@ -3190,8 +3190,11 @@ connection_bucket_write_limit(connection_t *conn, time_t now)
  * shouldn't send <b>attempt</b> bytes of low-priority directory stuff
  * out to <b>conn</b>.
  *
- * If we are a directory authority, always answer dir requests thus true is
- * always returned.
+ * If we are a directory authority, false is returned (indicating that we
+ * should answer the request) if one of these conditions is met:
+ *  - Connection is from a known relay (address is looked up).
+ *  - AuthDirRejectRequestsUnderLoad is set to 0.
+ *  - Compression is being used for the request (looking at c_method).
```
teor2345, line 3197: This comment doesn't match what the function does in 0.4.3.

changes/ticket33072
```diff
@@ -0,0 +1,4 @@
+  o Minor bugfix (directory authority):
+    - Always allow compressed directory requests and introduce option
```
teor2345, line 2: Oh, the changes file is out of sync as well.

We let AuthDirRejectRequestsUnderLoad and the bandwidth buckets decide about compressed requests.

PR #1864, opened 2020-04-15
Ticket33817

teor2345:


src/lib/net/address.c
```diff
@@ -903,6 +903,20 @@ tor_addr_from_in6(tor_addr_t *dest, const struct in6_addr *in6)
   tor_addr_from_ipv6_bytes(dest, (const char*)in6->s6_addr);
 }
 
+/** Set the 16 bytes at <b>dest</b> to equal the IPv6 address <b>src</b>.
+ * <b>src</b> must be an IPv6 address, if it is not, log a warning, and clear
+ * <b>dest</b>. */
+void
+tor_addr_get_ipv6_bytes(char *dest, const tor_addr_t *src)
```
nmathewson, line 910: This should use uint8_t, not char for the output.

Would "copy" make more sense than "get" for the function name?

src/feature/relay/circuitbuild_relay.c
```diff
@@ -123,31 +123,38 @@ circuit_extend_add_ed25519_helper(struct extend_cell_t *ec)
...
  * If they are valid, return 0.
- * Otherwise, if they are invalid, log a warning at <b>log_level</b>,
- * and return -1.
+ * Otherwise, if they are invalid, return -1.
+ * If <b>log_zero_addrs</b> is true, log warnings about zero addresses at
+ * <b>log_level</b>. If <b>log_internal_addrs</b> is true, log warnings about
+ * internal addresses at <b>log_level</b>.
  */
 static int
 circuit_extend_addr_port_helper(const struct tor_addr_port_t *ap,
```
nmathewson, line 135: We should probably rename this function so that it reflects the predicate that it evaluates.

src/core/or/channel.c
```diff
@@ -2409,11 +2412,15 @@ channel_get_for_extend,(const char *rsa_id_digest,
       continue;
     }
 
+    const int matches_target =
```
nmathewson, line 2415: const bool instead?

src/test/test_circuitbuild.c
```diff
@@ -219,6 +219,7 @@ test_circuit_extend_state_valid(void *arg)
   expect_log_msg("Got an extend cell, but running as a client. Closing.\n");
   mock_clean_saved_logs();
 
+#ifndef ALL_BUGS_ARE_FATAL
```
nmathewson, line 223: In cases where these blocks cover an entire test, the test should use TT_SKIP() or tt_skip() rather than ifdef.  But that can be another ticket.

src/feature/relay/circuitbuild_relay.c
```diff
@@ -258,13 +266,61 @@ circuit_open_connection_for_extend(const struct extend_cell_t *ec,
     return;
   }
 
+  /* Check the addresses, without logging */
```
nmathewson, line 318: This part would probably like to be its own function, since it's logically disconnected from the other stuff.

src/feature/relay/circuitbuild_relay.c
```diff
@@ -258,13 +266,61 @@ circuit_open_connection_for_extend(const struct extend_cell_t *ec,
...
+   * at random, with probability 1 in N.
+   *   1 means "always IPv6 (and no IPv4)"
+   *   2 means "equal probability of IPv4 or IPv6"
+   *   ... (and so on) ...
+   *   (UINT_MAX - 1) means "almost always IPv4 (and almost never IPv6)"
+   * To disable IPv6, set ipv6_supported to 0.
+   */
+#define IPV6_CONNECTION_ONE_IN_N 2
+
+  const bool ipv6_supported = router_has_advertised_ipv6_orport(get_options());
```
nmathewson: I think we should have a different function for this. For now it can just alias `router_has_advertised_ipv6_orport`, but someday we'll want to separate the idea of "I can connect to IPv6 addresses" and "I have an IPv6 address."

src/feature/relay/circuitbuild_relay.c
nmathewson: Ah, looks like you split it later on.

src/feature/relay/circuitbuild_relay.c
teor2345: Sounds sensible.

Let's call this one "router_can_extend_to_ipv6_orports()", because it's also possible that:
* relays or clients try connecting to IPv6 ORPorts, to see if they themselves have IPv6
* clients may be configured to connect to IPv6 ORPorts, without extending

src/test/test_circuitbuild.c
teor2345: I just checked all instances of ALL_BUGS_ARE_FATAL, and I couldn't find any which:
* cover the entire unit test, and
* #ifdef out the code.
That code wouldn't compile anyway, because "done" would be unused.

For future reference, if we do find any of these blocks, they need a coverity / clang analyser wrapper, like this:
https://github.com/torproject/tor/blob/3253c357eeae3434da62bf720a451aa19f0ddd32/src/test/test_circuitbuild.c#L130

src/lib/net/address.c
teor2345: I matched tor_addr_from_ipv6_bytes(), but I'll change them both to uint8_t.

I can also change the function name.

teor2345: There were only 3 address.h functions that still used `char *` for bytes, so I fixed them all in cd7e2fc210349615c8e9cdb879f98cb0c9cac57b.

Rename in f62b051e873aa4be564e9f83745902d4541f79f4.

src/feature/relay/circuitbuild_relay.c
teor2345: Renamed and returned bool in 496ddd2877051a43aaa7bbeb68bd5b255c71adc7. The code is much nicer now.

src/core/or/channel.c
teor2345: Returned bool and changed variable to bool in 066d2deb3d322e960600d80739739199d8c4cfa6.

src/feature/relay/circuitbuild_relay.c
teor2345: Used a slightly different name in 15a4180a7e041531d6923a41684e9e6ffc833760, also renamed all the mocking functions and variables.

PR #1870, opened 2020-04-26
[Tor Trac #33956]: Define and use TOR_ADDRPORT_BUF_LEN

neelchauhan:
PR: https://trac.torproject.org/projects/tor/ticket/33956

### DIFF
--- a/src/lib/net/address.c
+++ b/src/lib/net/address.c
@@ -1173,7 +1173,6 @@ fmt_addr_impl(const tor_addr_t *addr, int decorate)
 const char *
 fmt_addrport(const tor_addr_t *addr, uint16_t port)
 {
-  /* Add space for a colon and up to 5 digits. */
   static char buf[TOR_ADDRPORT_BUF_LEN];
   tor_snprintf(buf, sizeof(buf), "%s:%u", fmt_and_decorate_addr(addr), port);
   return buf;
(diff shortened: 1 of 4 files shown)


src/lib/net/address.h
```diff
@@ -209,6 +209,15 @@ tor_addr_eq_ipv4h(const tor_addr_t *a, uint32_t u)
...
 #define TOR_ADDR_BUF_LEN 48
 
+/** Length of a buffer containing an IP address along with a port number and
+ * a seperating colon.
+ *
+ * This allows enough space for
+ *   "[ffff:ffff:ffff:ffff:ffff:ffff:255.255.255.255]:12345",
+ * plus a terminating NUL.
+ */
+#define TOR_ADDRPORT_BUF_LEN TOR_ADDR_BUF_LEN + 6
```
teor2345, line 219: Please include brackets in the macro definition. Otherwise, operators in the surrounding code could bind to parts of the macro definition, leading to an incorrect result.

For example, `TOR_ADDRPORT_BUF_LEN*2` becomes `TOR_ADDR_BUF_LEN + (6*2)`, which is not what we want.
```suggestion
#define TOR_ADDRPORT_BUF_LEN (TOR_ADDR_BUF_LEN + 6)
```

src/lib/net/address.c
```diff
@@ -1174,7 +1174,7 @@ const char *
 fmt_addrport(const tor_addr_t *addr, uint16_t port)
 {
   /* Add space for a colon and up to 5 digits. */
```
teor2345, line 1176: This comment doesn't make sense any more.
```suggestion
```

teor2345:
Squashed and merged to master.

PR #1873, opened 2020-04-29
Prop310 rebase

frochet:


src/feature/client/entrynodes.h
```diff
@@ -117,7 +117,7 @@ struct entry_guard_t {
    * confirmed guard. */
   time_t confirmed_on_date; /* 0 if not confirmed */
   /**
-   * In what order was this guard sampled without replacement? Guards with
+   * In what order was this guard sampled? Guards with
    * lower indices appear earlier on the sampled list
    */
```
asn-d6, line 122: Hm, if I understand correctly `sampled_idx` influences both the order of the sampled list but also the order of the confirmed list, right? In this case this comment should clarify this since it only mentions the sampled list atm.

Further, what is the relationship between `sampled_idx` and `confirmed_idx`? It seems like both of them influence the order of the confirmed list, but is `sampled_idx` what actually dictates the order in the end? I might be confused but some docs would be nice here.

frochet: Yes, i've added a comment simply saying the sampled ordering in used over all sampled, confirmed and primary lists.

There is actually no much use for any ordering by confirmed idx, but we still use it to detect whether the confirmed list changed. However this might be done with the sampled ordering instead. It does not really matter which ordering is used to detect change. We may just remove confirmed ordering if it brings confusion.

PR #1874, opened 2020-04-30
Ticket33222

teor2345:


src/feature/relay/selftest.c
```diff
@@ -70,7 +70,7 @@ router_reachability_checks_disabled(const or_options_t *options)
  *   - the network is disabled.
  */
 int
-check_whether_orport_reachable(const or_options_t *options)
+router_skip_orport_reachability_check(const or_options_t *options)
```
nmathewson, line 80: I would suggest a different name -- this one seems to imply that the function _makes us skip_ reachability checks, when it actually tests _whether we are skipping_ those checks.  Maybe `router_should_skip...` or `router_is_skipping...`?

src/feature/nodelist/routerinfo.c
```diff
@@ -17,14 +17,31 @@
 #include "feature/nodelist/node_st.h"
 #include "feature/nodelist/routerinfo_st.h"
 
-/** Copy the primary (IPv4) OR port (IP address and TCP port) for
- * <b>router</b> into *<b>ap_out</b>. */
+/** Copy the OR port (IP address and TCP port) for <b>router</b> and
+ * <b>family</b> into *<b>ap_out</b>.
+ *
+ * If the requested ORPort does not exist, sets *<b>ap_out</b> to the null
```
nmathewson, line 23: This API is a bit strange to me.  Why not also return 0 or -1 depending on whether the port exists?

src/feature/relay/selftest.c
```diff
@@ -242,7 +255,9 @@ router_do_reachability_checks(int test_or, int test_dir)
 
   if (router_should_check_reachability(test_or, test_dir)) {
     if (test_or && (!orport_reachable || !circuit_enough_testing_circs())) {
-      router_do_orport_reachability_checks(me, orport_reachable);
+      /* We'll separate IPv4 and IPv6 reachability detection in #34067. */
```
nmathewson, line 258: I'm guessing that this comment means we need separate "orport_reachable" variables for an ipv6 and ipv4 address

src/feature/relay/selftest.c
```diff
@@ -281,20 +281,28 @@ inform_testing_reachability(void)
...
+  const bool has_ipv6 = tor_addr_port_is_valid(&me->ipv6_addr,
+                                               me->ipv6_orport, 0);
+  if (has_ipv6) {
+    tor_snprintf(extra_addrbuf, sizeof(extra_addrbuf), " and %s",
+                 fmt_addrport(&me->ipv6_addr, me->ipv6_orport));
+    /* We'll add an IPv6 control event in #34068. */
+  }
   if (me->dir_port) {
-    tor_snprintf(dirbuf, sizeof(dirbuf), " and DirPort %s:%d",
+    tor_snprintf(extra_addrbuf, sizeof(extra_addrbuf), " and DirPort %s:%d",
```
nmathewson, line 292: This printf and the one above are bouth writing into extra_addrbuf. I think this one will overwrite the previous under some circumstancs.

src/feature/relay/selftest.c
```diff
@@ -272,7 +272,7 @@ router_do_reachability_checks(int test_or, int test_dir)
 int
 inform_testing_reachability(void)
```
nmathewson, line 273: It would probably be better for this function to take as an argument which ports it is testing.  Otherwise the message it sends risks getting out of sync with the tests that were actually launched.

(That is to say, this function claims that an IPv6 orport check was launched when an IPv6 address exists -- not directly when an IPv6 orport check was launched.)

teor2345: I agree in principle.

But we don't report that we're testing reachability until at least one circuit has finished extending. (That circuit can be any purpose, not just testing reachability.)

So there are a bunch of subtle constraints and race conditions here. I split this change into another ticket:
https://trac.torproject.org/projects/tor/ticket/34137

And documented the issues in the function, see 492c512af4a4f9ee143b945296fd1c61fc68dba1.

src/feature/relay/selftest.c
teor2345: You're right, I was tripped up by the inconsistent patterns in this function. I made them more consistent in 504b16fb757d14f58e20294f7aca338dc6826140.

src/feature/relay/selftest.c
teor2345: We'll need separate variables, and potentially separate flags that make up each variable. I clarified the comment in e887887cd32f9e78e3b6e9c47661eaf750b4c0fb.

See also https://trac.torproject.org/projects/tor/ticket/33222 , particularly:
> 4.2.3. Separate IPv4 and IPv6 Reachability Flags
> 
> Testing relays (and bridges) will record reachability separately for IPv4
> and IPv6 ORPorts, based on the ORPort that the test circuit was received on.
> 
> Here is a reliable way to do reachability self-tests for each ORPort:
> 
> 1. Check for create cells on inbound ORPort connections 
> 
> Check for a cell on any IPv4 and any IPv6 ORPort. (We can't know which listener(s) correspond to the advertised ORPorts, particularly when using port forwarding.) Make sure the cell was received on an inbound OR connection.
> 
> 2. Check for created cells from testing circuits on outbound OR connections 
> 
> Check for a returned created cell on our IPv4 and IPv6 self-test circuits. Make sure those circuits were on outbound OR connections.
> 
> By combining these tests, we confirm that we can:
> * reach our own ORPorts with testing circuits
> * send and receive cells via inbound OR connections to our own ORPorts
> * send and receive cells via outbound OR connections to other relays' ORPorts 
> 
> This isn't a perfect test, there are a few false positives...

src/feature/nodelist/routerinfo.c
teor2345: Sure, see c76f310fe724c3d93416b4e23a219feda8e6f217.

I've possibly been writing too much Rust, where errors are part of the returned type :-)

src/feature/nodelist/routerlist.c
```diff
@@ -469,7 +469,7 @@ router_reload_router_list(void)
  * and reachability checks be skipped?
  */
 int
-router_skip_or_reachability(const or_options_t *options, int try_ip_pref)
+router_connect_assume_or_reachable(const or_options_t *options, int try_ip_pref)
```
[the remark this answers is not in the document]
teor2345: I have made the names longer and hopefully clearer in 05c56ae455de1f8e1dc9816ab9dcd7f0cdc2c994 and e1eecac88866fa66e9d77c251a31d85131f11aef.

src/feature/relay/selftest.c
teor2345: I have made the names longer and hopefully clearer in 05c56ae455de1f8e1dc9816ab9dcd7f0cdc2c994.

src/core/or/or.h
```diff
@@ -841,12 +845,9 @@ typedef struct protover_summary_flags_t {
   unsigned int supports_v3_rendezvous_point: 1;
 
   /** True iff this router has a protocol list that allows clients to
-   * negotiate hs circuit setup padding. Requires Padding>=2. */
+   * negotiate hs circuit setup padding. Requires Padding=2. */
```
nmathewson, line 860: Good catch that our current behavior is =, not >=.  But do you suppose that our behavior should be >= instead?  If so I can open a ticket to write a proposal.

src/feature/nodelist/routerlist.c
```diff
@@ -541,6 +541,11 @@ router_add_running_nodes_to_smartlist(smartlist_t *sl, int need_uptime,
     /* Don't choose nodes if we are certain they can't do ntor. */
     if ((node->ri || node->md) && !node_has_curve25519_onion_key(node))
       continue;
+    /* Exclude relays that allow single hop exit circuits. This is an
+     * obsolete option since 0.2.9.2-alpha and done by default in
+     * 0.3.1.0-alpha. */
+    if (node_allows_single_hop_exits(node))
```
nmathewson, line 576: This refactoring looks okay, but is this tested?

src/feature/nodelist/routerlist.c
```diff
@@ -546,6 +547,11 @@ router_add_running_nodes_to_smartlist(smartlist_t *sl, int need_uptime,
      * 0.3.1.0-alpha. */
     if (node_allows_single_hop_exits(node))
       continue;
+    /* Exclude relays that can not become a rendezvous for a hidden service
+     * version 3. */
+    if (rendezvous_v3 &&
```
nmathewson, line 580: This refactoring looks okay, but is this tested?

src/feature/nodelist/routerlist.c
```diff
@@ -520,6 +520,7 @@ router_add_running_nodes_to_smartlist(smartlist_t *sl, int need_uptime,
                                       int need_capacity, int need_guard,
                                       int need_desc, int pref_addr,
                                       int direct_conn,
+                                      bool rendezvous_v3,
```
nmathewson: The number of arguments to this function are getting out of hand.  Can we have a ticket to use flags or a struct instead?

nmathewson: Ah never mind, it looks like you did that in a later commit.

src/feature/nodelist/node_select.c
```diff
@@ -947,7 +947,6 @@ router_choose_random_node(smartlist_t *excludedsmartlist,
   const int need_uptime = (flags & CRN_NEED_UPTIME) != 0;
   const int need_capacity = (flags & CRN_NEED_CAPACITY) != 0;
   const int need_guard = (flags & CRN_NEED_GUARD) != 0;
-  const int weight_for_exit = (flags & CRN_WEIGHT_AS_EXIT) != 0;
```
nmathewson, line 957: qq: why was this unused?

src/core/or/or.h
teor2345: There are a few instances of ">=" and "=" confusion in protocol versions. We should try to fix them all.

It only matters when we remove protocol versions. We haven't really specified, tested, or exercised this functionality in practice. And so our reviewers lack experience. (And when we did discover a need for it with NSS and LinkAuth, it was more complicated than we expected.)

I'd like to see a proposal that tells us how to check future protocol versions as they are added. Along with a migration plan for disabling dependent and orthogonal protocol versions.

I originally thought we might want an "all future versions" check. But that's a bad idea, because that check would make it impossible to add a version that disables a feature. (In that case, we would choose a new name for the new feature.)

So let's also open a ticket to use the versions that currently support the feature being tested. We should:
* replace all ">=" checks (unless there is a specific exception), and
* replace all "=" checks.
Let's make sure we check all the places where we use protocol versions, even if they don't have a summary flag.

There's only one ">=" check exception right now: ed25519 link specifiers for extends. We could assume that any future link authentication method would use ed25519 keys, or safely ignore them. Or we could assume that any future protocol that dropped ed25519 keys for direct connections, would also make them unsuitable for extends.

Let's try to avoid designs where we have to predict the future :-)

Overall, I think it would be helpful if future protocol versions were orthogonal. Or if they depend on earlier features, that dependency should be clearly documented. (For example, Relay=3 IPv6 extends depends on Relay=2 EXTEND2 cells. So if we were checking EXTEND2 cell support, it would be Relay=2 or Relay=3.)

src/feature/nodelist/node_select.c
teor2345: Because choose_good_exit_server_general() contains an outdated copy of parts of:
* router_choose_random_node()
* router_choose_random_node_helper()
* router_add_running_nodes_to_smartlist()
All interspersed with code that checks exits against pending circuits.

So there are probably some checks that aren't being:
* added to the flags for exits, or
* checked by choose_good_exit_server_general().

It would be great to resolve this technical debt, but it's out of scope for this pull request, and Sponsor 55. Let's open a ticket?

teor2345: Actually, there's a quick refactor that fixes 90% of this issue. It requires creating another function, but I think that's better than duplication of a whole bunch of checks.

Then we can open a ticket for the remaining work.

teor2345: See https://trac.torproject.org/projects/tor/ticket/34200

src/feature/nodelist/routerlist.c
teor2345: This commit is now in https://github.com/torproject/tor/pull/1888

src/feature/nodelist/routerlist.c
teor2345: This commit is now in https://github.com/torproject/tor/pull/1888

src/core/or/or.h
teor2345: I've changed my mind, I think we want to check for "any later protocol version".

See my tor-dev email here:
https://lists.torproject.org/pipermail/tor-dev/2020-May/014297.html

PR #1896, opened 2020-05-17
[Tor Trac 34238]: Space out some function calls in parse_short_policy()

neelchauhan:
Ticket: https://trac.torproject.org/projects/tor/ticket/34238

### DIFF
--- a/src/core/or/policies.c
+++ b/src/core/or/policies.c
@@ -2780,7 +2780,8 @@ parse_short_policy(const char *summary)
         high = low;
         break;
       case '-':
-        high = (unsigned) tor_parse_ulong(next+1, 10, low, 65535, &ok, &next);
+        high = (unsigned)
+          tor_parse_ulong(next + 1, 10, low, 65535, &ok, &next);
         if (!ok)
           goto bad_ent;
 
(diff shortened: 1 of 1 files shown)


src/core/or/policies.c
```diff
@@ -2780,7 +2780,8 @@ parse_short_policy(const char *summary)
         high = low;
         break;
       case '-':
-        high = (unsigned) tor_parse_ulong(next+1, 10, low, 65535, &ok, &next);
+        high = (unsigned)
+          tor_parse_ulong(next + 1, 10, low, 65535, &ok, &next);
```
ahf, line 2784: This looks odd to me. Unless we are low on horizontal space, then I see no need to do something like this. Getting rid of the space between `(unsigned)` and the function is fine IMO.

src/core/or/policies.c
```diff
@@ -2817,12 +2818,13 @@ parse_short_policy(const char *summary)
       sizeof(short_policy_entry_t)*(n_entries);
     result = tor_malloc_zero(size);
 
-    tor_assert( (char*)&result->entries[n_entries-1] < ((char*)result)+size);
+    tor_assert((char *) &result->entries[n_entries - 1] <
+               ((char *) result) + size);
```
ahf, line 2822: Same as above. This shouldn't be split on multiple lines, and I don't think we should have a space between the type-cast and the variables.

src/core/or/policies.c
```diff
@@ -2817,12 +2818,13 @@ parse_short_policy(const char *summary)
...
   }
 
   result->is_accept = is_accept;
   result->n_entries = n_entries;
-  memcpy(result->entries, entries, sizeof(short_policy_entry_t)*n_entries);
+  memcpy(result->entries, entries, sizeof(short_policy_entry_t) * n_entries);
```
ahf, line 2827: Seems fine. This is the multiply operator and not a cast.

PR #1898, opened 2020-05-17
[Tor Trac #24844]: Add HS v3 status to the SIGUSR1 dumpstats()

neelchauhan:
Ticket: https://trac.torproject.org/projects/tor/ticket/24844

### DIFF
--- a/src/feature/hs/hs_service.c
+++ b/src/feature/hs/hs_service.c
@@ -4111,9 +4111,7 @@ hs_service_dump_stats(int severity)
 {
   origin_circuit_t *circ;
 
-  for (hs_service_t **iter = HT_START(hs_service_ht, hs_service_map);
-       iter != NULL; iter = HT_NEXT(hs_service_ht, hs_service_map, iter)) {
-    hs_service_t *hs = *iter;
+  FOR_EACH_SERVICE_BEGIN(hs) {
 
     tor_log(severity, LD_GENERAL, "Service configured in %s:",
             service_escaped_dir(hs));
@@ -4135,7 +4133,7 @@ hs_service_dump_stats(int severity)
       } DIGEST256MAP_FOREACH_END;
 
     } FOR_EACH_DESCRIPTOR_END;
-  }
+  } FOR_EACH_SERVICE_END;
 }
 
 /** Put all service object in the given service list. After this, the caller
--- a/src/feature/hs/hs_service.c
+++ b/src/feature/hs/hs_service.c
@@ -4123,9 +4123,13 @@ hs_service_dump_stats(int severity)
         const char *nickname;
 
         intro_node = get_node_from_intro_point(ip);
-        tor_assert(intro_node);
+        if (!intro_node) {
+          continue;
+        }
         nickname = node_get_nickname(intro_node);
-        tor_assert(nickname);
+        if (!nickname) {
+          continue;
+        }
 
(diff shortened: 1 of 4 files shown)


src/feature/hs/hs_service.c
```diff
@@ -4103,6 +4103,41 @@ hs_service_load_all_keys(void)
...
 
+/** Log the status of introduction points for all version 3 onion services
+ * at log severity <b>severity</b>.
+ */
+void
+hs_service_dump_stats(int severity)
+{
+  origin_circuit_t *circ;
+
+  for (hs_service_t **iter = HT_START(hs_service_ht, hs_service_map);
```
asn-d6, line 4114: Please check out `FOR_EACH_SERVICE_BEGIN`. For example see how `update_all_descriptors_intro_points()` does it. You've already used `FOR_EACH_DESCRIPTOR_BEGIN` so it should be similar.

src/feature/hs/hs_service.c
```diff
@@ -4103,6 +4103,41 @@ hs_service_load_all_keys(void)
...
+       iter != NULL; iter = HT_NEXT(hs_service_ht, hs_service_map, iter)) {
+    hs_service_t *hs = *iter;
+
+    tor_log(severity, LD_GENERAL, "Service configured in %s:",
+            service_escaped_dir(hs));
+    FOR_EACH_DESCRIPTOR_BEGIN(hs, desc) {
+
+      DIGEST256MAP_FOREACH(desc->intro_points.map, key,
+                           hs_service_intro_point_t *, ip) {
+        const node_t *intro_node = get_node_from_intro_point(ip);
```
asn-d6, line 4124: I think we should check that `intro_node` is not NUL. We've had bugs in the past with this one.

src/feature/hs/hs_service.c
```diff
@@ -4103,6 +4103,41 @@ hs_service_load_all_keys(void)
...
+        const char *nickname = node_get_nickname(intro_node);
```
asn-d6, line 4125: Just to be ultra safe let's also check that `nickname` is not NULL

src/feature/hs/hs_service.c
```diff
@@ -4119,8 +4119,13 @@ hs_service_dump_stats(int severity)
 
       DIGEST256MAP_FOREACH(desc->intro_points.map, key,
                            hs_service_intro_point_t *, ip) {
-        const node_t *intro_node = get_node_from_intro_point(ip);
-        const char *nickname = node_get_nickname(intro_node);
+        const node_t *intro_node;
+        const char *nickname;
+
+        intro_node = get_node_from_intro_point(ip);
+        tor_assert(intro_node);
```
asn-d6, line 4126: Please don't assert here (or below) and instead do a regular check like you are doing for `circ` below.

src/feature/hs/hs_service.c
```diff
@@ -4123,9 +4123,13 @@ hs_service_dump_stats(int severity)
         const char *nickname;
 
         intro_node = get_node_from_intro_point(ip);
-        tor_assert(intro_node);
+        if (!intro_node) {
```
asn-d6, line 4126: This first `intro_node` check is actually kinda plausible to happen so perhaps we can add a log here, like we did in the !circ case? We've seen it out in the wild.

I think the nickname check is much less likely to happen so I don't care about a log there, but perhaps we can add one anyway.

PR #1899, opened 2020-05-18
Add support for console control signals in Windows

TheDcoder:
This pull request is for Ticket [#34211](https://trac.torproject.org/projects/tor/ticket/34211):

> Hi everyone!
> 
> I am working on a [cross-platform Tor controller/wrapper](https://github.com/DcodingTheWeb/ProxAllium/tree/next-gen), and I encountered a minor inconvenience while writing code for stopping Tor in Windows.
> 
> Currently the only way to gracefully terminate Tor in Windows is to use the control port to issue the `SIGINT` signal, it would be beneficial for 3rd party programs which work with Tor to be able to invoke this procedure by using the native "control signals" provided to console programs in Windows.
> 
> I have made a patch which implements this functionality, I will post it soon, I am creating this ticket so that I can create a corresponding file in the `changes` directory :)
> 
> Thanks to nickm for some pointers about using the libevent `event_active` function to trigger the event, I decided to use the `activate_signal` function as it internally calls the suggested function and simplifies the code.

### DIFF
--- a/changes/ticket34211
+++ b/changes/ticket34211
@@ -1,0 +1,3 @@
+  o Minor features (windows):
+    - Add support for console control signals like Ctrl+C in Windows
+      Closes ticket 34211. Patch from Damon Harris (TheDcoder).
--- a/src/app/main/main.c
+++ b/src/app/main/main.c
@@ -294,6 +294,19 @@ process_signal(int sig)
   }
 }
 
+#ifdef _WIN32
+/** Activate SIGINT on reciving a control signal in console */
+static BOOL WINAPI
+process_win32_console_ctrl(DWORD ctrl_type)
+{
+  /* Ignore type of the ctrl signal */
+  (void) ctrl_type;
+
+  activate_signal(SIGINT);
+  return TRUE;
+}
+#endif
+
 /**
  * Write current memory usage information to the log.
  */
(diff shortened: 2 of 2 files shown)


changes/ticket34211
```diff
@@ -0,0 +1,3 @@
+  o Minor features (control signal in windows):
```
ahf, line 1: The category we usually use here is just: `o Minor features (windows)`. Then it will be grouped together with the other windows changes.

src/app/main/main.c
```diff
@@ -294,6 +294,17 @@ process_signal(int sig)
   }
 }
 
+#ifdef _WIN32
+/** Activate SIGINT on reciving a control signal in console */
+static BOOL WINAPI process_win32_console_ctrl(DWORD ctrl_type) {
```
ahf, line 299: Looks like CI is failing because of the whitespace here. Usually we keep the type on a line on its own, so if you reformat this to:

```
static BOOL WINAPI
process_win32_console_ctrl(DWORD ctrl_type) {
```

Then I think `make check-spaces` will be happy again :-)

TheDcoder: Ah! I didn't even pay attention to this convention, I guess old habits die hard :)

changes/ticket34211
TheDcoder: Thanks for the pointer, I have changed it!

src/app/main/main.c
```diff
@@ -294,6 +294,18 @@ process_signal(int sig)
   }
 }
 
+#ifdef _WIN32
+/** Activate SIGINT on reciving a control signal in console */
+static BOOL WINAPI
+process_win32_console_ctrl(DWORD ctrl_type) {
```
ahf, line 300: Okay, I missed this one because of the type-name on its own line situation. The `{` at the end of the function line have to go on its own line to satisfy the code style checker.

TheDcoder: Sorry, I kept forgetting to pay attention to the style, I was a bit busy so I didn't pay much attention. However I have now made several stylistic changes and the code should match other code now, I even ran `make check` locally this time to make sure :smile:

I did not get any errors related to style, but there were some failed checks in unrelated areas. Hope everything goes smoothly this time!

ahf:
This looks good! Thank you for working on this. Our review process is always a bit ping/pong back and forward :-)

PR #1902, opened 2020-05-21
#33788: Make things safer when tor_inet_ntop() and tor_inet_ntoa() fail (2)

rl1987:
https://trac.torproject.org/projects/tor/ticket/33788

### DIFF
--- a/src/feature/relay/dns.c
+++ b/src/feature/relay/dns.c
@@ -1881,7 +1881,7 @@ evdns_wildcard_check_callback(int result, char type, int count, int ttl,
         int ntoa_res;
         in.s_addr = addrs[i];
         ntoa_res = tor_inet_ntoa(&in, answer_buf, sizeof(answer_buf));
-        tor_assert_nonfatal(ntoa_res > 0);
+        tor_assert_nonfatal(ntoa_res >= 0);
         if (ntoa_res > 0)
           wildcard_increment_answer(answer_buf);
       }
--- a/src/feature/relay/dns.c
+++ b/src/feature/relay/dns.c
@@ -1591,12 +1591,17 @@ evdns_callback(int result, char type, int count, int ttl, void *addresses,
     } else if (type == DNS_IPv6_AAAA && count) {
       char answer_buf[TOR_ADDR_BUF_LEN];
       char *escaped_address;
+      const char *ip_str;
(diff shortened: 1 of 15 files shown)


src/feature/relay/dns.c
```diff
@@ -1872,16 +1878,22 @@ evdns_wildcard_check_callback(int result, char type, int count, int ttl,
       for (i = 0; i < count; ++i) {
         char answer_buf[INET_NTOA_BUF_LEN+1];
         struct in_addr in;
+        int ntoa_res;
         in.s_addr = addrs[i];
-        tor_inet_ntoa(&in, answer_buf, sizeof(answer_buf));
-        wildcard_increment_answer(answer_buf);
+        ntoa_res = tor_inet_ntoa(&in, answer_buf, sizeof(answer_buf));
+        tor_assert_nonfatal(ntoa_res > 0);
```
nmathewson, line 1884: This is incorrect; look at the documentation for tor_inet_ntoa().  It says "returns -1 on failure" and "returns non-negative integer on success."  So the test should be `>= 0`.

src/lib/net/address.c
```diff
@@ -1196,14 +1196,24 @@ fmt_addrport(const tor_addr_t *addr, uint16_t port)
...
 {
   static char buf[INET_NTOA_BUF_LEN];
   struct in_addr in;
+  int success;
+
   in.s_addr = htonl(addr);
-  tor_inet_ntoa(&in, buf, sizeof(buf));
+
+  success = tor_inet_ntoa(&in, buf, sizeof(buf));
+  tor_assertf_nonfatal(success > 0,
```
nmathewson, line 1210: Should be >= 0.

src/lib/net/address.c
```diff
@@ -1196,14 +1196,24 @@ fmt_addrport(const tor_addr_t *addr, uint16_t port)
...
+      "Failed to convert IP %04X to string", addr);
+
+  if (success <= 0) {
```
nmathewson, line 1213: Should be < 0.  And may we should be using IF_BUG_ONCE for this?

src/feature/relay/dns.c
rl1987: All three check for `tor_inet_ntoa()` return value fixed.

PR #1907, opened 2020-05-28
[Tor Trac #28279]: control: Add a key to GETINFO to fetch the circuit onion handshake rephist values

neelchauhan:
Ticket: https://trac.torproject.org/projects/tor/ticket/28279


src/feature/control/control_getinfo.c
```diff
@@ -1436,6 +1437,37 @@ getinfo_helper_liveness(control_connection_t *control_conn,
   return 0;
 }
 
+/** Implementation helper for GETINFO: answers queries about shared random
+ * value. */
```
dgoulet-tor, line 1441: Comment is not accurate. Most likely copy paste :).

src/feature/control/control_getinfo.c
```diff
@@ -1436,6 +1437,37 @@ getinfo_helper_liveness(control_connection_t *control_conn,
...
+    result =
+      rep_hist_get_circuit_handshake_requested(ONION_HANDSHAKE_TYPE_NTOR);
+  } else if (!strcmp(question, "rephist/tap/onion_handshakes_assigned")) {
+    result =
+      rep_hist_get_circuit_handshake_assigned(ONION_HANDSHAKE_TYPE_TAP);
+  } else if (!strcmp(question, "rephist/tap/onion_handshakes_requested")) {
+    result =
+      rep_hist_get_circuit_handshake_requested(ONION_HANDSHAKE_TYPE_TAP);
+  }
+  /* Else statement here is unrecognized key so do nothing. */
```
dgoulet-tor, line 1464: Shouldn't we send an error here? Spec specifies:

`  The server sends a 551 or 552 error on failure.`

src/feature/control/control_getinfo.c
```diff
@@ -1436,6 +1437,37 @@ getinfo_helper_liveness(control_connection_t *control_conn,
...
+/** Implementation helper for GETINFO: answers queries about shared random
+ * value. */
+STATIC int
+getinfo_helper_rephist(control_connection_t *control_conn,
+                       const char *question, char **answer,
+                       const char **errmsg)
+{
+  (void) control_conn;
+  (void) errmsg;
+  int result = -1;
```
dgoulet-tor, line 1449: I would not assign it here so the compiler can tell us when we use it unassigned. And I don't think we should ever use -1 as for an unknown `question`, an control error code should be returned?

src/feature/stats/rephist.c
```diff
@@ -2455,6 +2455,20 @@ rep_hist_note_circuit_handshake_assigned(uint16_t type)
     onion_handshakes_assigned[type]++;
 }
 
+/** Get the circuit handshake value that is requested. */
+MOCK_IMPL(int,
+rep_hist_get_circuit_handshake_requested, (uint16_t type))
+{
```
dgoulet-tor, line 2461: We should be extra extra careful here with `type` so I would recommend you `BUG()` on `type > MAX_ONION_HANDSHAKE_TYPE` and return 0 as in "no stats".

src/feature/control/control_getinfo.c
```diff
@@ -1660,6 +1692,16 @@ static const getinfo_item_t getinfo_items[] = {
...
+  PREFIX("rephist/ntor", rephist, "NTor circuit handshake stats."),
+  ITEM("rephist/ntor/onion_handshakes_assigned", rephist,
+       "Assigned NTor circuit handshake stats."),
+  ITEM("rephist/ntor/onion_handshakes_requested", rephist,
+       "Requested NTor circuit handshake stats."),
+  PREFIX("rephist/tap", rephist, "TAP circuit handshake stats."),
+  ITEM("rephist/tap/onion_handshakes_assigned", rephist,
+       "Assigned TAP circuit handshake stats."),
+  ITEM("rephist/tap/onion_handshakes_requested", rephist,
+       "Requested TAP circuit handshake stats."),
```
dgoulet-tor, line 1704: See torspec comment about the naming. We'll wait on a decision there.

dgoulet-tor: So, the `PREFIX` are not correct.

But also I would propose we use `stats/ntor/requested` and `stats/ntor/assigned`.

Reason I say that is that I think `onion` and `handshake` are a bit redundant since `ntor` is by definition an onion handshake.

PR #1910, opened 2020-06-02
#32888: Log address config info when tor starts up

c-chroniko:
Add debug logging as per [Trac #32888](https://trac.torproject.org/projects/tor/ticket/32888)

### DIFF
--- a/src/app/config/resolve_addr.c
+++ b/src/app/config/resolve_addr.c
@@ -110,6 +110,8 @@ resolve_my_address(int warn_severity, const or_options_t *options,
 
   if (address && *address) {
     strlcpy(hostname, address, sizeof(hostname));
+    log_debug(LD_CONFIG, "Trying configured Address '%s' as local hostname",
+              hostname);
   } else { /* then we need to guess our address */
     explicit_ip = 0; /* it's implicit */
     explicit_hostname = 0; /* it's implicit */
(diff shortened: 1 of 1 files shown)


src/lib/net/address.c
```diff
@@ -1693,12 +1693,20 @@ get_interface_address6,(int severity, sa_family_t family, tor_addr_t *addr))
   /* Find the first non-internal address, or the last internal address
    * Ideally, we want the default route, see #12377 for details */
   SMARTLIST_FOREACH_BEGIN(addrs, tor_addr_t *, a) {
+    char *addr_str;
```
nmathewson, line 1696: Style issue: if a variable is only set once, it's often best to put the declaration at the point where it is set, as in `char *addr_str = tor_addr_to_str_dup(addr);`

Same with is_internal, which could be a `const bool`

src/lib/net/address.c
```diff
@@ -1693,12 +1693,20 @@ get_interface_address6,(int severity, sa_family_t family, tor_addr_t *addr))
...
+    int is_internal;
     tor_addr_copy(addr, a);
+    is_internal = tor_addr_is_internal(a, 0);
     rv = 0;
 
+    addr_str = tor_addr_to_str_dup(addr);
```
nmathewson, line 1702: Rather than allocating and freeing manually, it could be better to use the fmt_addr(), which uses a static buffer.

PR #1931, opened 2020-06-11
[Tor Trac #34065]: Make routerset_contains_router() support IPv6

neelchauhan:
PR: https://trac.torproject.org/projects/tor/ticket/34065

### DIFF
--- a/src/feature/nodelist/routerset.c
+++ b/src/feature/nodelist/routerset.c
@@ -223,11 +223,11 @@ routerset_len(const routerset_t *set)
  *
  * (If country is -1, then we take the country
  * from addr.) */
-STATIC int
-routerset_contains(const routerset_t *set, const tor_addr_t *addr,
-                   uint16_t orport,
-                   const char *nickname, const char *id_digest,
-                   country_t country)
+static int
+routerset_contains2(const routerset_t *set, const tor_addr_t *addr,
+                    uint16_t orport, const tor_addr_t *addr2,
+                    uint16_t orport2, const char *nickname,
+                    const char *id_digest, country_t country)
 {
   if (!set || !set->list)
     return 0;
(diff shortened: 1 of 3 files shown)


src/feature/nodelist/routerset.c
```diff
@@ -306,14 +306,16 @@ int
...
+  tor_addr_t addr_v4;
+  tor_addr_from_ipv4h(&addr_v4, ri->addr);
+  int v4_ret =
+    routerset_contains(set, &addr_v4, ri->or_port, ri->nickname,
+                       ri->cache_info.identity_digest, country);
+  int v6_ret =
+    routerset_contains(set, &ri->ipv6_addr, ri->ipv6_orport, ri->nickname,
+                       ri->cache_info.identity_digest, country);
+
+  return MAX(v4_ret, v6_ret);
```
nmathewson, line 318: Why not `||`

src/feature/nodelist/routerset.c
```diff
@@ -306,14 +306,16 @@ int
...
-                            ri->nickname,
-                            ri->cache_info.identity_digest,
-                            country);
+  tor_addr_t addr_v4;
+  tor_addr_from_ipv4h(&addr_v4, ri->addr);
+  int v4_ret =
+    routerset_contains(set, &addr_v4, ri->or_port, ri->nickname,
+                       ri->cache_info.identity_digest, country);
+  int v6_ret =
+    routerset_contains(set, &ri->ipv6_addr, ri->ipv6_orport, ri->nickname,
```
nmathewson, line 315: 1) Doing these two separate checks means that routerset_contains_router will check the nickname and identity twice.  That's wasted computation that can add up.  It would be better to arrange things so that those only get checked once.

src/feature/nodelist/routerset.c
neelchauhan: Initially, this was because (at least on FreeBSD) `||` only returns 0 or 1.

nmathewson: This function is only documented as "returns true if", though.  If anything is depending on its return value being any specific value, then the documentation should get fixed.

PR #1932, opened 2020-06-14
#40060: Add IPv6 ORPort config debug logging

c-chroniko:
Per [ticket #40060](https://gitlab.torproject.org/tpo/core/tor/-/issues/40060) this should log "the IPv6 address of the first IPv6
ORPort torrc option".

### DIFF
--- a/src/feature/relay/router.c
+++ b/src/feature/relay/router.c
@@ -1475,6 +1475,9 @@ router_get_advertised_ipv6_or_ap(const or_options_t *options,
     return;
   }
 
+  log_debug(LD_CONFIG, "Found configured IPv6 ORPort '%s'.",
+            fmt_addrport(addr, port));
+
   /* If the relay is configured using the default authorities, disallow
    * internal IPs. Otherwise, allow them. For IPv4 ORPorts and DirPorts,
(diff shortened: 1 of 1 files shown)


src/feature/relay/router.c
```diff
@@ -1475,6 +1475,9 @@ router_get_advertised_ipv6_or_ap(const or_options_t *options,
     return;
   }
 
+  log_debug(LD_CONFIG, "Found configured IPv6 ORPort '%s'.",
+            fmt_addrport(addr, port));
+
```
dgoulet-tor, line 1480: So this function can be called quite often and thus we would spam the debug log here a bit too much in my opinion.

I think we might want to log this when we actually parse the ports and after they are sanitized. I think in `port_parse_ports_relay()` could be good after the "check and prune" process.

c-chroniko: I think I understand what you mean. So, we only want to print the ORPort config that is actually going to be in effect when Tor runs?

If there's no objection I'll make a second PR for this just to keep commit history clean.

c-chroniko:
Never mind, talked on #tor-dev and it seems this is unnecessary now

PR #1935, opened 2020-06-16
[Ticket #34400]: control: HSFETCH command fails to validate v2 addresses

neelchauhan:
Ticket: https://gitlab.torproject.org/tpo/core/tor/-/issues/34400


src/feature/control/control_cmd.c
```diff
@@ -1430,7 +1430,7 @@ handle_control_hsfetch(control_connection_t *conn,
              rend_valid_descriptor_id(arg1 + v2_str_len) &&
              base32_decode(digest, sizeof(digest), arg1 + v2_str_len,
                            REND_DESC_ID_V2_LEN_BASE32) ==
-                REND_DESC_ID_V2_LEN_BASE32) {
+                DIGEST_LEN) {
```
dgoulet-tor, line 1433: Can we just use `sizeof(digest)` here so if it ever (even though impossible) change size, we caught it. Just good practice overall in tor.git.

Thanks!

PR #1942, opened 2020-06-18
Ticket34067

nmathewson:


src/feature/relay/relay_periodic.c
```diff
@@ -202,19 +202,45 @@ reachability_warnings_callback(time_t now, const or_options_t *options)
...
-      if (address) {
+      /* We need to warn that one or more of our ORPorts isn't reachable.
+       * Determine which, and give a reasonable warning. */
+      char *address4 = tor_dup_ip(me->addr);
+      char *address6 = tor_addr_to_str_dup(&me->ipv6_addr);
+      bool v4_ok =
+        router_should_skip_orport_reachability_check_family(options,AF_INET);
+      bool v6_ok =
+        router_should_skip_orport_reachability_check_family(options,AF_INET6);
+      if (address4 || address6) {
```
dgoulet-tor, line 213: So hmm... this entire block can be skipped if `v4_ok` **and** `v6_ok` are `true` thus I would be incline to get those value before and bail if they are true.

If not, then we can start doing the warning dance and only then we can do more expensive address dup(), asprintf and control port message.

src/feature/relay/selftest.c
```diff
@@ -75,13 +79,35 @@ router_reachability_checks_disabled(const or_options_t *options)
...
-         can_reach_or_port;
+  if (reach_checks_disabled) {
+    return true;
+  }
+
+  if (family != AF_INET6) {
+    if (have_orport_for_family(AF_INET) && !can_reach_or_port_ipv4) {
+      return false;
+    }
+  }
```
dgoulet-tor, line 103: Huh, this took me a bit to process. Why this "if not IPv6 then try IPv4" logic instead of "if IPv4 then try Ipv4" ?

src/feature/relay/selftest.h
```diff
@@ -15,24 +15,29 @@
 #ifdef HAVE_MODULE_RELAY
 
 struct or_options_t;
-int router_should_skip_orport_reachability_check(
-                                         const struct or_options_t *options);
+#define router_should_skip_orport_reachability_check(opts) \
+  router_should_skip_orport_reachability_check_family((opts),0)
+int router_should_skip_orport_reachability_check_family(
+                                         const struct or_options_t *options,
+                                         int family);
```
dgoulet-tor, line 22: Hmm, is the `_family` really necessary in the name? I'm asking because some other functions take a `family` argument but do not have it in their name but this one does for some reasons.

For instance `router_do_orport_reachability_checks()` or `router_orport_found_reachable()`. I don't have a strong opinion here although I would lean towards **not** having the "family" in the name but either way, I would just make it the same standard for everyone regardless of what you decide.

src/feature/relay/relay_periodic.c
nmathewson: Should be addressed in b365179ee0061db2981f09025041a1065d0b283c

src/feature/relay/selftest.h
nmathewson: My reason for using a special name for this function was that there is already a `router_should_skip_orport_reachability_checks()` that checks for _all_ families.  

Honestly I think we should rename all of these functions completely -- 'should skip' seems wrong here.  Okay if I open a new ticket for this?

src/feature/relay/selftest.c
nmathewson: Tried to clarify in ac648fa3070b9f

src/feature/relay/selftest.h
dgoulet-tor: Sounds good!

PR #1958, opened 2020-06-30
Refactor reachability test logging: log exactly what we launch.

nmathewson:
Previously we had two chains of logic for reachability tests: one
for launching them, and one for telling the user that we had
launched them.  Now, we simply have the launch code inform the user:
this way, we can't get out of sync.

Closes ticket 34137.

### DIFF
--- a/src/feature/relay/selftest.c
+++ b/src/feature/relay/selftest.c
@@ -45,6 +45,9 @@
 #include "feature/relay/selftest.h"
 
 static bool have_orport_for_family(int family);
+static void inform_testing_reachability(const tor_addr_t *addr,
+                                        uint16_t port,
+                                        bool is_dirport);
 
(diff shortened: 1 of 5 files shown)


src/feature/relay/selftest.c
```diff
@@ -255,6 +258,9 @@ router_do_orport_reachability_checks(const routerinfo_t *me,
     log_info(LD_CIRC, "Testing %s of my %s ORPort: %s.",
              !orport_reachable ? "reachability" : "bandwidth",
              family_name, fmt_addrport(&ei->addr, ei->port));
+    if (!orport_reachable) {
+      inform_testing_reachability(&ei->addr, ei->port, false);
```
dgoulet-tor, line 262: I don't understand why this is logged only if ORPort is reachable but we still launch the circuit after?

Maybe a comment explaining why this is?

PR #1962, opened 2020-07-01
hs-v2: Add deprecation warning for service

dgoulet-tor:
If at least one service is configured as a version 2, a log warning is emitted
once and only once.

Closes #40003

### DIFF
--- a/src/feature/rend/rendservice.c
+++ b/src/feature/rend/rendservice.c
@@ -130,6 +130,22 @@ static smartlist_t *rend_service_list = NULL;
  * service on config reload. */
 static smartlist_t *rend_service_staging_list = NULL;
 
+/** Helper: log the deprecation warning for version 2 only once. */
+static void
+log_once_deprecation_warning(void)
+{
+  static bool logged_once = false;
+  if (!logged_once) {
+    log_warn(LD_REND, "DEPRECATED: Onion service version 2 are deprecated. "
+             "Please use version 3 which is the default now. "
+             "Currently, version 2 is planned to be obsolete in "
+             "the Tor version 0.4.6 stable series.");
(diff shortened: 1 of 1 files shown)


src/feature/rend/rendservice.c
```diff
@@ -731,6 +747,9 @@ rend_config_service(const config_line_t *line_,
   tor_assert(options);
   tor_assert(config);
 
+  /* We are about to configured a version 2 service. Warn of deprecation. */
```
asn-d6, line 750: `s/configured/configure/g`

src/feature/rend/rendservice.c
```diff
@@ -130,6 +130,22 @@ static smartlist_t *rend_service_list = NULL;
...
+/** Helper: log the deprecation warning for version 2 only once. */
+static void
+log_once_deprecation_warning(void)
+{
+  static bool logged_once = false;
+  if (!logged_once) {
+    log_warn(LD_REND, "DEPRECATED: Onion service version 2 are deprecated. "
+             "Please use version 3 which is the default now. "
+             "Currently, version 2 is planned to be obsolete in "
+             "the Tor version 0.4.6 stable series.");
```
asn-d6, line 142: Do you think we should link to the tor-dev post or the blog post here?

src/feature/rend/rendservice.c
dgoulet-tor: Fixup pushed!

src/feature/rend/rendservice.c
dgoulet-tor: As discussed, maybe not ideal to have links hardcoded in our tor codebase as in the link then has to be a super long permalink...

PR #1964, opened 2020-07-01
[Tor Gitlab #27315]: Fix seccomp sandbox rules for openat

Jigsaw52:
The need for casting negative syscall arguments depends on the
glibc version. This affects the rules for the openat syscall which
uses the constant AT_FDCWD that is defined as a negative number.
This commit adds logic to only apply the cast when necessary, on
glibc versions from 2.27 onwards.


Jigsaw52:
Merged.

PR #1966, opened 2020-07-01
[Tor Gitlab #40020]: Fix seccomp sandbox rules for opening directories

Jigsaw52:
Different versions of glibc use either open or openat for the
opendir function. This commit adds logic to use the correct rule
for each glibc version, namely:
- Until 2.14 open is used
- From 2.15 to to 2.21 openat is used
- From 2.22 to 2.26 open is used
- From 2.27 onwards openat is used


Jigsaw52:
Merged.

PR #1985, opened 2020-07-09
Strip '\r' characters when reading text files on Unix.

ahf:
This patch ensures that we strip "\r" characters on both Windows as well
as Unix when we read text files. This should prevent the issue where
some Tor state files have been moved from a Windows machine, and thus
contains CRLF line ending, to a Unix machine where only \n is needed.

We add a test-case to ensure that we handle this properly on all our
platforms.

See: https://bugs.torproject.org/tpo/core/tor/33781


ahf:
This needs to be based on 0.3.5.

PR #1994, opened 2020-07-10
Rename blacklist and whitelist wording

dgoulet-tor:
Closes #40033

### DIFF
--- a/src/app/config/fallback_dirs.inc
+++ b/src/app/config/fallback_dirs.inc
@@ -5,8 +5,8 @@
 /* timestamp1=20190628085927 */
 /* source=allowlist */
 /* ===== */
-/* 0: Whitelist excluded 1550 of 1711 candidates. */
-/* 1: Whitelist excluded 1601 of 1765 candidates. */
+/* 0: Allowlist excluded 1550 of 1711 candidates. */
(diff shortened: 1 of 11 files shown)


src/app/config/fallback_dirs.inc
```diff
@@ -3,7 +3,7 @@
 /* timestamp=20190625114911 */
 /* timestamp0=20190625114911 */
 /* timestamp1=20190628085927 */
-/* source=whitelist */
+/* source=allowlist */
 /* ===== */
 /* 0: Whitelist excluded 1550 of 1711 candidates. */
 /* 1: Whitelist excluded 1601 of 1765 candidates. */
```
nmathewson, line 9: Why wasn't this instance caught?

src/app/config/fallback_dirs.inc
```diff
@@ -3,7 +3,7 @@
 /* timestamp=20190625114911 */
 /* timestamp0=20190625114911 */
 /* timestamp1=20190628085927 */
-/* source=whitelist */
+/* source=allowlist */
```
nmathewson, line 6: Does this require a corresponding change in the fallback scripts?

PR #2002, opened 2020-07-14
#33781 [035]: Handle \r characters better when reading files.

ahf:
Strip '\r' characters when reading text files on Unix platforms. This should resolve an issue where a relay operator migrates a relay from Windows to Unix, but does not change the line ending of Tor's various state files to match the platform, the CRLF line endings from Windows ends up leaking into other files such as the extra-info document.


ahf:
This was merged, but not backported to 0.3.5 yet.

PR #2019, opened 2020-07-22
relay: Support IPv6 when checking if our address changed

dgoulet-tor:
Now support IPv4 _and_ IPv6.

This also cleans up nicely the function that was moving IPv4 addresses from
uint32_t to tor_addr_t.

Fixes #40058

### DIFF
--- a/src/feature/relay/router.c
+++ b/src/feature/relay/router.c
@@ -2621,16 +2621,15 @@ check_descriptor_ipaddress_changed(time_t now)
       previous = &my_ri->ipv6_addr;
     }
 
-    if (!find_my_address(get_options(), family, LOG_INFO, &current, &method,
-                         &hostname)) {
-      /* Not found, continue onto next family. Logging is done within the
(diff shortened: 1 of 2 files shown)


src/feature/relay/router.c
```diff
@@ -2621,16 +2621,15 @@ check_descriptor_ipaddress_changed(time_t now)
...
 
-    if (!find_my_address(get_options(), family, LOG_INFO, &current, &method,
-                         &hostname)) {
-      /* Not found, continue onto next family. Logging is done within the
-       * function called. */
-      continue;
-    }
+    /* Ignore returned value because we want to notice not only an address
+     * change but also if an address is lost (current == UNSPEC). */
+    find_my_address(get_options(), family, LOG_INFO, &current, &method,
```
nmathewson, line 2626: It looks like this can leave `current` unset?  That's probably a bug.  We should at least set `current` to AF_UNSPEC or something above.

PR #2028, opened 2020-07-25
#40062: Tor always overwrites onion service hostname file, but should leave it alone if it shouldn't change

neelchauhan:
Ticket: https://gitlab.torproject.org/tpo/core/tor/-/issues/40062

### DIFF
--- a/changes/bug40062
+++ b/changes/bug40062
@@ -1,0 +1,5 @@
+  o Major bugfixes (onion services):
+    - When writing an onion service hostname file, first read it to make
+      sure it contains what we want before attempting to write it. This
+      prevents a bug where we can't read an onion service directory if it's
+      read-only. Fixes bug 40062; bugfix on 0.0.6. Patch by Neel Chauhan.
--- a/src/feature/relay/router.c
+++ b/src/feature/relay/router.c
@@ -870,15 +870,11 @@ router_write_fingerprint(int hashed, int ed25519_identity)
   tor_asprintf(&fingerprint_line, "%s %s\n", options->Nickname, fingerprint);
 
   /* Check whether we need to write the (hashed-)fingerprint file. */
-
-  cp = read_file_to_str(keydir, RFTS_IGNORE_MISSING, NULL);
-  if (!cp || strcmp(cp, fingerprint_line)) {
-    if (write_str_to_file(keydir, fingerprint_line, 0)) {
-      log_err(LD_FS, "Error writing %s%s line to file",
-              hashed ? "hashed " : "",
-              ed25519_identity ? "ed25519 identity" : "fingerprint");
-      goto done;
-    }
+  if (write_str_if_not_equal(keydir, fingerprint_line)) {
+    log_err(LD_FS, "Error writing %s%s line to file",
+            hashed ? "hashed " : "",
+            ed25519_identity ? "ed25519 identity" : "fingerprint");
+    goto done;
   }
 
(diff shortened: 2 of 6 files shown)


src/lib/fs/files.c
```diff
@@ -718,6 +718,25 @@ read_file_to_str, (const char *filename, int flags, struct stat *stat_out))
   return string;
 }
 
+/** Attempt to read a file <b>fname</b>. If the file's contents is
+ * equal to the string <b>str</b>, return 0. Otherwise, attempt to
+ * overwrite the file with the contents of <b>str</b> and return
+ * the value of write_str_to_file().
+ */
+int
+write_str_if_not_equal(const char *fname, const char *str)
```
armadev, line 727: This one might be clearer for future readers as write_str_to_file_if_not_equal. That way it's more obvious that it's like write_str_to_file, just a special case of it.

src/lib/fs/files.c
```diff
@@ -718,6 +718,25 @@ read_file_to_str, (const char *filename, int flags, struct stat *stat_out))
...
+{
+  char *fstr;
+
+  fstr = read_file_to_str(fname, RFTS_IGNORE_MISSING, NULL);
```
armadev, line 731: This can be turned into a single line where you declare the variable and assign to it

src/feature/relay/router.c
```diff
@@ -870,15 +870,11 @@ router_write_fingerprint(int hashed, int ed25519_identity)
   tor_asprintf(&fingerprint_line, "%s %s\n", options->Nickname, fingerprint);
 
   /* Check whether we need to write the (hashed-)fingerprint file. */
-
-  cp = read_file_to_str(keydir, RFTS_IGNORE_MISSING, NULL);
```
armadev, line 874: Are we using cp for anything else in this function? If not we should stop declaring it too.

src/lib/fs/files.c
```diff
@@ -718,6 +718,25 @@ read_file_to_str, (const char *filename, int flags, struct stat *stat_out))
...
+
+  if (!fstr || strcmp(str, fstr)) {
```
armadev, line 733: Careful! read_file_to_str() allocates a new string, and by returning without any calls to tor_free(fstr), you might be leaking the contents of fstr.

changes/bug40062
```diff
@@ -0,0 +1,5 @@
+  o Major bugfixes (onion services):
+    - When writing an onion service hostname file, first read it to make
+      sure it contains what we want before attempting to write it. This
+      prevents a bug where we can't read an onion service directory if it's
+      read-only. Fixes bug 40062; bugfix on 0.0.6. Patch by Neel Chauhan.
```
armadev, line 5: I'd actually call this a 'Minor feature', since the behavior didn't exist before, and now it does. And then change 'This prevents a bug where' to something like 'Now onion services can set their existing onion service directories to read-only and Tor will still work'. I think the convention for features is to say "Resolves ticket x" rather than "Fixes bug x".

I'd also suggest squashing the "Use write_str_if_not_equal() for v2 onion services" and "Use write_str_if_not_equal() for v3 onion services" commits into this one, so ultimately there is one commit that refactors the function, and one commit that makes the two new changes plus adds the changes file.

Having the two behavior changes, plus the changes file, in a single commit might help somebody in the distant future who is trying to figure out what changed, and they see one commit but they don't realize there is a nearby very similar one.

Thanks!

PR #2029, opened 2020-07-27
#40062: Tor always overwrites onion service hostname file, but should leave it alone if it shouldn't change (Revision 1)

neelchauhan:
Ticket: https://gitlab.torproject.org/tpo/core/tor/-/issues/40062

### DIFF
--- a/src/lib/fs/files.c
+++ b/src/lib/fs/files.c
@@ -729,6 +729,9 @@ write_str_to_file_if_not_equal(const char *fname, const char *str)
   char *fstr = read_file_to_str(fname, RFTS_IGNORE_MISSING, NULL);
 
   if (!fstr || strcmp(str, fstr)) {
+    if (fstr) {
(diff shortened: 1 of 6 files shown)


src/lib/fs/files.c
```diff
@@ -718,6 +718,26 @@ read_file_to_str, (const char *filename, int flags, struct stat *stat_out))
...
+ * overwrite the file with the contents of <b>str</b> and return
+ * the value of write_str_to_file().
+ */
+int
+write_str_to_file_if_not_equal(const char *fname, const char *str)
+{
+  char *fstr = read_file_to_str(fname, RFTS_IGNORE_MISSING, NULL);
+
+  if (!fstr || strcmp(str, fstr)) {
+    return write_str_to_file(fname, str, 0);
```
nmathewson, line 732: In this case, nothing will free `fstr`.

PR #2030, opened 2020-07-27
#40066: Make policies such as ExitPolicy truly dual-stack

c-chroniko:
See [#40066](https://gitlab.torproject.org/tpo/core/tor/-/issues/40066) ticket for further details. Basically, add new `private4` and `private6` keywords for policies, make sure all warnings/errors in parsing are what we expect, and ensure we can pave the way toward obsoleting `accept6` and `reject6` in favour of `accept` and `reject` in the distant future.

### DIFF
--- a/src/feature/dirparse/policy_parse.c
+++ b/src/feature/dirparse/policy_parse.c
@@ -35,13 +35,17 @@ static addr_policy_t *router_parse_addr_policy_private(directory_token_t *tok);
  * on error - the rest of the policy list can continue to be processed and
  * used.
  *
+ * Set <b>is_torrc</b> to true if torrc-specific parsing should take place.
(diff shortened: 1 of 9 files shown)


src/feature/dirparse/policy_parse.c
```diff
@@ -201,11 +202,24 @@ router_parse_addr_policy_private(directory_token_t *tok)
...
     return NULL;
 
+  switch (*arg) {
+    case '4':
+      private_flag = ADDR_POLICY_PRIVATE4;
+      break;
+    case '6':
+      private_flag = ADDR_POLICY_PRIVATE6;
+      break;
+    default:
```
nmathewson, line 212: can this really be any character here?  or are these some characters we don't want to accept?

PR #2032, opened 2020-07-28
[Tor Gitlab #40072 - maint-0.4.4]: Fix startup crash with seccomp sandbox enabled

Jigsaw52:
Fix crash introduced in #40020. On startup, tor calls
check_private_dir on the data and key directories. This function
uses open instead of opendir on the received directory. Data and
key directoryes are only opened here, so the seccomp rule added
should be for open instead of opendir, despite the fact that they
are directories.


Jigsaw52:
Merged.

PR #2055, opened 2020-08-07
v3 control: Persist ONION_CLIENT_AUTH_ADD client name

neelchauhan:


src/feature/control/control_hs.c
```diff
@@ -134,6 +134,13 @@ handle_control_onion_client_auth_add(control_connection_t *conn,
         }
       } SMARTLIST_FOREACH_END(flag);
     }
+    if (!strcasecmpstart(line->key, "ClientName")) {
```
nmathewson, line 137: Probably strcasecmp() would be better than strcasecmpstart(). Do you see why?

src/feature/control/control_hs.c
```diff
@@ -134,6 +134,13 @@ handle_control_onion_client_auth_add(control_connection_t *conn,
...
+      if (strlen(line->value) > REND_CLIENTNAME_MAX_LEN) {
+        control_printf_endreply(conn, 512, "ClientName longer than %d chars",
+                                REND_CLIENTNAME_MAX_LEN);
+      }
+      memcpy(creds->client_name, line->value, strlen(line->value));
```
nmathewson, line 142: strlcpy would be more usual here.

src/feature/control/control_hs.c
```diff
@@ -255,6 +262,10 @@ encode_client_auth_cred_for_control_port(
     }
   }
 
+  if (*cred->client_name) {
```
asn-d6, line 265: Hmm, I think this is naughty but I understand why you did it.

Instead of doing this dereference to understand if a static array is filled up, let's do the following:
Let's change `client_name` to a pointer, and then in `handle_control_onion_client_auth_add()` if a client name is provided we can allocate some memory and fill it in. So now in `encode_client_auth_cred_for_control_port()` we can do the more natural `if (cred->client_name) {`. We will also need to free `client_name` in `client_service_authorization_free_()`. It's a good time to do this now that this struct is not really used so it's easy.

src/feature/control/control_hs.c
```diff
@@ -134,6 +134,13 @@ handle_control_onion_client_auth_add(control_connection_t *conn,
         }
       } SMARTLIST_FOREACH_END(flag);
     }
+    if (!strcasecmp(line->key, "ClientName")) {
+      if (strlen(line->value) > REND_CLIENTNAME_MAX_LEN) {
+        control_printf_endreply(conn, 512, "ClientName longer than %d chars",
+                                REND_CLIENTNAME_MAX_LEN);
+      }
+      strlcpy(creds->client_name, line->value, strlen(line->value) + 1);
```
asn-d6, line 142: Hmm, I don't like this `strlen() + 1`.
Let's instead change it to use the `sizeof()` of the destination like it happens a bit above:
`  strlcpy(creds->onion_address, hsaddress, sizeof(creds->onion_address));`

PR #2141, opened 2023-10-28
Update TUNING for OpenBSD

wouterprins:


alirezaarzehgar:
> Why did you remove sudo with doas?

By default OpenBSD haven't `sudo`. You should use `doas` on OpenBSD.
Using `doas` for OpenBSD tuning doc is better idea than `sudo`.

A1exKH (changes requested):
Please resolve conflicts in this PR.

